PR #30, opened 2016-01-15
bugfix/firefox-datagrid-filter

cerdman:
Fixes an issue with the Faceted Data Table component filter not working in Firefox.
* Tested on Firefox edge version on OSX (Would be good to confirm on windows just in case)
* In the event that this issue is still present elsewhere in the app, the root cause relates to this knockout.js issue: #752 https://github.com/knockout/knockout/issues/752


vojtechhuser:
nice to see that it was a real problem. I can test it on windows. If you give me an URL. best email to my work email to ping me

PR #141, opened 2016-06-09
Explicit check of d.Y_PREVALENCE_1000PP (fixes #137)

### LINKED ISSUE #137: Data source reports contain hidden age decile line graphs and NaNs
Age decile line graphs are hidden and showing the text "NaN" (see screenshot below). The impacted age deciles differ depending on the selected concept. We suspect it could be related to outliers in the data, but it is unclear.
hidden-line-graph-nan

mark-velez:
Evaluation of `Y_PREVALENCE_1000PP` when `Y_PREVALENCE_1000PP=0` will cause `yPrevalence1000Pp` to be evaluated (a property which doesn't exist in later versions of Achilles output) and yields undefined and thus results in NaN values in graphs. Using `hasOwnProperty` seems to produce the desired behavior in the contexts I have described in the issue.

### DIFF
--- a/js/jnj.chart.js
+++ b/js/jnj.chart.js
@@ -1665,14 +1665,14 @@
 			var minY = d3.min(dataByTrellis, function (trellis) {
 					return d3.min(trellis.values, function (series) {
 						return d3.min(series.values, function (d) {
-							return d.hasOwnProperty('Y_PREVALENCE_1000PP') ? d.Y_PREVALENCE_1000PP : d.yPrevalence1000Pp;
+							return (d.Y_PREVALENCE_1000PP === 0 || d.Y_PREVALENCE_1000PP) ? d.Y_PREVALENCE_1000PP : d.yPrevalence1000Pp;
 						});
 					});
 				}),
 				maxY = d3.max(dataByTrellis, function (trellis) {
 					return d3.max(trellis.values, function (series) {
 						return d3.max(series.values, function (d) {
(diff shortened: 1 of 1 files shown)


chrisknoll:
Mark, 
  I understand the root of the issue now:  the code you changed is an 'default' handler where if the value isn't provided, it uses a default stored in yPrevalence1000Pp.  Apparantly, the default wasn't passed in, so it returned an 'undefined' value which I think becomes NaN.    But, if the value checked has the value of 0, that's 'falsy' and returns the 'else' clause, which is to return the default value (which isn't specified, resulting in NaN, etc etc).  0 should be a valid case!

So, while using hasOwnProperty fixes this, it doesn't handle the case where the property IS provided but it's null.  So, i think the check should be changed from:
~~~~javascript
return d.Y_PREVALENCE_1000PP ? d.Y_PREVALENCE_1000PP : d.yPrevalence1000Pp;
~~~

to:
~~~javascript
return (d.Y_PREVALENCE_1000PP == 0 || d.Y_PREVALENCE_1000PP)  ? d.Y_PREVALENCE_1000PP : d.yPrevalence1000Pp;
~~~

I'm wondering if there's another simpler way in javascript to say 'treat every falsy value EXCEPT 0 as falsy' in this case....

What do you think?

mark-velez:
@chrisknoll yes, you're right. I made the fix, except I used strict equality `===` checks. Javascript paranoia.

PR #253, opened 2016-10-05
Added binding to title of page.

### LINKED ISSUE #211: provide more information in title
It would be helpful especially when you have more than one tab open for ATLAS to provide more information in the title that would appear in the tab.

### LINKED ISSUE #252: (Feature) Add more detail to browser title
Hi,

Currently, the browser title, on any page, is just 'ATLAS'. I was thinking it might be useful to have something like 'ATLAS: Search for Clopidogrel' or 'ATLAS: Cohort 1234' or 'ATLAS: Concept Set 1234'. Additionally, including an asterisk to the title if there are unsaved changes could also be useful in identifying your active cohorts or concept sets.

Hopelessly lost in Chrome tabs,
Ajit

chrisknoll:
Fixes #252
Fixes #211

The title computed handled specific currentView() values. If any are missing, please report it here, and updates can be made.

### DIFF
--- a/index.html
+++ b/index.html
@@ -22,7 +22,7 @@
 	<link rel="stylesheet" type="text/css" href="js/styles/achilles.css">
 	<!-- cartoon.css, maybe others, loaded dynamically -->
 
-	<title>ATLAS</title>
+	<title data-bind="text: pageTitle">ATLAS</title>
 	<!--Adobe Edge Runtime-->
 	<meta http-equiv="X-UA-Compatible" content="IE=Edge">
 </head>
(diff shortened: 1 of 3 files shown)


chrisknoll:
I've added the * as a prefix to the title, for future reference, let's include an example of the treatment of how we'd want this displayed in the feature request (could have been as a prefix or suffix, so let's be clear on those sorts of details).

For item 2, I don't think it's worth the effort without more structure to our lower level components as being structured as a 'page' of the app that we can pass page level state between pages.  But that goes a bit beyond what is being accomplished in this PR. Maybe that would be a suggestion for a separate PR.

PR #258, opened 2016-10-11
trying to use linters and fixing achilles

fdefalco:
It turns out the issues we were seeing with the legacy achilles (data sources) charts was due to some undeclared variables throwing errors once "use strict" was applied on the charting module.  I've put a few things into this branch that I wanted to open up for comment before pushing to master.

I added .jshintrc which is a preference file for JSHint.  JSHint is a javascript code analysis tool.  The .jshintrc file will allow us to set conventions for our code that we can then test for compliance.  The current .jshintrc is by no means "our accepted standards" just a template I found that seemed well formed.  I then made changes to bring the charting module file down from a few hundred JSHint problems to 39.  I stopped there as the problems were getting harder to resolve and I wasn't sure if this was a path we would want to adopt.

If we are in agreement and get approval on this branch then we can use JSHint (with extensions for brackets, vim, emacs, visual studio, notepad++ and other editors) going forward and make progress on "house cleaning" of our javascript code as time allows.  While this doesn't address the broader refactoring we need to do it should help get us moving in the right direction and give us a way to hold ourselves accountable to some quality standard.

This branch does fix the data sources modules and reports are now working so I've re-enabled it on the left menu.  I also made some style updates to the dropdown buttons for reports and data sources to fix  the display issues with the buttons.

I also changed so that "use strict" is occurring at a function level instead of module as that was the approach I've seen recommended.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+/web.config
--- a/.jshintrc
+++ b/.jshintrc
@@ -27,7 +27,7 @@
 	// with different skill levels.
 	"bitwise": true, // Prohibit bitwise operators (&, |, ^, etc.).
 	"curly": true, // Require {} for every new block or scope.
-	"eqeqeq": true, // Require triple equals i.e. `===`.
+	"eqeqeq": false, // Require triple equals i.e. `===`.
 	"forin": true, // Tolerate `for in` loops without `hasOwnPrototype`.
 	"immed": true, // Require immediate invocations to be wrapped in parens e.g. `( function(){}() );`
 	"latedef": true, // Prohibit variable use before definition.
@@ -35,7 +35,7 @@
 	"noarg": true, // Prohibit use of `arguments.caller` and `arguments.callee`.
 	"noempty": true, // Prohibit use of empty blocks.
 	"nonew": true, // Prohibit use of constructors for side-effects.
-	"plusplus": true, // Prohibit use of `++` & `--`.
+	"plusplus": false, // Prohibit use of `++` & `--`.
 	"regexp": true, // Prohibit `.` and `[^...]` in regular expressions.
 	"undef": true, // Require all non-global variables be declared before they are used.
 	"strict": false, // Require `use strict` pragma in every file.
@@ -51,8 +51,6 @@
 	"boss": false, // Tolerate assignments inside if, for & while. Usually conditions & loops are for comparison, not assignments.
 	"debug": false, // Allow debugger statements e.g. browser breakpoints.
 	"eqnull": false, // Tolerate use of `== null`.
-	"es5": false, // Allow EcmaScript 5 syntax.
-	"esnext": false, // Allow ES.next specific features such as `const` and `let`.
 	"evil": false, // Tolerate use of `eval`.
 	"expr": false, // Tolerate `ExpressionStatement` as Programs.
 	"funcscope": false, // Tolerate declarations of variables inside of control structures while accessing them later from the outside.
@@ -78,10 +76,10 @@
 	// These options pre-define global variables that are exposed by
 	// popular JavaScript libraries and runtime environments—such as
 	// browser or node.js.
-	"esversion": 6,
+	"esversion": 5,
 	"browser": true, // Standard browser globals e.g. `window`, `document`.
 	"couch": false, // Enable globals exposed by CouchDB.
-	"devel": true, // Allow development statements e.g. `console.log();`.
+	"devel": false, // Allow development statements e.g. `console.log();`.
 	"dojo": false, // Enable globals exposed by Dojo Toolkit.
 	"jquery": false, // Enable globals exposed by jQuery JavaScript library.
 	"mootools": false, // Enable globals exposed by MooTools JavaScript framework.
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,0 +1,124 @@
+{
+	// --------------------------------------------------------------------
+	// JSHint Configuration, Strict Edition
+	// --------------------------------------------------------------------
+	//
+	// This is a options template for [JSHint][1], using [JSHint example][2]
+	// and [Ory Band's example][3] as basis and setting config values to
+	// be most strict:
+	//
+	// * set all enforcing options to true
+	// * set all relaxing options to false
+	// * set all environment options to false, except the browser value
+	// * set all JSLint legacy options to false
+	//
+	// [1]: http://www.jshint.com/
+	// [2]: https://github.com/jshint/node-jshint/blob/master/example/config.json
+	// [3]: https://github.com/oryband/dotfiles/blob/master/jshintrc
+	//
+	// @author http://michael.haschke.biz/
+	// @license http://unlicense.org/
+
+	// == Enforcing Options ===============================================
+	//
+	// These options tell JSHint to be more strict towards your code. Use
+	// them if you want to allow only a safe subset of JavaScript, very
+	// useful when your codebase is shared with a big number of developers
+	// with different skill levels.
+	"bitwise": true, // Prohibit bitwise operators (&, |, ^, etc.).
+	"curly": true, // Require {} for every new block or scope.
+	"eqeqeq": true, // Require triple equals i.e. `===`.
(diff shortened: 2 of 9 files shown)


t-abdul-basser:
+1 on adding a JSHint preference file and making the associated changes.
 
Thanks for all of this work @fdefalco!

chrisknoll:
Did you alter any of the lint settings from the default, or did you fix the Achilles code based on the default lint settings?  I was reviewing the settings, but not sure if they were changed from the default, and if so, what was the reasoning of changing a setting vs. de-liniting the source?

fdefalco:
The .jshintrc I added was referenced in a few projects that I found using JSHint.  I did alter some of the settings that were in the template.  

- I commented out the entire section focused on cyclomatic complexity, I didn't think it was necessary.  
- I added the option for "esversion" and set it to 6 to allow some of the features that have been added recently, this is certainly up for discussion.
- I changed "devel" from false to true to allow for console logging as it appears in quite a few places.

If you use brackets to review the remaining problems you'll want to update your preference file to include the following section, without it it will show you the Lint problems from multiple linters.

    "language": {
        "javascript": {
            "linting.prefer": [
                "JSHint"
            ],
            "linting.usePreferredOnly": true
        }
    },

chrisknoll:
Ok, so from what I can tell after reviewing, the main fix for the Achilles is to remove the "use strict" from the file to remove the errors, and add the nav back on the left side. I saw some other things, related to an observable "minibar", which not sure what the change there addresses, in addition to some style changes.  That would be a good PR to get that fix into master.  

The second item about setting up the linters looks like  a good idea, but incomplete in the specification.  Does the PR get approved and the unfinished specs get merged in and people start following them for the sake of getting the achilles reports back online?  or do we hold up the merge until the .jshintrc file gets vetted?

I try to advocate for focused branches and PRs to avoid these sorts of conflicts, but if we all understand that we should ignore the lint warnings until they are all worked out, I don't see the harm in merging in functionality that only would result in some IDE noise from the linter.

Anyone else have thoughts?

fdefalco:
The specification for the .jshintrc is complete, unless you would like to suggest changes, which I welcome.  

To be clear, "use strict" was not removed entirely but moved into the functions that @Sigfried had added to the charting library to try and maintain his intentions.  The variables in other functions that had been effected have also been updated to be declared (var) prior to use although "use strict" has not been retroactively added to all functions.  I think that is something we would want to discuss as a separate refactoring.

The reason I left these changes combined in a single PR is because I made use of the JSHint listing to help me identify the issues that were causing the charting library to fail and break the reports.  I feel like this is the type of refactoring we want to encourage as it directly impacts the functionality of the application.

Anyone using Brackets has been ignoring the lint warnings since the beginning of time as it does have integrated linting so there is nothing new there.  The inclusion of the .jshintrc will only provide warnings in editors that support it and will not change the functionality of the site.

t-abdul-basser:
+1

anthonysena:
I'm on board with merging this into master - it allows us to re-enable data sources and also gives us a good start at defining JS standards in the code base. I'm totally fine with any additional noise in the IDE with the benefit that it will help us improve the consistency of the JS. I'm sure we'll want to revisit some of these settings down the line but for now I think this is an excellent start.

chrisknoll:
Several things here, I'll try to cover them one at a time:

1. Thanks for clarifying, I found those added 'use strict' lines to the local functions. Moreover, we shouldn't be using 'use strict' at a global script level for reasons described here:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode
So, your change addresses two problems, so that's good.

2. I think you're conflating the issues of refactoring, adding linting and fixing an issue into a single PR as the encouraged activity.  Factoring is good,  Bug fixing is good.  Adopting better practices is good.  Putting all of those things into one PR?  Not good because now we have to address all those elements as part of a singe PR review.  If i were to fix a bug by introducing a testing framework and build system, I think you'd agree that a single PR that fixes a bug, introduces a test framework and introduces a build system would probably be too much for a single PR.  instead, separate PRs for those things would be preferred so that each one can be evaluated in isolation (unless the bug with Achilles will remain broken without the linting configuration changes, but that's not the case here). So, we'll have to deal with the linting configuration part of the PR which leads us to...

3. I don't think the .jshintrc file you've provided should be used as is. I could either add comments into the commuted file directly, or call it out here, not sure which would be more effective.  Here's the items I think should be changed:

"eqeqeq": sounds good to always use === for comparison, but there's a special case described here: http://stackoverflow.com/questions/359494/which-equals-operator-vs-should-be-used-in-javascript-comparisons.  I'm not sure of the case where we get a value from a AJAX call from the server and we want to compare a string value to a value from the AJAX, will === return false because they aren't referencing the same string reference?  Testing is required for this.

"plusplus": I don't think using for (var i = 0; i< 20; i++) should be a linting error because of the ++ use.
 
"smarttabs":  I agree with the setting of 'false' but we need to all get into alignment as to how to indent/format our code.  Many commits are in the repo with lines changed where it was a tab-to-space or space-to-tab or -number of tabs in an indent problem.  

"esversion": I'm not sure 6 is what we want here, the docs give the options of 3, 5 and 6: http://jshint.com/docs/options/.  An example of an ES6 feature that caused us problems recently is here: http://forums.ohdsi.org/t/atlas-on-tomcat-works-from-remote-host-but-not-localhost/1741/9

"devel": I think this should be false, I think people should be warned about leaving console.log messages in the final code, since it probably represents debugging and not actual functionality.  And the editor would be able to show you where that is happening.

Formatting and Complexity: I agree most of these are things that really depend on the complexity of the code.  So having a hard rule about "no more than 10 function params" is not something I want to see. However, there's one value that is worth discussing:

"indent": This item is actually going to be removed from jslint due to "JSHint is limiting its scope to issues of code correctness. If you would like to enforce rules relating to code style, check out the JSCS project"  So, perhaps the issues raised above about code style and formatting is actually beyond scope of this PR but it is definitely something we all need to come to agreement over.

Again, this is a good idea and i'm not in any way pushing back against having linting in code.  I just want to make sure that we get as close to the 'accepted settings' as possible before we release it out to the general public.

fdefalco:
Understood and I didn't mean to complicate the PR but in this particular case I used the linting to identify the problem variables so I would have had to introduce linting in my environment, configure it, leverage it to fix the bugs, then go back and remove/ignore the linting to submit a PR with only the bug fix.  I understand that would have been cleaner and will try to work towards simpler PRs.

I agree with making the changes to eqeqeq (change to false), plusplus (change to false), and devel (change to false).

It looks like based on the issues we are seeing with browser compatibility that we should set esversion to 5 in order to warn on es6 based code.  I'm also going to suggest we remove esnext (deprecated) and es5 (deprecated) from the .jshintrc file.

On the topic of tabs and spaces I've setup a poll on the OHDSI forums here:

http://forums.ohdsi.org/t/indent-rules-at-least-for-js-at-least-for-atlas/1783/4

We can let people weigh in and see where we land.  Ultimately, as you pointed out, I'm not sure that JSHint will enforce this for us but we should establish a standard and then stick with it.  Once we determine what we will use we can document it on the wiki under the developer guidelines section:

http://www.ohdsi.org/web/wiki/doku.php?id=development:guidelines

chrisknoll:
Great, thanks.  If you want to push up those commits to make those adjustments, and I'll throw in my review approval for what it's worth.  Thanks.

Sigfried:
Just to chime in late to the conversation: 

=== is fine for strings where the strings contain the same characters.

on esversion, sorry for creating problem with using newer features, but rather than backing these out and discouraging their use, we should use babel. then code will work in older browsers.

chrisknoll:
@Sigfried : not true.  From the above link I posed:

Now it becomes interesting:

var a = "12" + "3";
var b = "123";

alert(a === b); // returns true, because strings behave like value types
But how about this?:

var a = new String("123");
var b = "123";

alert(a === b); // returns false !! (but they are equal and of the same type)

b has the same string "123" as a ("123") but === will evaluate to false.  So what I'm saying we need to check is that if we get a string field back from an AJAX call, and we want to compare a value in javascript to the ajax'd field using ===, will it return true or false?

PR #264, opened 2016-10-19
Apache Shiro integration

ganisimov:


chrisknoll:
Whoever does the merge for this, please ensure that either this PR is squelched down to a single commit, or use the squelch and merge option via the 'merge pull request' option in github.

chrisknoll:
@ganisimov , 
  Thank you very much for putting this work together.  I've tried to digest most of what is happening here, but I think as part of this PR, we should also put together some minimum documentation about what this PR is bringing into the codebase.  I've started a wiki page here:
http://www.ohdsi.org/web/wiki/doku.php?id=development:security

Would you mind providing the details to answer the questions there, or if there's someone you can delegate to do that, it would be appreciated.

chrisknoll:
@ganisimov ,
  Hi, I'm looking through the code, and wanted to ask you about how you're auto numbering the entity IDs for the shiro tables:  you're using a sequence which is a bit different then the way the other entities IDs are managed (although the nice way about how you are doing it is that it's very explicit about using a sequencer, and the other cases use an 'automatic' mode which means it uses a sequence for oracle/postgresql but an identity column in MSSQL).  Are you certain that the syntax for the 'default next value' that you're specifying in the DDL for the security tables works correctly in the different database platforms? And a related question: i noticed that the sequence you set up starts at 1000 for those new sequences, is there any significance for starting the IDs of these entities at 1000?  I'm not so much concerned about not having 1000 IDs in the sequence, i just wasn't sure if < 1000 in the id has a special meaning.

Thanks again for your efforts on this, it's really great!

ganisimov:
@chrisknoll,
In regards to sequences, I tested it with PostgreSQL and MSSQL. 
IDs before 1000 are reserved for convenience. They may be used to assign predefined roles and permission to each other using hardcoded IDs. Please, see shiro_security-initial_values migration. 
I’ll complete the wiki page a bit later. Thank you!

chrisknoll:
Ok, the wiki page looks good to me.  I do have additional questions on the back-end side of the security integration so I'll post those on the other PR related to WebAPI.  From the Atlas side, I'm not sure I'm feeling that all the UI elements are polished off or just in a proof-of-concept state (such as choosing the type of authentication provider you want to use is currently a dropdown i think?)  but I'm not sure what the UI spec for this feature looks like to say that this branch is covering those features.  So, my main concern was getting the technical details covered in a wiki article, and you nailed that, so thank you.

Can anyone chime in as to what an 'approval criteria' would look like for this PR? Is it just a first step or are there more features that should be added?

t-abdul-basser:
Thanks @chrisknoll! Re 'approval criteria': how do we feel about unit test coverage?

chrisknoll:
On the UI side, we don't have a test framework in place. On the server side, we have a unit test framework, tho not entirely sure how to embed security/authentication unit tests in that context.  Would anyone who grabs the repo have to have access to the same authentication providers as any other person who grabs the repo?  TBH, not really sure how this would work,this isn't the style of unit tests I've worked with in the past.

On the other hand, I can imagine unit tests for manipulating the entity repository (SEC_ROLE, SEC_USER, etc).  That type of test would just be a matter of testing that Hibernate works, but beyond that it would actually ensure that the webapi tables were properly initialized, so this could be a good thing, actually.

But to answer your question, not sure what measure we use to determine adequate unit test coverage.

fdefalco:
The scope of work for @ganisimov was primarily on the back-end additions to WebAPI with minor updates to the UI in order to incorporate basic security features.  I would not suggest that we base a release off of the result of merging this branch into master in fact if we want we could instead look to merge this PR into a separate ATLAS repository branch and then go forward with finalizing the UI changes there.  Would that be preferable?

chrisknoll:
If you're saying that @ganisimov isn't the one who will be finalizing the UI changes, and the people who will be aren't going to want to commit changes into the branch on his fork, then I don't see any alternative.  If you want he work to continue on his own branch, @ganisimov  can just grant commit writes to his fork and work can continue there as normal.  Otherwise, you create a new branch off of master in the OHDSI repository and have this PR use that new branch as the base and the merge will happen there. Which would you prefer?

fdefalco:
I think the latter is the better alternative, creating a new branch off of master is have this PR use that new branch as the base.  Would that be alright @ganisimov ? Any preference @anthonysena ?

anthonysena:
Keeping it on a separate branch sounds good to me.

chrisknoll:
Ok, if there's no objections, I can push up a new branch to the OHDSI repo called 'shiro-ui' and then merge this PR onto that branch..

t-abdul-basser:
@chrisknoll makes sense to me :)

fdefalco:
No objections here.

chrisknoll:
Ok, I left the individual commits in the shiro-ui branch but when we finally merge it into master, the individual commits should be squelched first.  The shiro-ui branch is just another dev branch so it can be more noisy than the commits into master.

PR #273, opened 2016-11-03
Fixes #215 and adds further granularity for parent/child by vocabulary

### LINKED ISSUE #215: ATC Hierarchy View
The hierarchy view in ATLAS does not work for ATC classes since this is not defined in the metatrix property within js/app.js::495. We will need to add entries for the ATC hierarchy in order for this to work properly.

anthonysena:
Adding in the ability to see parent/child relationships for ATC 4th and 5th concepts. Also extended the metatrix object to support an optional "vocabulary" property which allows you to specify an array of vocabularies to consider along with the concept relationship and range. This was added to provide the ability to show parent ETC and ATC ancestor concepts in the hierarchy view for a selected RxNorm Ingredient concept. The dependent function hasRelationship was modified to support this optional property.

@fdefalco - I'd like your feedback on the approach since your familiar with this functionality
@pbr6cornell  - Please provide your feedback around the hierarchy configuration I've specified. For RxNorm Ingredient concepts, we originally considered all parent concepts that had a relationship_name of "Has inferred drug class (OMOP)" - this relationship appears to be deprecated in the vocabulary. Now I've specified that for a given RxNorm Ingredient, we'll consider the following as "parent" concepts if they have one of these relationships:

RxNorm to ATC (RxNorm) (any level)
RxNorm to ETC (FDB) (any level)
Has ancestor of - 1st level and only concepts in the ATC and ETC vocabularies

For ATC 5th classes, children will be Rx Norm Ingredient(s) and the parent will be ATC 4th. 
For ATC 4th classes, children will be ATC 5th classes and RxNorm Ingredients and the parents will be ATC 1st-3rd.

### DIFF
--- a/js/app.js
+++ b/js/app.js
@@ -565,6 +565,16 @@ define(['jquery', 'knockout', 'jnj_chart', 'd3', 'ohdsi.util', 'appConfig', 'fac
             }]
 		};
 		self.metatrix = {
+            'ATC.ATC 4th': {                
+				childRelationships: [{
+					name: 'Has descendant of',
+					range: [0, 1]
+                }],
+				parentRelationships: [{
+					name: 'Has ancestor of',
+					range: [0, 5]
+                }]
+            },
 			'ICD9CM.5-dig billing code': {
 				childRelationships: [{
 					name: 'Subsumes',
@@ -604,8 +614,9 @@ define(['jquery', 'knockout', 'jnj_chart', 'd3', 'ohdsi.util', 'appConfig', 'fac
 					range: [0, 999]
                 }],
 				parentRelationships: [{
-					name: 'Has inferred drug class (OMOP)',
-					range: [0, 999]
+                    name: 'Has ancestor of',
+                    vocabulary: ['ATC', 'ETC'],
+                    range: [0, 1]
                 }]
 			},
 			'RxNorm.Brand Name': {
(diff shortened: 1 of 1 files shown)


cgreich:
Friends:

Hang on a second here. Looks like you are double-guessing what's going on in the vocabulary build, and then applying fixes. If you do that, ATLAS will deviate in its behavior from what people see who are not using it. 

ATC5: There is a reason it is not included in the hierarchy: Almost all of them are Ingredients, and making them Standard would violate the rule that we have only one Concept represents any one semantic meaning. If you want to offer the users to find Ingredients for a given ATC5, handle it like NDC. But don't do artificial hierarchy extensions.

"Has inferred drug class (OMOP)" relationship were abolished, because they created havoc: For all relationships between drug classes and drugs, it inferred the same relationship for the Drug Components and Ingredients. In many cases, that is ok, but not for those that are limited to certain routes of administration, or otherwise subsets the selection of drugs beyond ingredients. It completely broke up on us with the introduction of SPLs, because we got a permutational explosion with every drug containing at least one ingredient of an SPL becoming its child.

Instead, we fixed the hierarchy. It no longer is fully executed. In other words, your grandfather (drug class) of you (drug) may not be the father of your father (Ingredient). So, no fiddling around with that is necessary at all.

pbr6cornell:
We should not be making up or inferring new concept relationships or
ancestral relationships in any atlas view.

The 'related concept' view should show all concept relationships and all
concept ancestor connections.

The 'hierarchy' view should only show concept relationships that are
hierarchical.  It should not be trying to traverse multiple hops in the
ancestor table.

If there is use case for a relationship you feel is missing in vocab, that
should be posted as a vocabulary issue.

anthonysena:
Thanks all - I'll leave this pull request open for now but would direct any/all comments to the forums here: http://forums.ohdsi.org/t/vocabulary-hierarchy-exploration-via-atlas/1890. 

We can then decide next steps once we get more input/feedback from the community. Thanks!

anthonysena:
Hi, I'm reviving this pull request to see if we can get it incorporated into master. I've revised the original commit to ensure that the hierarchy is using the "Has ancestor of" and "Has descendant of" relationships respectively. The scope is limited to ATC 4th and RxNorm Ingredients. Hoping that we can incorporate this as a "band aid" until we can think through a design based on the forum discussion here:

http://forums.ohdsi.org/t/vocabulary-navigation-proposal/1895

chrisknoll:
Looks good, I'll merge it.

PR #292, opened 2016-12-07
HTTP verb normalization.

chrisknoll:
This PR normalizes the HTTP verbs in WebAPI such that POST = create, and PUT = update.

This is necessary to support backwards compatibility int he upcoming Shiro branch.

### DIFF
--- a/js/app.js
+++ b/js/app.js
@@ -192,7 +192,7 @@ define(['jquery', 'knockout', 'jnj_chart', 'd3', 'ohdsi.util', 'appConfig', 'fac
 					},
 					'/estimation/:cohortComparisonId:': function (cohortComparisonId) {
 						require(['cohort-comparison-manager', 'cohort-definition-browser', 'components/atlas.cohort-editor', 'cohort-comparison-print-friendly', 'cohort-comparison-r-code'], function () {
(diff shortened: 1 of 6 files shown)


js/app.js
```diff
@@ -192,7 +192,7 @@ define(['jquery', 'knockout', 'jnj_chart', 'd3', 'ohdsi.util', 'appConfig', 'fac
 					},
 					'/estimation/:cohortComparisonId:': function (cohortComparisonId) {
 						require(['cohort-comparison-manager', 'cohort-definition-browser', 'components/atlas.cohort-editor', 'cohort-comparison-print-friendly', 'cohort-comparison-r-code'], function () {
-							self.currentCohortComparisonId(cohortComparisonId);
+							self.currentCohortComparisonId(+cohortComparisonId);
```
chrisknoll: yes, that's correct.  Elsewhere I have to decide on whether to put POST or PUT and i was trying to do the thing:

currentCohortComparisonId() || '' 

Which should return the value if it's not 'falsy' or empty string if it is, but 0 is falsy but "0" is not, so I just make sure that when we put a value in to the observable it's put in as an int.

PR #321, opened 2017-01-31
Create ISSUE_TEMPLATE.md

chrisknoll:
To be used as a template for creation of new issues.


chrisknoll:
I was thinking that I'd put this as a PR to the integration candidate branch, but I think it would be a good idea to get this into master sooner so that we have a template to report issues with.

PR #385, opened 2017-03-22
Fixes #384.

### LINKED ISSUE #384: Concept set name writes right to left
Within the cohort definition and you select concept sets, then new concept set... 

While changing the concept name it enters keystrokes backwards. For instance, here I typed 12345:
image

chrisknoll:
Use contenteditable databind instead of htmlValue.  Chrome 57 breaks the htmlValue binding.


t-abdul-basser:
Good catch @chrisknoll!

PR #436, opened 2017-08-10
allowing for local configuration

fdefalco:
attempting to provide a way to allow local configuration without editing of config.js, which we would like to remain out of the .gitignore as new features / configurations are introduced.
 
config-local.js would be added to .gitignoreafter initial commit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 /web.config
+js/config-local.js
--- a/js/config.js
+++ b/js/config.js
@@ -1,17 +1,15 @@
-define([], function () {
+define(['optional!config-local'], function (localConfig) {
 	var config = {};
 
-	config.services = [
-		{
-			name: 'Local',
-			url: 'http://localhost:8080/WebAPI/'
-          }
-		];
-
-	config.webAPIRoot = config.services[0].url;
-	// config.rServicesHost = 'http://localhost:8081/';
+	// default configuration
+	config.services = [{
+		name: 'Local',
+		url: 'http://localhost:8080/WebAPI/'
+	}];
 	config.cohortComparisonResultsEnabled = false;
 	config.userAuthenticationEnabled = false;
 
+	Object.assign(config, localConfig);
+	config.webAPIRoot = config.services[0].url;
 	return config;
 });
--- a/js/config.js
+++ b/js/config.js
@@ -1,5 +1,8 @@
 define(['optional!config-local'], function (localConfig) {
 	var config = {};
+	if (JSON.stringify(localConfig) == JSON.stringify({})) {
+		console.warn(`Local configuration not found.  Using default values. To use a local configuration and suppress 404 errors, create a file called config-local.js under the /js directory`);
+	}
 
 	// default configuration
 	config.services = [{
--- a/js/main.js
+++ b/js/main.js
@@ -63,6 +63,7 @@ requirejs.config({
 		"bootstrap": "https://maxcdn.bootstrapcdn.com/bootstrap/3.3.2/js/bootstrap.min",
 		"text": "plugins/text",
 		"css": "plugins/css.min",
+		"optional": "plugins/optional",
 		"clipboard": "clipboard.min",
 		"knockout": "knockout.min",
 		"ko.sortable": "https://cdnjs.cloudflare.com/ajax/libs/knockout-sortable/1.1.0/knockout-sortable.min",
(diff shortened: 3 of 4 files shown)


anthonysena:
I've reviewed and having the config-local.js works well for me. Can you add a commit to add this to the .gitignore and push that into this PR? 

As a note, we should also update this wiki page to reflect the new approach for configuration: http://www.ohdsi.org/web/wiki/doku.php?id=documentation:software:atlas:setup

chrisknoll:
Have you looked into the optional plugin:
https://www.npmjs.com/package/requirejs-optional

You can define the config.js file as:
```javascript
// config.js
define([optional!'config-local'], function (configLocal) {
  var config = {};
  Object.assign(config, configLocal); // this will copy all properites from configLocal to config, and if config-local moduleID is not found, configLocal will contain {}
...
});
```

We'll still need a .gitignore value for the local config, but we don't need to make a trick about checking in the file first so that it's in the repo. This way works when the file doesn't exist,

fdefalco:
I've incorporated the optional plugin and made the necessary changes to .gitignore.  Here is a sample of my config-local.js for convenience in testing:

```js
define([], function () {
	var configLocal = {};

        // clearing local storage otherwise your source cache will obscure the override settings
	localStorage.clear();

	configLocal.services = [{
		name: 'Local 4',
		url: 'http://localhost:8080/WebAPI/'
	}];

	return configLocal;
});
```

anthonysena:
The functionality works as expected during my testing whereby when a 'js/config-local.js' file is present based on @fdefalco's example, it overrides the application configuration. The optional plugin also works so that if the file is not present, it continues to use the settings found in 'js/config.js'.

My hesitation with using the 'optional' plugin approach is the fact that running this code without the 'js/config-local.js' present (which will be the default when cloning the repo) will produce the 404 error every time the app loads:

image

So perhaps we would be well served in provide a base 'js/config-local.js' based on Frank's code that is commented out and is also in the .gitignore so that users can reference the approach while also avoiding the error?

chrisknoll:
What bothers me is having to put a file into the repo that is also in the .ignore list.  Seems off.  I can understand the 404's being distracting, but the only way to find out if the optional file is there is you have to ask for it, and then you'll get a 404 if it's not.  That's just a chrome thing, not sure if there's a way to suppress it.  So, my position is that if the error becomes bothersome to me, then I'd drop a empty config file in there that stops the 404.  If you feel strongly the other way, then I defer to your judgement.

anthonysena:
Yeah, I'm with you - I don't like including a file that is then part of the .gitignore. I'm good with providing guidance to let people know that they should include an empty file in their install to avoid this error. Tagging @leeevans since it might be good to get that file added to the OHDSI docker image so we don't see that error when this is eventually deployed out to the community.

chrisknoll:
Not sure why we got a config.js conflict now, but it appears we have a conflict.  @fdefalco, please resolve the conflict. Thanks.

anthonysena:
To share some new-found git knowledge from a session between @chrisknoll and I based on review of this PR and from review of this SO article: https://stackoverflow.com/questions/1274057/how-to-make-git-forget-about-a-file-that-was-tracked-but-is-now-in-gitignore/20241145.

1. If a file is added to a repository and included to the .gitignore, that file's changes will still be tracked by git. So in this case, if we add config-local.js to the repository and add it to the .gitignore, we still face the problem that any local configuration would still be perceived as a change that would need to be committed which seems contrary to our original goal: provide a way to have a local Atlas configuration that is ignored by git and allows for overriding of settings to the main config.js.

2. Git supports the ability to ignore a specific file in the repository using the following command:

     `git update-index --assume-unchanged <file> `

     However, this command only modifies the local repository and there is no way to push it to the repository. Therefore, any/all contributors would need to also run this same command to ignore the file on their local repository. Again, this doesn't feel aligned to our original goal.

So, it looks like the 404s errors that I referenced above are unavoidable. In an effort to try and help, we will provide a console.warn message in the optional.js plugin to attempt to guide people to creating & configuring this module.

chrisknoll:
@anthonysena : a thought:  the warning message can be added in the config.js:

```javascript
define(['optional!config-local'], function (localConfig) {
  var config = {};
  if (JSON.stringify(obj) == JSON.stringify({})) {
    console.warm(`Local configuration not found.  Using default values. To use a local configuration and supress 404 errors, create a file called config-local.js under the /js directory`);
  }

... rest of code here
```

anthonysena:
Thanks @chrisknoll - I like that since we can make the warning more specific to the configuration. I've rebased and force pushed this branch so @fdefalco please remove your local branch before pulling these changes.

chrisknoll:
OOh, no conflicts now, nice :) ty @anthonysena

Issue #454, opened 2017-09-07
Atlas Cohort issue - cohort with certain procedure in the same visit with a certain diagnosis


gowthamrao:
### Expected behavior
Be able to create a cohort that had a Percutaneous coronary intervention (procedure) in the same visit where they also had an inpatient visit with a primary diagnosis non-st myocardial infarction

### Actual behavior
Be able to create a cohort that had a Percutaneous coronary intervention (procedure) x-days before and x-days after an inpatient visit with a primary diagnosis of non-st myocardial infarction

### Steps to reproduce behavior
http://www.ohdsi.org/web/atlas/#/cohortdefinition/922794

condition and procedure must have the same visit_occurrence_id. visit must be inpatient visit.

gowthamrao:
@chrisknoll is current Atlas able to do 'expected behavior' right now?

chrisknoll:
@gowthamrao : no, it's based on the temporal proximity (as you describe in Actual Results).  We could tag the cohort events that are selected with the visit_id that the event came from, and leverage that when looking for the correlated events (at least 1 of X during the same visit...).  My concern here is that visit standardization is sort of a problem, as you know from the discussions about changing the visit structure in the CDM.  It's more consistent when talking across a network of CDMs to say that you're looking at a time context (time is fairly well standardized across all environments :) ).  Everyone agrees that when you say something happened within X days, we're all talking about the same time window.

However, some cases have many different visits within the same hospital due to moving between departments while others consider the stay in the same location the visit.  Since the meaning of a visit can vary, you can't really say consistently 'received procedure at the same visit as the diagnosis'.

gowthamrao:
Correct -- the distinction is the functionality between nested query vs additional event criteria. Both appear to achieve similar functionality.

Could nested query be made to imply -- nested within the same visit record (i.e. join by visit-id) vs 'additional event criteria' to be the temporal?

chrisknoll:
> Could nested query be made to imply -- nested within the same visit record (i.e. join by visit-id) vs 'additional event criteria' to be the temporal?

No, the nested query is just to establish a new index to anchor off for the nested correlated criteria...So that you can do something like:

Had a diagnosis->
  Which had a procedure within 30d after diagnosis
    Which had a follow up visit within 30d after procedure

In cases I've used this, it was explicitly to look beyond the current visit, often finding things that occur at a subsequent visit.  Looking backwards in time has the same effect.

[...]

PR #462, opened 2017-09-21
New front end functionality to restrict to records with the same visit_occurrence_id

gowthamrao:
This PR adds new functionality in Atlas. 

UX to circe-be functionality to restrict qualified-events to the same visit_occurrence_id. _era tables, death, observation period, specimen are excluded from this functionality.

https://github.com/OHDSI/Atlas/issues/454
https://github.com/OHDSI/WebAPI/issues/217
https://github.com/OHDSI/circe-be/pull/7
https://github.com/OHDSI/circe-be/pull/8

@chrisknoll

### DIFF
--- a/js/modules/cohortbuilder/AdditionalCriteria.js
+++ b/js/modules/cohortbuilder/AdditionalCriteria.js
@@ -4,6 +4,7 @@ define(function (require, exports, module) {
 	var CriteriaTypes = require('./CriteriaTypes');		
 	var Occurrence = require('./InputTypes/Occurrence');
 	var Window = require('./InputTypes/Window');
+	var RestrictVisit = require('./InputTypes/RestrictVisit');
 	
 	var debug = false;
 
@@ -15,6 +16,7 @@ define(function (require, exports, module) {
 		self.StartWindow = new Window(data.StartWindow);
 		self.EndWindow = ko.observable(data.EndWindow && new Window(data.EndWindow));
 		self.Occurrence = new Occurrence(data.Occurrence);
+		self.RestrictVisit = new RestrictVisit(data.RestrictVisit);
 	}
 	
 	module.exports = AdditionalCriteria;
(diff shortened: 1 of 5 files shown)


chrisknoll:
Hi, Thanks for the pull request, I think the 'print friendly' also needs to be updated to render the selected visit restriction. I think that you only would print out 'occurring within he same visit'.

See this line:

gowthamrao:
@chrisknoll done!

chrisknoll:
@gowthamrao , thanks for the update, that looks good.  There's one other issue:

In AdditionalCriteria.js, you're defining a new type to store the RestrictedVisit data:

```javascript
var RestrictVisit = require('./InputTypes/RestrictVisit');
```

This type has a property 'RestrictVisit':
```
define(['knockout'], function (ko) {
	
		function RestrictVisit(data) {
				var self = this;
				data = data || {};
				
				self.RestrictVisit = ko.observable(data.RestrictVisit || false);
		}
		
		return RestrictVisit;
});
```

And then when you initalize the RestrictVisit field in AdditionalCriteria.js:

```
self.RestrictVisit = new RestrictVisit(data.RestrictVisit);
```

This leads to a JSON object of the form:
```
{
  StartWindow: {...}
  RestrictVisit: {
    RestrictVisit: false
  }
```

When it should simply be:
```
{
  StartWindow: {...}
  RestrictVisit: false
}
```

Hence, you do not need to make a wrapper type to store th RestrictValue (it's not a complex type). Instead, drop the RestrictVisit module, and just set the RestrictVisit value in AdditionalCriteria like so:

```
self.StartWindow = new Window(data.StartWindow);
self.EndWindow = ko.observable(data.EndWindow && new Window(data.EndWindow));
self.Occurrence = new Occurrence(data.Occurrence);
self.RestrictVisit = ko.observable(data.RestrictVisit || false);
```

-Chris

Issue #414, opened 2017-06-23
End date - end of a visit


gowthamrao:
### Expected behavior
Create a cohort whose cohort start date and cohort end date corresponds to subject's visit_start_date and visit_end_date

### Actual behavior
(default) visit_end_date is the end of observation period (not useful)
(options) censor
-- fixed date or relative number of days after trigger event (not useful)
-- use condition_era or drug_era (not useful)

gowthamrao:
Issue is detailed here OHDSI/WebAPI#234

Issue #407, opened 2017-05-30
Overlapping dates - enhancement request


gowthamrao:
Current cohort outputs may have cohorts where a subjects with multiple records may have overlapping intervals. OHDSI PLP package requires non overlapping periods.

It would be nice to have an option to that rolls up overlapping periods into continuous periods. The SQL syntax would be based on the SQL syntax for condition_era. The option may be an option similar to cohort-exit criteria options or censor options

See attached code based on era-constructor.

[code.txt](https://github.com/OHDSI/Atlas/files/1038351/code.txt)

gowthamrao:
Closed - addressed thru Cohort era-fy option OHDSI/WebAPI#234

Issue #375, opened 2017-03-14
Duplicate rows in Cohort


gowthamrao:
[Request for Atlas UX changes for new Cohort Exit Criteria.pdf](https://github.com/OHDSI/Atlas/files/1218562/Request.for.Atlas.UX.changes.for.new.Cohort.Exit.Criteria.pdf)
In our instance, duplicate rows were produced. This probably can be resolved by using distinct as part of the SQL script. But, we dont know if this is an Atlas issue or how ETL was done into OMOP.

### Expected behavior
Rows should be unique combination of cohort_definition_id, subject_id, cohort_start_date, cohort_end_date

### Actual behavior
Duplicate rows with same data

### Steps to reproduce behavior
http://www.ohdsi.org/web/atlas/#/cohortdefinition/98643

For the above query, visit_occurrence has duplicate rows. During our ETL, we allowed visits with same dates to be different rows i.e. there are multiple rows for the same person with same visit_start_date and visit_end_date. Is this the reason?

chrisknoll:
Yes, that's the reason, but if you are only interested in a single person in the cohort, you can specify 'use earliest event per person' and that will limit the cohort to a single value per person.

However, it does seem reasonable to select the primary events by doing:

```
select person_id, start_date, max(end_date) as end_date
FROM #qualified_events
GROUP BY person_id, start_date
```

This would ensure you get 1 person per start date, but also if you have the same start date, but different end dates, the longest of the end dates will be selected (for example, if you selected a primary event based on a drug class, and drug A and drug B were found that both started on the same day, but drug B had a longer exposure, it would use drug B's exposure as the exposure end date.

If that sounds agreeable, I can implement this.

gowthamrao:
Thank you @chrisknoll . Your suggestion makes sense - longest of the end dates for the same person_id and start_date will be of immense help.

Please implement. 👍

gowthamrao:
@chrisknoll  - Good morning. Thank you for making this suggestion for the feature. Is this something that has been implemented?

chrisknoll:
No, this is still an open issue.

gowthamrao:
Closed - Cohort era-fy option OHDSI/WebAPI#234

PR #463, opened 2017-09-21
UI functionality to choose collapse strategy exit criteria

gowthamrao:
UI functionality only. Additional work is needed to make changes to cohort-constructor logic based on user input.

UI for era-fy 

https://github.com/OHDSI/WebAPI/pull/234
https://github.com/OHDSI/WebAPI/issues/217
https://github.com/OHDSI/Atlas/issues/414
https://github.com/OHDSI/Atlas/issues/375
https://github.com/OHDSI/Atlas/issues/407

@chrisknoll

### DIFF
--- a/js/modules/cohortbuilder/CohortExpression.js
+++ b/js/modules/cohortbuilder/CohortExpression.js
@@ -26,7 +26,7 @@ define(function (require, exports) {
 		self.CensoringCriteria = ko.observableArray(data.CensoringCriteria && data.CensoringCriteria.map(function (criteria) {
 			return CriteriaTypes.GetCriteriaFromObject(criteria, self.ConceptSets);
 		}));
-		self.CollapseSettings = {CollapseType: ko.observable(data.CollapseSettings && data.CollapseSettings.CollapseType || "Era"), EraPad: ko.observable(data.CollapseSettings && data.CollapseSettings.EraPad || 0 ) }
+		self.CollapseSettings = {CollapseType: ko.observable(data.CollapseSettings && data.CollapseSettings.CollapseType || "ERA"), EraPad: ko.observable(data.CollapseSettings && data.CollapseSettings.EraPad || 0 ) }
 		
 	}
 	return CohortExpression;
--- a/js/modules/cohortbuilder/CohortExpression.js
+++ b/js/modules/cohortbuilder/CohortExpression.js
@@ -7,6 +7,7 @@ define(function (require, exports) {
 	var InclusionRule = require('./InclusionRule');
 	var EndStrategies = require('./EndStrategies');
 	var CriteriaTypes = require('./CriteriaTypes');
+
 	
 	function CohortExpression(data) {
 		var self = this;
@@ -26,6 +27,7 @@ define(function (require, exports) {
 		self.CensoringCriteria = ko.observableArray(data.CensoringCriteria && data.CensoringCriteria.map(function (criteria) {
 			return CriteriaTypes.GetCriteriaFromObject(criteria, self.ConceptSets);
 		}));
+		self.CollapseSettings = {CollapseType: ko.observable(data.CollapseSettings && data.CollapseSettings.CollapseType || "Era"), EraPad: ko.observable(data.CollapseSettings && data.CollapseSettings.EraPad || "0") }
 		
 	}
 	return CohortExpression;
--- a/js/modules/cohortbuilder/CohortExpression.js
+++ b/js/modules/cohortbuilder/CohortExpression.js
@@ -7,7 +7,6 @@ define(function (require, exports) {
 	var InclusionRule = require('./InclusionRule');
 	var EndStrategies = require('./EndStrategies');
 	var CriteriaTypes = require('./CriteriaTypes');
-
 	
 	function CohortExpression(data) {
 		var self = this;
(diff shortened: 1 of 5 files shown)


js/modules/cohortbuilder/components/CohortExpressionEditorTemplate.html
```diff
@@ -46,7 +46,7 @@
 					<col span="1" class="rule" />
 					<col span="1" class="delete" />
 				</colgroup>
-				<tbody data-bind="sortable: {data: expression().PrimaryCriteria().CriteriaList, connectClass : 'primaryCriteria', options: {cancel: ':input, button, [contenteditable], .undraggable'}}">
+				<tbody data-bind="sortable: {data: expression().PrimaryCriteria().CriteriaList, connectClass : 'primaryCriteria', options: {cancel: ':input, button, [contenteditable]'}}">
```
chrisknoll: Hi, the 'undraggable' class is used to keep certain elements from being dragged in the ui, but you removed it from the class.  Why?

js/modules/cohortbuilder/components/CohortExpressionEditor.js
```diff
@@ -288,29 +288,31 @@ define(['knockout', 'jquery', '../options', '../CriteriaGroup', '../CriteriaType
...
 
 		self.copyInclusionRule = function (inclusionRule) {
 			var copiedRule = new InclusionRule(ko.toJS(inclusionRule), self.expression().ConceptSets);
 			copiedRule.name("Copy of: " + copiedRule.name());
 			self.expression().InclusionRules.push(copiedRule);
 			self.selectedInclusionRule(copiedRule);
-		};
+		}
+
+		self.addConceptSet = function (item) {}
```
chrisknoll: This function might be unnecessary, I noticed elsewhere in the code that there's an event handler for the .addconceptset selector, but I'm wondering if that selector should be removed vs. adding a empty addConceptSet function...was there an error you were receiving where you needed to add this event handler?

js/modules/cohortbuilder/CohortExpression.js
```diff
@@ -26,6 +27,7 @@ define(function (require, exports) {
 		self.CensoringCriteria = ko.observableArray(data.CensoringCriteria && data.CensoringCriteria.map(function (criteria) {
 			return CriteriaTypes.GetCriteriaFromObject(criteria, self.ConceptSets);
 		}));
+		self.CollapseSettings = {CollapseType: ko.observable(data.CollapseSettings && data.CollapseSettings.CollapseType || "Era"), EraPad: ko.observable(data.CollapseSettings && data.CollapseSettings.EraPad || "0") }
```
chrisknoll: Should this be initialized as numeric?

js/modules/cohortbuilder/components/CohortExpressionEditorTemplate.html
```diff
@@ -99,7 +99,8 @@
 		</div>
 		<div data-bind="eventListener: [
 			 { event: 'click', selector: '.copyInclusionRule', callback: copyInclusionRule},
-			 { event: 'click', selector: '.deleteInclusionRule', callback: deleteInclusionRule}]">
+			 { event: 'click', selector: '.deleteInclusionRule', callback: deleteInclusionRule},
+			 { event: 'click', selector: '.addConceptSet', callback: addConceptSet}]">
```
chrisknoll: This event handler for .addConceptSet selector: is it necessary?  The callback function doesn't actually do anything, so I'm wondering why this was added?

gowthamrao:
@chrisknoll  there was an error on our part that caused those issues. Can you delete this PR and we resubmit a clean new one -- or, can you review the corrections to the original PR we made.

chrisknoll:
Hi, your corrections are fine, when I do the merge we squash it all, so feel free to make additional commits to this branch and it'll be folded into the final PR.  What I do when I review the PR is look at the overall lines changed and such, so your latest commit made the number of lines changed look less, so let me review again and I'll let you know if I see any more problems.

chrisknoll:
Ok, 2 things:
1: I think code needs to be added to the cohortdefinitionviewer module to provide a print-friendly output for the selected collapse value.

2: there is no 'collapse type: none' option on the backed. Only option right now is 'Era'.  We always collapse:

```java
  public String getCollapseConstructorQuery(CollapseSettings collapseSettings) {
		// default constructor is era constructor. as more collapse strategies are introduced, the query template and parameters need to be changed to match.
		String query = ERA_CONSTRUCTOR_TEMPLATE;

		query = StringUtils.replace(query, "@eraGroup", "person_id");
		query = StringUtils.replace(query, "@eraconstructorpad", Integer.toString(collapseSettings.eraPad));
		return query;
  }
```

I wouldn't change anything to support a 'none' option.  We always need to collapse so showing 'none' is not a valid case  The question is how many different ways can we think of to collapse?  Right now, we just have era. We don't have any other option, so we just should put in the UI that we collapse events to eras, and they can specify a gap size between events to chain them together.  So I'd modify the UI to remove the dropdown, and just say 'Specify era collapse gap size: ' or something like that. It defaults to zero, if they leave it blank, it's assumed to be 0d etc.  

3. The value being default in the javascript code is 'Era' but the value should be "ERA".  It's case sensitive, and is based on the enum in java:
```java
public enum CollapseType {
	@JsonProperty("ERA")
	ERA;
}
```

gowthamrao:
@chrisknoll done!

js/modules/cohortbuilder/components/CohortExpressionEditorTemplate.html
```diff
@@ -204,5 +204,20 @@
...
+	<div role="tabpanel" data-bind="css: { active: $component.expressionMode() == 'exit' || $component.expressionMode() == 'all'}" class="tab-pane">
+		<div class="heading">
+			<b>Cohort Collapse Strategy</b>
+			This strategy collapses events into eras of a continous period and eliminates the duplicate rows for the same subjects.  
+		</div>
+		<div style="padding: 5px 0px" data-bind="with: expression().CollapseSettings">
+			<ul>
+				<li>
+					Specify era collapse gap size: 
+					<select data-bind="options: $component.options.dayOptions, value: EraPad"></select> days
```
chrisknoll: The input type for a numeric input isn't a select, rather it's done this way:

```
<span contenteditable="true" class="numericInputField dropdown" data-bind="htmlValue: EraPad.numeric(), eventType: 'blur', ko_autocomplete: { value: EraPad, source: $component.options.dayOptions, minLength: 0, maxShowItems: 10, scroll: true }" />
```

Could you modify the <select> element to be the <span> element that's content editable?

chrisknoll:
I'm going to combine the 2 PRs open for the circe UI changes into one PR, and do the final changes to get this merged in.

PR #464, opened 2017-09-26
PLP Specification Editor

anthonysena:
Atlas changes related to issue OHDSI/WebAPI#224. Must be tested with the WebAPI plp branch submitted through PR: OHDSI/WebAPI#248.


anthonysena:
Sure! I've pushed some commits to clean up the formatting so let me know if you face any problems.

js/makeTable.js
```diff
@@ -0,0 +1,123 @@
+function makeTable() {
```
chrisknoll: This should be wrapped in an AMD module if it's going to be some reusable piece of code.

js/makeTable.js
```diff
@@ -0,0 +1,123 @@
...
+	
+		if(typeof sort_by !== 'undefined'){
+			// if we have a sort_by column, format it according to datatables.
+			sort_by[0] = colnames.indexOf(sort_by[0]); //colname to col idx
+			sort_by = [sort_by]; //wrap it in an array
+		}
+		
+
+	  // Apply DataTable formatting: https://www.datatables.net/
+	  $(document).ready(function() {
```
chrisknoll: I'm curious where we want to use a DOMReady function, shouldn't this be handled by databinding or some other event trigger?

anthonysena:
I've removed makeTable.js as a dependency as it is already embedded into [plp-inspector](https://github.com/OHDSI/Atlas/blob/69c049f4ba5b13d5e16bde278d977357c86dea02/js/components/plp-inspector.html)

PR #488, opened 2017-10-20
Updated plot dimensions for datasources.

chrisknoll:
This PR sets the dimensions for the various plots in the datasources so that the font size and scaling looks consistent.

@t-abdul-basser , @mark-velez : while fixing this up, I noticed there were plots missing from the Measurement reports (recordsByType, etc).  I was going to fix that in this PR, but as I was going through it, I think there's a bug in the WebAPI query for the measurement data, as the recordsByType collection was pulling down more data than expected and the donut chart was not working. So, I left place holders for the plots in this PR, and was hoping you guys would take a look. I will open a separate issue for the Measurement reports.

### DIFF
--- a/js/components/data-sources.html
+++ b/js/components/data-sources.html
@@ -111,7 +111,44 @@
 				<div class="evidenceVisualization" id="ageAtFirstOccurrence"></div>
 				<div class="evidenceDescription"></div>
 			</div>
-		<!-- ko if: $component.currentReport().byFrequency && !$component.loadingReport() -->
+			
+			<!-- ko if: $component.currentReport().byUnit && !$component.loadingReport() -->
+			<div class="row">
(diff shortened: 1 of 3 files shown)


anthonysena:
This PR looks good from my side. One thought I had when reviewing: there are some descriptions of the visualizations on the 'Person' report that we might want to revisit (these were not introduced as part of this PR). For example, the 'Year of Birth' has a description of:

> The number of people in this cohort shown with respect to their year of birth.

The other charts on this report have similar descriptions. I'm thinking we should replace 'cohort' with 'data source'?

chrisknoll:
Good idea, @anthonysena , maybe folks on our internal team could combine minds and come up with a list of those changes (maybe try to leverage some of our 'standard language').

PR #494, opened 2017-10-25
Adjustments to estimation designer and r-code.

anthonysena:
Per feedback from @schuemie, I've made some adjustments to the estimation designer and r-code. I've also commented out references to the nvd3 library as it is not compatible with D3 v4.


fdefalco:
It would be fine to remove the entirety of the nvd3 library references.  I had used them as a stopgap to get the estimation result visualizations into an earlier build but I do not think we'll be reintroducing that any time soon and if we did we could address the visualizations with d3 directly.

anthonysena:
Added another commit to remove the nvd3 references completely. Thanks!

PR #527, opened 2018-01-16
Estimation code fixes

anthonysena:
If multiple outcomes are defined, the currentAnalysisSubset$targetId is a vector which causes the code to break.


anthonysena:
Fixed as part of 59869b24b671dd41ab562399620f2755f79f85c6

PR #541, opened 2018-01-25
Allow change of source_daimon priorities via UI

mpozhidaeva:
Issue
https://github.com/OHDSI/WebAPI/issues/305


anthonysena:
This change along with OHDSI/WebAPI#306 attempts to allow a user to set a daimon priority via Atlas and have it persist the change in the source_daimon table.

I've tested this under 2 different circumstances:

1. When authentication/authorization are turned off, Atlas will allow you to change the priority of a daimon and that change will persist in the database. This changes existing behavior where a change to the priority of a daimon is reflected on the client-side and persists only for that session. When we are in a situation where security is disabled, we should continue to only allow the priority change to last during the session otherwise users could change the vocabulary priority on one another.

2. When authentication/authorization are turned on, Atlas will prevent you from changing the priority of a daimon all-together. Once you have signed in and have obtained the 'admin' role, the UI will then allow you to change the priority of a daimon and persist this change via WebAPI. Again, in the situation where a person is not an administrator, they may want/need to change a daimon's priority.

I think overall the idea of having a means for changing and persisting a daimon's priority is good and needed. I'm just not sure if this implementation is exactly what we need considering the points above.

Tagging @fdefalco, @chrisknoll, @pavgra for thoughts on this item.

chrisknoll:
@anthonysena : am I understanding the issue correctly:

There's a distinction between the WebAPI setting of 'default priority' (which is saved in the DB and used to initialize atlas) vs. a user preference of a priority where the user can come in and change the source to prioritize (for vocabulary searches, record counts etc) which is on a per-session basis.

I haven't looked at the implementation, but is the problem that we make the 'Configuration' module an admin screen where restricting changes to an admin role means users can't set their preference anymore?

If so, maybe the solution is to allow anyone to change the preference, but only admins will get the button to send the current preferences back to the server as the 'webapi default priorities' ?

anthonysena:
Thanks @chrisknoll - that's exactly right. Right now the Atlas UI has no way to set the priority of a daimon - this must be done directly on the WebAPI database. This implementation provides a way to change the daimon priority from the front-end which would be useful for administrators. The downside in the implementation in this PR is that we lose the ability for a user to select the priority during their session. This is particularly useful when you would like to view the record count (RC) and descendant record counts (DRC) during vocabulary searches. So, I'd hate to lose the ability to control that priority on a per-user basis.

I like the idea of having a per-user configuration that is separate from the overall system administration so that we can support both needs via the interface. This PR is good in that we could consider creating a "Admin" menu item and then provide a mechanism for managing roles, sources, daimon priority, etc in a single place. The 'Configuration' screen could be a place where a user can configure Atlas on a per-user basis.

anthonysena:
@mpozhidaeva - Can we also merge this into #561 for testing since it has a dependency on security? Furthermore, can we update the functionality per our discussion on OHDSI/WebAPI#306 so that when security is disabled, a user can still change their daimon without it posting the information to the server? Thanks!

pavgra:
Was closed as part of https://github.com/OHDSI/Atlas/pull/622

PR #559, opened 2018-02-01
Added Censor Window settings block into Cohort Definition Builder

pavgra:
For https://github.com/OHDSI/circe-be/issues/26

### DIFF
--- a/js/modules/cohortbuilder/CohortExpression.js
+++ b/js/modules/cohortbuilder/CohortExpression.js
@@ -7,6 +7,7 @@ define(function (require, exports) {
 	var InclusionRule = require('./InclusionRule');
 	var EndStrategies = require('./EndStrategies');
 	var CriteriaTypes = require('./CriteriaTypes');
+	var Period = require('./InputTypes/Period');
 	
 	function CohortExpression(data) {
 		var self = this;
@@ -27,7 +28,10 @@ define(function (require, exports) {
 			return CriteriaTypes.GetCriteriaFromObject(criteria, self.ConceptSets);
 		}));
 		self.CollapseSettings = {CollapseType: ko.observable(data.CollapseSettings && data.CollapseSettings.CollapseType || "ERA"), EraPad: ko.observable(data.CollapseSettings && data.CollapseSettings.EraPad || 0 ) }
-		
+		self.CensorWindow = ko.observable(new Period({
+			StartDate: data.CensorWindow ? data.CensorWindow.StartDate : null,
+			EndDate: data.CensorWindow ? data.CensorWindow.EndDate : null,
+		}));
 	}
 	return CohortExpression;
 });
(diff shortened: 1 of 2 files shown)


js/modules/cohortbuilder/CohortExpression.js
```diff
@@ -27,7 +28,10 @@ define(function (require, exports) {
 			return CriteriaTypes.GetCriteriaFromObject(criteria, self.ConceptSets);
 		}));
 		self.CollapseSettings = {CollapseType: ko.observable(data.CollapseSettings && data.CollapseSettings.CollapseType || "ERA"), EraPad: ko.observable(data.CollapseSettings && data.CollapseSettings.EraPad || 0 ) }
-		
+		self.CensorWindow = ko.observable(new Period({
+			StartDate: data.CensorWindow ? data.CensorWindow.StartDate : null,
+			EndDate: data.CensorWindow ? data.CensorWindow.EndDate : null,
+		}));
```
chrisknoll: You don't need to construct the JSON in the new Period() call. You can pass the element directly from the raw data:

```
  self.CensorWindow = ko.observable(data.CensorWindow && new Period(data.CensorWindow));
```
This will only allocate a Period object if CensorWindow is not null.

js/modules/cohortbuilder/components/CohortExpressionEditorTemplate.html
```diff
@@ -220,4 +220,18 @@
...
+			Drops all persons, who do no fall inside the censor window.
+		</div>
+		<div style="padding: 5px 0px">
+			<ul>
+				<li>
+					<period-input params="Period: expression().CensorWindow"></period-input>
+				</li>
+			</ul>
+		</div>
+	</div>
```
chrisknoll: This looks like it always is always showing the censor window:  I think we'd want to have a button that when CensorWindow == null it shows and says 'Set Censor Window' and clicking it adds a new Period() to the CensorWindow.  (which will hide the button and show the UI).  Elsewhere, there would be a 'Remove censor window' which would set CensorWindow  to null, which would update the UI to show the 'Add Censor Window button' etc.

However, @pavgra , the main problem is that you're working off of a out of date version of the CohortExpressionEditor component. This was drastically redone in the 2.3 branch.  @anthonysena  or @fdefalco should give you guidance on how you should apply those changes to that feature branch.

pavgra: @chrisknoll, so which branch should I be branching off?

chrisknoll: @anthonysena or @fdefalco  should be able to provide you with that.

anthonysena:
Hi @pavgra - I'd suggest you pull the master branch to get the latest v2.3 work and rebase these changes on that work. Then I'll take a look at this and we can get it merged into the code base.

PR #567, opened 2018-02-13
Enable Heracles Heel for full analysis by default

anthonysena:


chrisknoll:
When running full analysis, I can't seem to get the other reports to show (including heel), even after refreshing the browser.

Issue #573, opened 2018-02-20
Remove XSS vulnerability


johnSamilin:
### Expected behavior
Data returned from the server is sanitized before injecting in HTML
### Actual behavior
Raw data is being dangerously injected in HTML causing XSS possibilities
### Steps to reproduce behavior
Create Conept Set/Cohort with the name that contains <script> tag

PR #574, opened 2018-02-20
Modified bindings to filter out XSS

johnSamilin:
https://github.com/OHDSI/Atlas/issues/573


chrisknoll:
Hi, @johnSamilin . I've been following this PR closely, and I'd like to address the XSS issues as part of a post 2.3 release (since we're getting close to the 2/28 release of 2.3, I think we can do something with XSS in a 2.3.1 release).

I wanted to know your thoughts about dealing with the XSS issues as a server-side solution.  I like how you incorporated the xss filtering within the knockout bindings as  a way to 'inject' the sanatizer into the UI.  However, this is only a client-side solution, and there's other ways to POST/PUT data to the WebAPI REST endpoint that has XSS vulnerabilities.   Therefore, I think the ultimate solution should be on the server side using a java implementation to sanitize user input, perhaps something like described here: https://stackoverflow.com/questions/3413297/how-to-sanitize-html-code-in-java-to-prevent-xss-attacks.  My feeling is: as long as we sanitize prior to writing into a database, you won't inject any malicious script/html into other clients by reading from the database.  So in this way, it's possible client-based XSS may not be even necessary as long as the server prevents all injection paths.

Thoughts?

pavgra:
@chrisknoll, although sanitizing of data at server side sounds as reasonable action, it has some drawbacks:
- we change original data passed into system, which can not always be applicable
- there is no single place to apply such sanitizer at back-end side, so massive changes would be required

Whereas anyway front-end filtering for XSS is required - all modern frameworks contain such feature out-of-the-box (React, Vue, Angular), and XSS itself is front-end issue, not back-end one.

js/main.js
```diff
@@ -168,16 +171,24 @@ requirejs.config({
...
 	}
 });
 
 requirejs(['bootstrap'], function () { // bootstrap must come first
-	requirejs(['knockout', 'app', 'appConfig', 'webapi/AuthAPI', 'ohdsi.util', 'lscache', 'atlas-state', 'vocabularyprovider', 'director', 'search', 'localStorageExtender', 'jquery.ui.autocomplete.scroll', 'loading', 'user-bar', 'welcome'], function (ko, app, config, authApi, util, lscache, sharedState, vocabAPI) {
+	requirejs(['knockout', 'app', 'appConfig', 'webapi/AuthAPI', 'ohdsi.util', 'lscache', 'atlas-state', 'vocabularyprovider', 'xss', 'director', 'search', 'localStorageExtender', 'jquery.ui.autocomplete.scroll', 'loading', 'user-bar', 'welcome'], function (ko, app, config, authApi, util, lscache, sharedState, vocabAPI, filterXSS) {
 		var pageModel = new app();
 		window.pageModel = pageModel;
+
+		const originalTextBinding = ko.bindingHandlers.text.update;
```
chrisknoll: Question about this binding override:  according to the documentation about the text binding (http://knockoutjs.com/documentation/text-binding.html) it says:

```
Note 2: About HTML encoding
Since this binding sets your text value using a text node, it’s safe to set any string value without risking HTML or script injection. For example, if you wrote:

viewModel.myMessage("<i>Hello, world!</i>");

… this would not render as italic text, but would render as literal text with visible angle brackets.

```

So why sanitize the text binding if it is already safe?

chrisknoll:
I think the main XSS exposure is related to our datatable binding (that you are addressing in the update to the custom binding).  Details of how to mitigate that is found here:
https://datatables.net/manual/security and https://datatables.net/forums/discussion/32150/xss-prevention-from-unsanitized-server-input

I think what you've proposed is fine, i need to do a little testing on it, just to understand if we need to override the render method or just ensure that our template bindings in the datatable options are not using anything unsafe.  it is _certainly_ a problem if we just bind a user input field to a column value of the dataTable (without using the safe-render functions described  above or using XSS filtering as this PR does).  I'm wondering if we need to sanitize the data or just at render time?

johnSamilin:
Hi @chrisknoll,
you were absolutely right, 'text' binding is not really necessary. I tried to reproduce the issue without it and it turns out 'table' binding is enough to fix it

chrisknoll:
Hi, @johnSamilin , I'm sorry for the delay but have been wrapped up in the 2.3 release. I can review this and should have the approval by tomorrow.  Thanks for your patience!

chrisknoll:
@johnSamilin : i'm goign to merge this into a local branch because there's a change that your fork didn't have with the datatables binding.  It's good that you removed the redundant code, but the problem is the binding in the package didn't get the fix that was applied in the redundant module.  I'll merge this into a local branch, make ethe fix and then merge into master.

Issue #583, opened 2018-02-28
No lockout policy


wivern:
It was found that the application did not enforce an account lockout policy. This could expose user accounts to password guessing attacks, which could lead to an account compromise.
Account lockout is a security measure by which a user account is prevented from logging in temporarily, or until unlocked by an administrator, after an incorrect password is supplied a given number of times, in order to prevent attackers from guessing the user’s password through brute-force attacks.
The application was found to not make use of the account lockout control, making it vulnerable to dictionary based brute-force password guessing attacks, which could lead to an account compromise.
It is normally trivial to automate the process of differentiating between a successful and failed login attempt.
Therefore, it is possible for an attacker to write a script to continuously attempt to log in with many different passwords and report on any that were successful. This technique can be either dictionary based or incremental.
An account lockout mechanism should be implemented so as to provide an extra layer of defense against password guessing attacks. Additionally, it is important to ensure that the lock out is not permanent; otherwise it could be used to perform DoS attacks against the system by purposely locking out valid accounts. There are a number of different ways in which this concern may be addressed, for example:
• A user may be required to contact an administrator to reactivate their account.
• The account may be re-enabled after a fixed amount of time.
• A user may be able to re-enable their account by confirming their identity on a predefined out-of-band channel, such as their email.

PR #584, opened 2018-02-28
No lockout policy

wivern:
for #583


chrisknoll:
This PR looks like it's mixing other PRs (I see the XSS changes in here), and the commits look like they go all the way back to the jnjchart library update?

For the sake of speed and organization, could the odysseus group look into using topic branches and "git rebase" before submitting pull requests.  The ideal pull request should look like just a few commits that cover the exact change that's being proposed in the PR.

Thank you for your consideration.

PR #589, opened 2018-03-06
XSS Prevention in Data Tables

chrisknoll:
modified bindings to filter out XSS and whitelist elements/attributes
Removed obsolete templateDatatable binding


chrisknoll:
@johnSamilin : if you could give this a review, I'll merge it into atlas.  This is effectively your original PR, however, there was an out of date knockout binding in your PR (you chose to delete one, keep the other, but the one you deleted had some updates) Your instinct was correct: you should have kept the one under databindings/ but there was another commit that introduced that confusion.  But this PR will correct all those.

In addition: there are some cases in our datatables where we put buttons into the cells. I had to add XSS options to put "button" in the whitelist (and only allow type and class attributes).

Issue #537, opened 2018-01-23
Evidence negative control generation hangs (v2.3)


anthonysena:
### Expected behavior

Specifying a large list of concepts to include/exclude as part of the evidence processing and the output should indicate that these are user included/excluded concepts.

### Actual behavior

Specifying a large list of concepts to include/exclude as part of the evidence processing. Generate results  but the processing never completes.

### Steps to reproduce behavior

See above.

ericaVoss:
Example exclude list 

```{TEXT}
194913,372888,373470,374032,374925,375260,375806,376112,376395,376686,378140,378420,380394,381551,4027396,4049810,40480668,43531027,43531028,
255841,257905,258780,261889,4050731,4112826,4137525,4138392,4166517,4172303,4230358,4236366,4309350,40481763,46270376,
[...]

PR #590, opened 2018-03-14
Use Common Evidence Model for negative controls

anthonysena:
Changes related to OHDSI/WebAPI#345


anthonysena:
This work addresses the bug with #537. This needs to be tested before closing the issue.

PR #598, opened 2018-03-20
PLE & PLP execution via Arachne Exec Engine

agackovka:


fdefalco:
I have execution engine running locally and I am going through testing.  Can @pavgra or @agackovka please resolve the conflicts to master?

agackovka:
@fdefalco merged

fdefalco:
Apologies in advance for dumping everything here but we can break out what needs to be addressed as individual issues; for now I'm going to communicate my experience with testing this PR.

I was able to configure my environment to test execution engine and have successfully connected to the execution engine and have the execution engine responding back to the WebAPI.  

The current documentation is here: https://wiki.arachnenetwork.com/display/DOC/Deployment and 
here: https://wiki.arachnenetwork.com/display/DOC/Integration+with+Execution+Engine. @pavgra  has suggested that we move this to the OHDSI wiki and I agree that would be welcome.

I'm testing in a MacOS environment.  We should include in the documentation that for a mac environment changes are required.

The settings in the maven profile for executionengine.update_status_callback and executionengine.result_callback need to use host.docker.internal instead of localhost, for example, because of how MacOS uses a hypervisor container around docker.  This is just a MacOS specific setting.

More generally, I would prefer if we changed these settings to not use snake case.

1. We need to document that ATLAS needs to be configured with:
 	configLocal.useExecutionEngine = true;

2. When ATLAS loads and it finds that useExecutionEngine = true it should attempt to connect to the execution engine.  I'm not sure if there is a call to the execution engine that could respond with its version and availability but when things fail we currently don't provide enough information on what has failed.  (connectivity to execution engine, connectivity from execution engine to webapi, etc) . I recommend having an indicator, on the configuration screen for now, that displays whether or not the execution engine is available and what version it is currently running.  

3. configLocal.plpResultsEnabled = false enables the PLP results tab, but also the inspector and perfomance tab, neither of which should be displayed.

4. PLP manager is current broken in this branch:
image

5. The WebAPI methods associated with execution engine are using snake case and I would prefer the naming stayed within the conventions of the rest of the WebAPI route names.

6. Execution of jobs through the execution engine should register with the ATLAS notification system.

7. When jobs fail and communicate back to WebAPI sometimes the job itself remains in 'Started' state although the UI on the estimation result screen knows that the execution has failed.  I'm not sure why this is happening.

pavgra:
@fdefalco, great thanks for the valuable and detailed feedback - we'll work on it

agackovka:
@fdefalco done

anthonysena:
@pavgra Could you or @agackovka resolve the merge conflicts and pull this into master? Thanks!

PR #607, opened 2018-03-22
Added LESS loader. Implemented Filter panel component

pavgra:


js/components/visualizations/filter-panel/filter-panel.js
```diff
@@ -0,0 +1,22 @@
+define(
+  [
+    'knockout',
+    'text!components/visualizations/filter-panel/filter-panel.html',
```
chrisknoll: should this use relative paths (text!./filter-panel.html)?

Also, does the less plugin support relative paths?  It should...

js/app.js
```diff
@@ -184,6 +184,11 @@ define(['jquery', 'knockout', 'ohdsi.util', 'appConfig', 'webapi/AuthAPI', 'atla
 							document.location = "#/welcome";
 						});
 					},
+          '/test-page': function () {
+            require(['test-page'], function () {
```
chrisknoll: Let's make a new foluder under /js called 'tests' and move our test components there. Then you can reference this componet as 'tests/test-page'

We should update all of our 'component' module refercts from 'moduleA' to 'components/moduleA' and then we can clean up all of the extra stuff in our requirejs config.

js/components/visualizations/table-baseline-exposure/table-baseline-exposure.js
```diff
@@ -0,0 +1,62 @@
+define(
```
chrisknoll: Let's discuss the organization of our modules/folder structure

chrisknoll:
@pavgra , 

On the note of modules and their folder structure:  I think you saw how the moduleIds get redundant when you have a one-module-per-folder setup, like we do here:

```
/components/reports/cost-utilization-persons-exposure/cost-utilization-persons-exposure.js
```

To load that we have a module id of: 'reports/cost-utilization-persons-exposure/cost-utlization-persons-exposure'

What if we had a different folder structure of /healthcareutlization/reports/  and we put all of our report components under thre, if we have different types of modules we can put it under /healthcareutilization/{other}

Then we can load our repot moduels as 'healthcareutlization/reports/cost-utilization-persons-exposure'.  we can put the visit reports under the same folder...

If we want to split the fodlers up we could have a root fo healthcareutlization/ and subfolders of /exposure, /visits and put the componetns for each report there, and load it as /healthcareutilication/exposure/exposure-report', but my feeling is that havign such fine-grained folders just makes modules longer without reason, I think i like the idea of loading the requried report as '/healthcareutlization/reports/exposure' and '/healthcareutlization/reports/visit' and the foder healthcareutilizaton/report has the files exposure.js{html] and vsit.js[html].  It's fine to have a longer name for the module in the report folder if you prefer, i was just making it short for simplicity.

I think there's a way to leverage the 'package' configuration to also help us here, i'd have to think on it.


Last thought:  I've also read the technique of if you want to load a module like modulePackage/subModule but still maintain separate folders for the submodules, then you create JS called subModule.js which loads the other components in the child folder 'subModule'. The folder structure looks like this:
```
moduleNamespace/
  module1/
     mainModule.js
     someMoreModulesForModule1.js
  module2/
    mainModule.js
    someMoreModuelsForModule2.js
  module1.js  // this module depends on ./module1/mainModule and someMoremodulesForModule1
  module2.js // this module depends on stuff from ./module2
```

With this structure we can now require modle1 and module 2 like:

```
require(['moduleNamespace/module1, 'moduleNamespace/module2'], function (module1, module2) {
  ... use the modules
}
```

When you want to add another module under moduleNamespace, you just add the module3.js under the package folder, and create a subfolder for you to house the components required by the module3

Hope this makes sense! But I'd like to avoid creating modules that resolve like: someLongModuleId/someLongModuleId

Issue #489, opened 2017-10-20
Measurement Reports are missing plots


chrisknoll:
### Expected behavior
Measurement Drilldown should have Records by Unit, Measurement Value Distribution, Lower Level Distribtuion, Upper Level distribution, and values relative to norm.

### Actual behavior
These plots are missing from the datasoucres report, but are available in AchillesWeb.

### Steps to reproduce behavior

Go To datasources,
Pick a source and the measurement report
click on any measurement
Notice 5 plots are missing from the report.

Example:
![image](https://user-images.githubusercontent.com/6818777/31833110-8625542e-b597-11e7-9ec7-135eca5fe256.png)

Note: this plot is not seen on the ohdsi.org achilles demo, since it appears the achillesWeb version is not up to date.  But if you use the current AchillesWeb in your own enviornment, you will see the proper measurement drilldown.

t-abdul-basser:
@chrisknoll Thanks.  To clarify, which reports are missing?

chrisknoll:
The 5 plots above are missing from the Measurement Report.

chrisknoll:
The queries involved can be found in Achilles exportToJson:
https://github.com/OHDSI/Achilles/blob/master/R/exportToJson.R#L1708

The missing reports start here:
https://github.com/OHDSI/Achilles/blob/master/R/exportToJson.R#L1773

I attempted to plug in these reports as part of the fix for #488 , however the report data that returned from the measurement drilldown query was 1.4MB json file which seemd to have measurment statistics for all measurements instead of the measurement that was selected in the treemap.  So it is likely that you have the queries coded, but they are just bugged, but also that the results of the query are not presented in the drilldown of measurements.

After #488 is approved, you will see the placeholders I added in the measurement report that need to be implemented using the results from WebAPI.  But to fix this completely, webAPI likely needs to be updated with the correct measurement drilldown queries.

vojtechhuser:
any update on this?

PR #610, opened 2018-03-23
Measurement Reports are missing plots

wivern:
for #489


chrisknoll:
@wivern : i'm not sure if this is a WebAPI or Atlas side issue, but the plots for 'Upper Limit Distibution' and 'Values Relative to Normal Range' are not producing any plots.  Can you confirm?

I checked with this concept in our internal environment: Cholesterol in HDL [Mass/volume] in Serum or Plasma.

Here's the Atlas and Achilles output:
atlasmeasurementreport
Achilles:
achillesreport

PR #622, opened 2018-03-30
User Interface for Daimon / Source Configuration

### LINKED ISSUE #57: User Interface for Daimon / Source Configuration
As per @leeevans

wivern:
closes #57

### DIFF
--- a/js/components/configuration.html
+++ b/js/components/configuration.html
@@ -84,9 +84,9 @@
 					</div>
 				</div>
 			</div>
-			<div data-bind="if: config.userAuthenticationEnabled && !$component.hasAccess">
-				<access-denied params="isAuthenticated: isAuthenticated"></access-denied>
-			</div>
 		</div>
 	</div>
+</div>
+<div data-bind="ifnot:$component.hasAccess">
+	<access-denied params="isAuthenticated: isAuthenticated"></access-denied>
 </div>
--- a/js/components/configuration.html
+++ b/js/components/configuration.html
@@ -22,7 +22,7 @@
 						<div class="padded priority" data-bind="if:$component.canChangePriority">
 							<div class="priority-label">Change source priorities in: </div><select class="form-control" style="margin-bottom: 0" data-bind="options: $component.priorityOptions, optionsText:'name', optionsValue:'id', value: $component.sharedState.priorityScope" />
 						</div>
-						<span class="btn btn-sm btn-primary" data-bind="click: $component.newSource, css: {disabled: !$component.canCreateSource()}">New Source</span>
+						<span class="btn btn-sm btn-primary" data-bind="click: $component.newSource, visible: $component.config.userAuthenticationEnabled, css: {disabled: !$component.canCreateSource()}">New Source</span>
 					</div>
 				</div>
 				<div data-bind="if:available==false">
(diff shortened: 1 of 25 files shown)


js/components/source-manager.js
```diff
@@ -0,0 +1,179 @@
...
+    Vocabulary: { tableQualifier: '', enabled: false, priority: 0, sourceDaimonId: null },
+    Results: { tableQualifier: '', enabled: false, priority: 0, sourceDaimonId: null },
+    Evidence: { tableQualifier: '', enabled: false, priority: 0, sourceDaimonId: null },
+  };
+
+  function Source(data) {
+
+    function mapDaimons(daimons) {
+      daimons = daimons || [];
+      var defaultKeys = lodash.keys(defaultDaimons);
```
pavgra: There is native Object.keys with enough browser support. Why not to use it?

js/styles/atlas.css
```diff
@@ -2095,4 +2095,14 @@ div#wrapper_ohdsi img {
...
+}
+
+.form input[type=text] {
+  width: 100%;
+}
+
+.form label {
+  font-weight: bold;
+  font-size: 1.2rem;
+  margin-left: 5px;
```
pavgra: We've discussed required styling changes in person

js/components/source-manager.js
```diff
@@ -0,0 +1,178 @@
...
+        name: self.selectedSource().name() || null,
+        key: self.selectedSource().key() || null,
+        dialect: self.selectedSource().dialect() || null,
+        connectionString: self.selectedSource().connectionString() || null,
+        daimons: ko.toJS(self.selectedSource().daimons()).filter(function(d){ return d.enabled; }).map(function(d){
+          return lodash.omit(d, ['enabled']);
+        }),
+      };
+      sourceApi.saveSource(self.selectedSourceKey(), source).then(function(){
+        return sourceApi.initSourcesConfig();
```
anthonysena: Is this call to init all of the sources necessary or could we process the new data source individually? In our environment with many data sources configured, the save seems to take a very long time due to this call. Alternatively, we could lock the screen while it is saving so that users cannot re-submit the same information.

js/components/configuration.html
```diff
@@ -29,27 +39,27 @@
...
 						<th>Incidence</th>
 					</thead>
-					<tbody data-bind="foreach:$data.sources">
+					<tbody data-bind="foreach:$component.sources()">
 						<tr>
 							<td><i class="fa" data-bind="if:!initialized, attr: { title : error }, css: { 'fa-check-circle': initialized, 'fa-exclamation-circle': !initialized }"></i>
 								<span data-bind="text:error"></span>
 							</td>
-							<td data-bind="text:sourceName + ' [' + sourceKey + ']'"></td>
+							<td><a href="#" data-bind="text:sourceName + ' [' + sourceKey + ']', click: $component.selectSource"></a></td>
```
anthonysena: The source names on the configuration screen should only have a link when the user has rights to view or edit the source details. Furthermore, we should consider using the source_id and not the source_key here for consistency with the way we're doing things in the rest of the app.

js/components/configuration.js
```diff
@@ -1,21 +1,89 @@
...
+        contentType: 'application/json',
+        success: function () {
+          sourceApi.initSourcesConfig();
+        },
+        error: function (err) {
+          self.isInProgress(false);
+          alert('Failed to update priority source daimon');
+        },
+        complete: function () {
+          self.isInProgress(false);
```
pavgra: From jQuery doc - "A function to be called when the request finishes (after success and error callbacks are executed)"

So no need to call same line in error callback

js/components/configuration.html
```diff
@@ -58,16 +71,25 @@
...
 				</table>
 			</div>
-
+			<div data-bind="if: isInProgress()" class="text-center">
+					<loading params="theme: 'light'" />
+			</div>
 			<div>
-				<div class="padded">
-					<div class="btn btn-sm btn-primary" data-bind="click: clearLocalStorageCache">Clear Configuration Cache</div>
+				<div style="display: flex">
```
pavgra: no css inlining pls

PR #640, opened 2018-04-11
Updating Concept Sets in a Cohort by Swapping for Another

### LINKED ISSUE #365: Updating Concept Sets in a Cohort by Swapping for Another
This is an enhancement request not a bug.  

It would be nice to be able to not only delete a concept set in a cohort but also be able to swap it out for another concept set.

Example, you defined one concept set and it is used multiple times in a cohort.  You learn there is a better definition for the concept set.  It would be nice to just replace it with another concept set already built in ATLAS.

wivern:
closes #365

### DIFF
--- a/js/components/atlas.cohort-editor.html
+++ b/js/components/atlas.cohort-editor.html
@@ -10,12 +10,12 @@
 		<!-- ko ifnot:canEdit-->
 		<cohort-expression-viewer params="expression: model.currentCohortDefinition().expression"></cohort-expression-viewer>
 		<!-- /ko -->
-		<div class="modal fade" id="conceptSetSelectorDialog" tabindex="-1" role="dialog">
+		<div class="modal fade" data-bind="modal: showModal" tabindex="-1" role="dialog">
 			<div class="modal-dialog modal-lg">
 				<div class="modal-content">
 					<div class="modal-header">Import Concept Set From Repository...</div>
 					<div class="paddedWrapper">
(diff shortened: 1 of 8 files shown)


js/components/atlas.cohort-editor.html
```diff
@@ -10,12 +10,12 @@
 		<!-- ko ifnot:canEdit-->
 		<cohort-expression-viewer params="expression: model.currentCohortDefinition().expression"></cohort-expression-viewer>
 		<!-- /ko -->
-		<div class="modal fade" id="conceptSetSelectorDialog" tabindex="-1" role="dialog">
+		<div class="modal fade" data-bind="modal: showModal" tabindex="-1" role="dialog">
```
chrisknoll: Thank you so much for ridding the code of element IDs! and using the bindings!! (big heart) (big hug)

anthonysena:
Thanks @wivern for this contribution. In reviewing the PR, I think their is either a bug or a gap in the way we described the need for #365. For purposes of illustration, I'm going to include some screen grabs and tag @ericaVoss for input. To start, use a cohort definition with a concept set that is already defined, navigate to the Concept Sets tab, select a concept set to view the details and then the "Import" tab:

image

The new "Repository" pill tab then allows the user to navigate to concept sets in the repository to import. From my testing, this imports the concept set from the repository and adds the concepts to the existing concept set in the cohort definition. What was desired per #365 was to completely replace the concept set definition that is in the cohort definition with the one imported from the repository. 

I'd also add that when importing the concept set from the repository, we should replace the concept set expression and rename the concept set using the name in the repository so it is clear that the concept set has changed. @ericaVoss your thoughts on this?

PR #646, opened 2018-04-16
closes #645

### LINKED ISSUE #645: Check visual appearence of the charts
### Expected behavior
All axis titles are the same size, have equal margins, labels have format "month/year" instead of "year"

johnSamilin:


chrisknoll:
Those pie charts: they look like they are in the proper dimension for the space in the div, but for some reason the SVG elemetns in the DOM are being skewed all the way to the right...is it possible there's some sort of styling being applied to those SVG elements to make them have a left-margin like that? Something that bootstrap may be doing?  I wonder if you inspect the pie-slice element in the middle chart if you get a hint as to if it's a positioning problem or some style is messign with the layout.

johnSamilin:
@chrisknoll @anthonysena thanks for the reivew, those issues are fixed

js/main.js
```diff
@@ -122,7 +122,7 @@ requirejs.config({
 		"cohortcomparison": "modules/cohortcomparison",
 		"r-manager": "components/r-manager",
 		"negative-controls": "components/negative-controls",
-		"atlascharts": "https://unpkg.com/@ohdsi/atlascharts@1.1.0/dist/atlascharts.min",
+		"atlascharts": "https://unpkg.com/@ohdsi/atlascharts@1.2.0/dist/atlascharts.min",
```
anthonysena: Tagging @chrisknoll - The scaling issue we were discussing offline appears to be related to bumping up the atlas charts library here. When I revert back to 1.1.0, I no longer experience the problem where the chart doesn't fill the container. Looking at the [changes between the library versions ](https://github.com/OHDSI/Visualizations/compare/v1.1.0...v1.2.0#diff-5a982a23c07be42ee2b02d0126e84715R66) it appears that a width/height were added to the base chart object - could this be the reason for the lack of scaling?

johnSamilin: I think this is the reason because of the preserveAspectRatio attribute of the SVG. It's current value means that there have to present height and width attributes. Which charts besides treemap scaling incorrectly?

anthonysena: @johnSamilin If you use this branch and take a look at the data density report, the line chart remains a fixed size even on larger resolutions. If you roll back the atlascharts library from 1.2.0 to 1.1.0 you will see that it then scales to fit the containing panel. Thanks!

johnSamilin: @anthonysena it seems like the problem is that in version 1.1.0 render method accepted width and height parameters but ingored them. It's not really correct behaviour so we should either stop passing those parameters and let atlascharts get bounding box of it's container or determine it before the render method just like I did with treemaps in my second commit. I'd suggest the latter way

anthonysena: @johnSamilin thanks for investigating this. I think applying the treemap approach as you proposed would be great. Could you add this to the branch and we'll review? Thanks!

chrisknoll: @johnSamilin : why do you say we ignore them?  Here's the code in the 1.1.0 release:
https://github.com/OHDSI/Visualizations/blob/v1.1.0/atlascharts/line.js#L38 and where it's used:
https://github.com/OHDSI/Visualizations/blob/v1.1.0/atlascharts/line.js#L55

The idea is that you need to specify the resolution of the SVG context to get the appropriate proportion of scale (for the initial rendering of fonts) and then the svg scales larger and smaller based on how the viewport is resized.

This is why  if you make the width and height smaller, it makes the SVG text look bigger (it's making the same font appear larger because it looks bigger at 600x300 vs 1200x600).

I noticed you added a feature (which I think is cool) that you can have the context 'guess' the appropriate width and height for the SVG element which I think is good, it saves the work from the coder, makes things render consistently (instead of one place using 600x300 and another 300x150).

but all of this doesn't explain how the chart seemed to get rendered in the right resolution, but somehow was offset by some amount in those piecharts.

johnSamilin: I fixed width detection and made some style improvements:
1. when determening bounding box's size, js can't keep in mind it's paddings, so I moved it to config that's being passed to render method
2. div#report should always have scroll bar because the widths could become irrelevant after all charts are rendered (size of the container will become smaller when scroll bar appears)

@chrisknoll regarding your question: render method receives width and height and passes it to createSvg method, which uses them only to specify viewBox attribute (which was a little bit confusing). And viewBox doesn't affect the size of the svg, it's only responsible for scaling. Those pie charts were given the width 1000px. In v 1.2.0 width and height are also used to set appropriate attributes, which led to this incorrect offsets

chrisknoll: Thanks @johnSamilin .  Somewhere along the way, I thought we had to set the w/h on the chart's SVG element but things evovled over time to support resizing the svg etc, and my recollection was wrong about where the w/h was being applied.  Thanks for digging into this.

pavgra: @anthonysena , since there seems to be no more open questions and remaining issues, can we merge the PR?

anthonysena: Yes - I'd like to review the changes but feel confident we'll be able to merge this shortly.

anthonysena: We are getting closer but I'm still facing an issue here. The new commits provide a good initial setup for the charts and allow them to scale to the available space. However, if I resize the window, the charts remain the same and then they wind up going outside the boundaries of the bounding box. This was not the case with the 1.1 version of the library so I'm hoping that this is a small change. Let me know if I can help to find a solution.

chrisknoll: OK: I remember now related to the widht and height settings:
in order to get the resizing int he SVG, you must omit a fixed width-height from the SVG element so that it fills the container (in this case a div).  Here's an example without the width/height applied to the element, before and after a resize:
Without Width properties:
nowidth
nowidth_resized
And with the width:
withwidth
withwidth_resized

So the width is used to specify the viewport as a way to tell the svg element 'this is the resolution that you should use to render the chart initially'. Then when the svg fills the containing element (like the div) it will scale up/down the SVG elements.

If you want to lock the SVG element to a fixed size, you would lock the containing element  to not let the SVG resize. But the SVG element shouldn't have a fixed width/height on it if you want to have the SVG resize.   Allt his being said: I still do not udnerstand how we get the piechart in the origional example 'pushed to the right' on all the charts as a result of fixing the chart with a fixed width. As you can see in my screenshots, when I took off the fixed width, the pie chart still had the correct position.  I just removed the width/height in my local repo on those charts and the piechart still appears int he correct position.

My guess is that there's something with the containing element and booststrap styles that is somehow interacting with the SVG elements and pushign them off their divs.  

But for starters, I'd revert the change that fixes the width/height on the charts if we want to resize those divs.  @anthonysena : maybe you could test again on your side setting the chart libary to 1.1 (where width/height are not set) and see if the latest comits to atlas still work under the 1.1 library. If it does, we should remove the width/height elements from createSvg() so that the SVG will resize naturally.

On the issue of fixed size elements in bootstrap:  this is one of the reasons we developed the atlas charts to be resized: iit seems if you put a fixed sized element inside the bootstrap 'row' layouts, the row/columns just ignore those elements and when you resize smaller, it looks like the layout 'pushed out' or 'bleeds over' the edge of the container instead of what you'd expect to happen is the container would stop shrinking and you'd get a scrollbar.

anthonysena: @chrisknoll thanks for going through all of these details! I tested this out by setting the chart library to 1.1 and found that the latest commits still work and allow the charts to properly scale to fill the containing element.

chrisknoll: Thanks, Anthony, I'll revert the width change in the SVG element for the 1.3.0 release of atlas chargs, but we may need to ask @johnSamilin  and @pavgra  to just signoff there may be other usecases where that width/height element may be important.

anthonysena: Sounds like a plan - thanks!

johnSamilin: Exactly, that's why I suggested two ways of solving this particular problem. I don't see any other use cases where setting fixed size of the chart is important.
@chrisknoll 
> still do not udnerstand how we get the piechart in the origional example 'pushed to the right' 
in the original PR they accidentally got width of 1000px (self.breakpoints.wide) https://github.com/OHDSI/Atlas/pull/646/commits/a6a89cae6e40031479b0214429fb002cf2d2bd4f#diff-5b98cce483fac3e765dee49534e89208R307

chrisknoll: sometimes the fonts in the chart may get too small if you shrink it too much, so what I'd do in this case is wrap the chart in a div with a min-height style to keep it from shrinking below a minimum.

Ok if we all agree, then I'll remove the width/height setting from the SVG element in the visualziation lib for the 1.3 release.

js/main.js
```diff
@@ -122,7 +122,7 @@ requirejs.config({
 		"cohortcomparison": "modules/cohortcomparison",
 		"r-manager": "components/r-manager",
 		"negative-controls": "components/negative-controls",
-		"atlascharts": "https://unpkg.com/@ohdsi/atlascharts@1.2.0/dist/atlascharts.min",
+		"atlascharts": "https://unpkg.com/@ohdsi/atlascharts@1.3.0/dist/atlascharts.min",
```
chrisknoll: This should be 1.3.1, due to NPM publishing error.

js/styles/chart.css
```diff
@@ -92,6 +92,7 @@ table.boxplotValues tr td
 }
 .axislabel, .axislabel text {
 	font-family: arial;
+	font-size: 1.2rem;
```
chrisknoll: What is style is '1.2rem'? Is that a real unit or just a typo from 'em'?

chrisknoll: Nevermind! rem is a real thing!

johnSamilin: rem is 'root em', which is calculated based on the font size of the body. 1rem = 100% body font-size, while 1em is 100% font-size of the parent element. I used it here to ensure that all labels have the same size

PR #647, opened 2018-04-16
Provide diagnostic checks for cohorts

### LINKED ISSUE #498: Enhancement request - Convey problems in cohort with use of concept sets
### Issues
1. When creating a cohort if a given concept set is not used anywhere in the cohort can ATLAS present a message to the user asking if it can be deleted? 
2. If a given concept set is used but references an incorrect domain can this be flagged to the user? 
3. If a given concept set is used but does not reference all of the domains can this be flagged to the user with the domains that are missing? 
### Ideas for Solutions
1. Provide a clean cohort button or present user a warning flag in the concept table alerting the user it is not used and the option to delete the concept set from there. 
2. Provide a warning to the user in the rule that the concept is used incorrectly and/or do so in the concept set table and provide reference to the rule number that incorrectly uses the concept set. 
3. Same as 2 immediately above except add references to the missing domains. 

Tagging posts #283 and #474 as they are related. 
Tagging @fdefalco @ericaVoss @chrisknoll @rmakadia to see what ideas they may have regarding how to resolve the issues. 

Thank you!

wivern:
closes #498


chrisknoll:
Can a filter be added to the Warnings tab so that we can filter to Warnings/Errors by default (but hide info) but if users what to see info warnings, they an un-hide info-level messages?

chrisknoll:
Ok, per #780, we're going to merge this into master after the refactoring is done. _*However*_, I'm going to continue to test this branch because issues found logically here will apply to whatever final refactor form. So let's proceed with pushing this PR through testing, and we'll plan on re-factoring this code into the #780 form once we're test-complete.

chrisknoll:
@wivern :  funtionality request:
It looks like the tab color changes based on if there is only info messages (it is blue) vs. if there is something non-info (it is red).  What do you think about having 3 levels of colors:  at least 1 error = red else at least 1 warning orange, else blue?

chrisknoll:
@wivern @johnSamilin @pavgra : with the cohort definition refactoring work, files are beign renamed and moved, which will make merging a nightmare.  I think you'll have to re-apply your code changes onto the refactored version of the code (once that gets into master), but in the future I'd recommend better coordination between refactoring and enhancements so that we don't have these sorts of collisions.

PR #672, opened 2018-05-03
Data sources new drilldown plots and refactor

MaximMoinat:
Adds three new plots to data sources
* donut chart of `value_as_concept_id` in observation and measurement
* donut chart of `qualifier_concept_id` in observation
* donut chart of `operator_concept_id` in measurement

Other improvements:
* Data sources table sorts by default on person count (mimicking default behaviour of AchillesWeb).
* Refactoring of boxplot and donut chart rendering methods to optimize reuse of code
* No filtering on measurement_concept_id for unit/value plots. The WebAPI does this since release 2.4.0

See also:
* https://github.com/OHDSI/WebAPI/pull/403 
* http://forums.ohdsi.org/t/achilles-proposal-for-new-observation-measurement-value-visualisation/4340


johnSamilin:
Hi @MaximMoinat 
due to recent refactoring (https://github.com/OHDSI/Atlas/pull/694), data sources component is now under /pages folder, could you move these changes there?

MaximMoinat:
@johnSamilin I started moving the changes, however, I encountered two blocking issues.
1. multiple donut charts on one page malfunction. They all show the same data. Probably a problem with observables referencing each other. e.g. look at the person page, all three donuts contain the same data:
image
2. The measurement treemap drilldown displays 'No data' for every concept I tried, while there is data available.

chrisknoll:
Right, this is an issue I uncovered as well from the refactor using the 'clean and reasonable project structure'.  There seems to be some issue where the bindings are overwritting with the last chart rendering.  

I'm speaking with @pavgra about reworking those classes that follow the knockout pattern of components.  We hope to have this addressed shortly.

MaximMoinat:
Changes moved to new data sources structure. Fixed the measurement drilldown and frequency histogram in the process.
@johnSamilin could you review?

johnSamilin:
Hi @MaximMoinat 
absolutely right, /components/data-sources will be removed in the future cleanup

PR #678, opened 2018-05-11
token refresh promise should be fulfilled before the list of cohorts is requested

### LINKED ISSUE #677: If admin deletes cohort, he will be logged out
### Expected behavior
Admin deletes cohort, then is redirected to the list of cohorts
### Actual behavior
Admin deletes cohort, then is being logged out
Log is, delete a cohort

johnSamilin:
fixes #677

### DIFF
--- a/js/components/cohort-definitions/cohort-definition-manager.js
+++ b/js/components/cohort-definitions/cohort-definition-manager.js
@@ -372,12 +372,14 @@ define(['knockout', 'text!./cohort-definition-manager.html',
 
 			// reset view after save
 			cohortDefinitionAPI.deleteCohortDefinition(self.model.currentCohortDefinition().id()).then(function (result) {
-				self.model.currentCohortDefinition(null);
 				if (config.userAuthenticationEnabled) {
-					authApi.refreshToken();
+					return authApi.refreshToken();
 				}
-				document.location = "#/cohortdefinitions"
-			});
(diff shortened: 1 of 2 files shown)


chrisknoll:
Hi, I've seen this 'logout' behavior in areas of the app outside of the 'delete cohort definition' usecase you've described. The fix seems to ensure token is refreshed before calling a WebAPI call, and I'm wondering if this 'refreshToken()' call needs to be called every time a WebAPI token is invoked?  If so, should this be centralized in the WebAPI set of components that communicates to the server (the other WebAPI components besides CohortDefinitionAPI)?

Another thing I am wondering if there's some more basic issue with how we're handling authentication: is it really the case that you need to refresh your authentication token before making any request to the server?  I'd have expected this to be handled more transparently (such as the token is fresh within a certain timeout).

pavgra:
@chrisknoll , I believe, you are right: the refresh functionality should be coded in a more transparent way and with one call in one place, not with a refresh call per each secured method usage. I believe, @johnSamilin won't have any objections?

johnSamilin:
Maybe it's better to refresh the token periodically instead, using setInterval?

pavgra:
@johnSamilin , then we'll lose session timeout functionality

chrisknoll:
Closed in favor of #685.

PR #685, opened 2018-05-14
Centralized token refresh

pavgra:
- Introduces centralized token refresh (on navigation to a new page)
- When the token is expired, displays login form with notification to let the user to re-login and continue working w/o loss of progress.

Closes https://github.com/OHDSI/Atlas/issues/677. Instead of https://github.com/OHDSI/Atlas/pull/678

(We cannot refresh token on ajax's "beforeSend", because some of ajaxs are called by setInterval and that would break session timeout functionality).

@chrisknoll , @johnSamilin

### DIFF
--- a/js/modules/WebAPIProvider/AuthAPI.js
+++ b/js/modules/WebAPIProvider/AuthAPI.js
@@ -16,6 +16,13 @@ define(function(require, exports) {
         return config.webAPIRoot;
     };
 
+    var token = ko.observable();
+    if (localStorage.bearerToken && localStorage.bearerToken != 'null') {
+        token(localStorage.bearerToken);
+    } else {
+        token(null);
+    }
+
     var tokenExpirationDate = ko.pureComputed(function() {
         if (!token()) {
             return null;
@@ -25,6 +32,33 @@ define(function(require, exports) {
         return new Date(expirationInSeconds * 1000);
 
     });
+
+    const tokenExpired = ko.observable(false);
+    const askLoginOnTokenExpire = (function() {
+        let expirationTimeout;
+        return () => {
+            if (expirationTimeout) {
+                clearTimeout(expirationTimeout);
+            }
+            if (tokenExpirationDate() > new Date()) {
+                tokenExpired(false);
+                expirationTimeout = setTimeout(
+                    () => {
+                        tokenExpired(true);
+                        $('#myModal').modal('show');
+                        expirationTimeout = null;
+                    },
+                    tokenExpirationDate() - new Date()
+                );
+            } else {
+                tokenExpired(true);
+            }
+        }
+    })();
+
+    askLoginOnTokenExpire();
+    tokenExpirationDate.subscribe(askLoginOnTokenExpire);
+
     var permissions = function() {
         var permissionsString = localStorage.getItem(LOCAL_STORAGE_PERMISSIONS_KEY);
         if (!permissionsString) {
@@ -38,12 +72,6 @@ define(function(require, exports) {
             ? parseJwtPayload(token()).sub
             : null;
     });
-    var token = ko.observable();
-    if (localStorage.bearerToken && localStorage.bearerToken != 'null') {
-        token(localStorage.bearerToken);
-    } else {
-        token(null);
-    }
 
     window.addEventListener('storage', function(event) {
         if (event.storageArea === localStorage && localStorage.bearerToken !== token()) {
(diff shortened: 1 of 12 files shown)


js/modules/WebAPIProvider/AuthAPI.js
```diff
@@ -25,6 +32,33 @@ define(function(require, exports) {
         return new Date(expirationInSeconds * 1000);
 
     });
+
+    const tokenExpired = ko.observable(false);
+    const askLoginOnTokenExpire = (function() {
```
johnSamilin: Token will be refreshed (when route is changed), but the tokenExpirationDate will remain the same, won't it?

js/modules/WebAPIProvider/AuthAPI.js
```diff
@@ -25,6 +32,33 @@ define(function(require, exports) {
...
+        return () => {
+            if (expirationTimeout) {
+                clearTimeout(expirationTimeout);
+            }
+            if (tokenExpirationDate() > new Date()) {
+                tokenExpired(false);
+                expirationTimeout = setTimeout(
+                    () => {
+                        tokenExpired(true);
+                        $('#myModal').modal('show');
```
johnSamilin: The login modal is already bound to `<div data-bind="if: !isLoggedIn() || tokenExpired()">`, why do we need to show this modal?

pavgra: @johnSamilin , where is the modal bound to the condition? Its contrnt does

js/modules/WebAPIProvider/AuthAPI.js
pavgra: @johnSamilin , why should it if there is “subscribe” below?

chrisknoll:
Hey @johnSamilin and @pavgra,
  Question for you on this approach...is it possible to address this from the other direction:  allow a long token timeout on the server side, but on the client side, perform a 'signout' or 'expireToken' after a certain timeout.

If you go in this direction, the the app just has a single setTimeout() that gets reset every time there is a user action (note: no WebAPI round trip), but if no activity happens, the setTimeout() will signal the WebAPI to signout/expireToken/whatever and the user will have to re-login again.

If we do it this way, we won't have the case were we're hitting WebAPI on client-side-only interactions, and we only need to hit the WebAPI to expire the token when we detect there's a timeout.

The other advantage of this approach:  in environments where we don't want to enforce this 'logout after 15 minutes', we can configure the Atlas timeout to be '0' or undefined and we won't do any setTimeout() at all.  I'm suggesting this because:  in environments where you are using Window authentication, the timeout actually doesn't do anything (Except annoy users):  when the timeout occurs, you have to 'sign in' again but in Windows authentication that just means 'push the windows login button (you don't type in a user/password), so the timeout hasn't really protected anyone from anything if they have Windows Auth).  But, if you are using database auth, you will be prompted for a username/password (at least until a 'remember me' option is implemented).

-Chris

pavgra:
@chrisknoll, 
> the the app just has a single setTimeout() that gets reset every time there is a user action
imagine that I've opened the website, did some actions and then closed the tab. Who is going to timeout my session then?

pavgra:
I would also like not to reinvent the wheel: JWT has a special section for storing the expiration date, so why to ignore it and invent own (and duplicative) mechanisms for doing timeout?

pavgra:
@chrisknoll , let me summarize all my statements here.

1. We follow pretty standard approaches for handling JWT. I would refer here to Auth0 as it is one of the biggest SaaS providers in Authentication and Authorization fields:
- We need to setup expiration dates for access (security) tokens. [Link](https://auth0.com/docs/jwt):
> Give tokens an expiration. Technically, once a token is signed – it is valid forever – unless the signing key is changed or expiration explicitly set. This could pose potential issues so have a strategy for expiring and/or revoking tokens.
- We mustn't have a refresh token. [Link](https://auth0.com/docs/tokens/refresh-token/current):
> A Single Page Application (normally implementing Implicit Grant) should not under any circumstances get a Refresh Token. The reason for that is the sensitivity of this piece of information.
- We refresh token silently when the user is active. [Link](https://auth0.com/docs/api-auth/tutorials/silent-authentication):
> When the Access Token has expired, silent authentication can be used to retrieve a new one without user interaction, assuming the user's SSO session has not expired.

1. Your case, when you don't want tokens to be invalidated shortly, is covered by ability to change setting in pom.xml called `security.token.expiration`

1. The mechanism, which I implemented in the PR is general, no new code will be required. The routing was chosen as a way to detect user's activity since it is common and general action and we cannot refresh token on ajax's "beforeSend", because some of ajaxs are called by setInterval and that would break session timeout functionality.

1. Out overall auth implementation was tested by security auditors, it works, it covers most of the cases and I do not think that it is reasonable to spend more time on it currently, while we have other opened issues

Tagging @anthonysena , @fdefalco to hear their opinion too

PR #705, opened 2018-05-23
Not finished Cohorts / IRs should be invalidated on startup

wivern:
closes OHDSI/WebAPI#429

### DIFF
--- a/js/components/cohort-definitions/cohort-definition-manager.js
+++ b/js/components/cohort-definitions/cohort-definition-manager.js
@@ -438,11 +438,9 @@ define(['knockout', 'text!./cohort-definition-manager.html',
 			if (source) {
 				switch (source.status()) {
 					case 'COMPLETE':
-						return false;
-						break;
+					case 'ERROR':
 					case 'n/a':
 						return false;
-						break;
 					default:
 						return true;
 				}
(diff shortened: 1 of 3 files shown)


chrisknoll:
Hi, @wivern,
  I'm trying to work through where the executionInfo.status is being set. I understand the ENUM has PENDING/RUNNING/COMPLETE/ERROR in it, but I think this was a divergence between what the execution status was originally designed for: the execution status should just be the status of execution, not if the result is valid or not (there should be a separate field 'isValid' to store this).

I'm looking in the tasklet: https://github.com/OHDSI/WebAPI/blob/master/src/main/java/org/ohdsi/webapi/ircalc/PerformAnalysisTasklet.java and the only status it sets is RUNNING when the execution starts, and COMPELTE when the execution ends (whether or not it was successful or not).

On the service REST endpoint, we look at the status here: https://github.com/OHDSI/WebAPI/blob/master/src/main/java/org/ohdsi/webapi/service/IRAnalysisService.java#L475, but we don't check for ERROR....I think that the ERROR value in the GenerationStatus is a mistake..it should only contain the status of the exectution (is it pending, running, or complete).

For unfinished cohorts/IR to be invalidated on startup, I think that the check should be:
If anything is in PENDING/RUNNING status, you should reset them to COMPLETE, set the isValid to 'FALSE' and set a fail message to 'Execution reset at startup.'

wivern:
Hi @chrisknoll,
thank you for detailed explanation. I was a little confused by GenerationStatus.ERROR,
but next PR about invalidation https://github.com/OHDSI/WebAPI/pull/444/files corrects this (AbstractDaoService.java) and doing exactly the same as you told.
So this PR seems unnecessary.

PR #732, opened 2018-05-31
Structure refactoring/data sources

johnSamilin:


js/components/charts/chart.html
```diff
@@ -1,4 +1,4 @@
-<div data-bind="reference: setContainer, if: data"></div>
+<div data-bind="chart: $component, if: data"></div>
```
pavgra: too abstract. why not to define those 5 props explicitly

js/modules/databindings/chart.js
```diff
@@ -0,0 +1,27 @@
...
+        minHeight,
+        format
+      );
+    }
+  }
+
+  ko.bindingHandlers.chart = {
+		update: function (element, valueAccessor) {
+      const chart = valueAccessor();
+			draw(chart.data, chart.prepareData, element, chart.minHeight, chart.format, chart.chart);
```
pavgra: why not to pass ready data?

js/components/charts/chart.html
johnSamilin: Beacause it's not really nessessary if you pass here a component which extends Chart

js/modules/databindings/chart.js
```diff
@@ -0,0 +1,27 @@
...
+        minHeight,
+        format
+      );
+    }
+  }
+
+  ko.bindingHandlers.chart = {
+		update: function (element, valueAccessor) {
+      const chart = valueAccessor();
+			draw(chart.data(), chart.prepareData, element, chart.minHeight, chart.format, chart.chart);
```
pavgra: `chart.chart` smells

js/components/data-table.js
```diff
@@ -1,83 +0,0 @@
-define([
```
pavgra: Why it was fully dropped?

chrisknoll:
Hi, @wivern , @pavgra ,
  I feel that this project re-structure is going after 2 objectives: 

- Clean up the folders so that the code can be better organized for easier navigation and understanding (so we have folders for pages, reports, etc, and you can understand what pages we have by looking at the pages folder, etc)
- Apply an abstraction layer on top of KO components in an effort to 'formalize' the construction of a component, but also (in my opinion) to make it resemble other other javascript frameworks (such as react), and I'll assume that this is because it's familiar with the authors of the code to structure it this way, or there's some effort to make the code look more attractive to a wider audience of developers.

On the first point: I think we're in good shape: the codebase is being re-organized and we already see the benefits of splitting up datasources into smaller, more identifiable sub-reports that we have more opportunities to identify where code can be re-used, etc.

On the idea of the abstraction layer for components:  I think we're losing the elegance of KO components and databindings in an effort to construct the notion of a 'component that you can instantiate', while this isn't the motivation behind KO: in KO you register a specific javavascript object as a 'compnonent' (which consists of a viewmodel and optional template) and place it on the screen via HTML markup.  You never need to ask the component to redraw, or refresh, it doesn't have data pass through it via 'uniflow', etc, and I think we should structure our code around the databind behavior, and not a more 'generalized unifflow' stucture (I could be misinterpreting the author's intent, but this is just how it feels to me...)

So, I'd like to focus on the second aspect of our code restructuring:  In looking at the code changes thus far, I'm sensing that there's some challenge with the 'statefulness' of components (the viewmodel is the state) vs. databindings are generally stateless (the data to bind is passed in the update() callback) vs. our AtlasCharts classes are stateful (you call new boxplot() to get a boxplot, and then render).  I think the statefulness of AtlasCharts and the statelessness of databindings is one source of friction in how we are developing code;  Maybe the thought process is: since state is involved we need a component to wrap our chart library.  But we should be really using a data-bind to associate the chart data to the chart visualization.  Databindings _can_ maintain some state (ie: in the init() of a chart, you could allocate an instance of the AtlasChart), or part of the databinding could be the instance of the chart to use.

So I'd like to take a step back from what we've done so far (defining 'chart component' that we extend to specific charts (boxPlot, line, scatter, etc), as well as how we have this generic 'Component' class that all UI components should extend)....and think about the components we are building as more supporting a class of use-cases.  

First, I just want to state what I mean by KO component and KO databind:

KO Components (full details found here: http://knockoutjs.com/documentation/component-overview.html):
- can have an associated template, but must have a view model (a constructor function that is invoked each time a component is placed into the DOM), and therefore can maintain state.
- Do not get a reference to a DOM element
- Are designed to yield DOM elements from a template of nodes or HTML string.
- Data is passed in via the 'params' property
- Remain alive until the name of the component binding changes (which will cause a different instance to load) or the containing DOM element is destroyed.  Data can be sent to the component via params, *but observables in the params that update does not cause the component to re-draw/dispose*.


KO DataBindings: (see: http://knockoutjs.com/documentation/custom-bindings.html)
- provide an init(), update(), and dispose method which get triggered when the databinding is attached to a DOM element, or when an observable value changes that re-triggers the databind's update().
- the attached element is passed in as the parameter 'element' to init and update.
- data is passed in via the syntax: ```data-bind="customDataBind: _theData_"``` and is accessed via valueAccessor() (which returns the value _theData_)
- Yields DOM elements programatically by adding children to the element, or using some external dom-generation process. But adding/removing elements is an operation of the code of the data-bind.

I think that you guys are making a good attempt to try to structure the report classes around 2 main ideas: an AbstractReport and a TreeMap report (which extends AbstractReport).  I think where it needs improvement is how there's 'render()' methods in the component and how DOM elements are being manipulated in the component, where DOM manipulation should be done at the data-bind level.

So, could we focus on the Report abstractions first, maybe we can identify places in the general 'Component hierarchy' that we could adjust to be a bit cleaner, and then once that is sorted out, we can apply some consistent principles across the other components of the application.

Would it be easier to discuss this via via a call?

pavgra:
@chrisknoll, thank you for the comment.

If I understood the message idea correctly, your concerns are about how we implement components and bindings. If it is so, the two things I'd like to state as an entry point for precise discussion and search for better options (not in abstract terms, but applied solutions):

1. *Components*. We tried to solve two problems:
- Simplify creation of components with shared code (behavior) using OOP. So _the only change here is the usage of class with constructor instead of plain function_. That gives us ability to move some logic into methods and overwrite it in children classes. Exactly the same approach was used in Cost & Util.
- Move redundant code of components registration into helper (write 1 line instead of 5). That's `Component.build`. Pretty simple but obviously useful.

2. *Bindings*. There is one new binding in the PR called `/js/modules/databindings/chart.js` and it is absolutely stateless, so I am not sure why it doesn't fit KO ideology.

js/pages/data-sources/classes/Treemap.js
```diff
@@ -85,12 +87,22 @@ define([
...
         },
       };
+      // to pass down to drilldown
+      this.currentReport = params.report;
+      this.getData()
+        .then(() => {
+          // in order to get jquery working, we should set isLoading here instead of .finally block
+          this.context.loadingReport(false);
+          this.isLoading(false);
+          $("#report_table").DataTable(this.chartFormats.table);
```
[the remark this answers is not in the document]
johnSamilin: fixed

js/pages/data-sources/components/reports/datadensity.js
```diff
@@ -52,7 +52,36 @@ define([
           yLabel: 'Concepts per Person'
         },
       };
-
+      
+      this.getData()
```
pavgra: Move getData to separate function and call it from constructor. Do the same with all similar cases

js/components/charts/frequencyHistogram.js
```diff
@@ -2,14 +2,20 @@ define([
 	'knockout',
   'pages/data-sources/const',
   'providers/Chart',
+  'providers/Component',
```
pavgra: Apply proper formatting (as I remember, somewhere OHDSI decided to use tabs, but it's better to double-check). It is impossible to read code now

js/components/charts/boxplot.js
```diff
@@ -1,25 +1,25 @@
...
 
-    render(params) {
-      super.render(params);
-      return this;
-    }
   }
 
-  const viewModel = new Boxplot();  
-	return viewModel.build();
+	return helpers.build(Boxplot, 'boxplot', view);
```
pavgra: Use destructuring. Put component's name first. Use original KO's keys for `viewModel` and `template`

js/pages/data-sources/components/reports/conditionEra.js
```diff
@@ -1,33 +1,29 @@
...
   'pages/data-sources/components/reports/treemapDrilldown',
 ], function (
 	ko,
-	view,
+  view,
+  Component,
   TreemapReport,
-  helpers
+  helpers,
+  globalHelpers
```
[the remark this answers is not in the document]
johnSamilin: because const is a reserved word

js/pages/data-sources/components/reports/datadensity.js
johnSamilin: it _is_ separate function (as a method of super class) and it is being called from constructor. Do you mean moving a callback which is parsing the data?

js/utils/CommonUtils.js
```diff
@@ -0,0 +1,231 @@
+define(
+  (require, factory) => {
+    const d3 = require('d3');
+		const _ = require('lodash');
+		const ko = require('knockout');
+
+    const mapConceptData = function (data) {
```
pavgra: Most of methods in the class are for smth like ChartUtils, not project-wide utils

js/utils/CommonUtils.js
```diff
@@ -0,0 +1,231 @@
+define(
```
pavgra: Indentation needs to be fixed

js/pages/data-sources/components/reports/achillesHeel.js
```diff
@@ -2,65 +2,60 @@ define([
...
 ], function (
 	ko,
 	view,
 	d3,
-  helpers,
-  Report
+	commonUtils,
+	Report,
+	Component,
+	commonUtils
```
pavgra: duplication

chrisknoll:
Hi, after meeting with Pavel, there's just 2 things you can decide if you want to change it now or later:

1. Report.js is defining a AbstractReport class, maybe either change it that the internal class is called 'Report' or you rename the file to 'AbstractReport' (and adjust all the define() calls to refer to the new module name.

2. The component names you've specified in the datasources are a little too generic (ex: 'person', 'conditoin') and it might be worthwhile to either prefix them with 'ds-' to indicate this is a datasource component, or call them 'person-report', 'condition-report' etc.   This isn't critical now since you're not running into any component name conflicts, but could bump into someone later if they want to make a 'condition' component.

Otherwise, the project structure looks good (and greatly improved) so it passes my check for merge into master if no one else has objections.

pavgra:
@chrisknoll, great! and thank you for your valuable comments. I'd like to proceed with the PR merge for now due to two reasons - I wouldn't like it to go out of sync with current master again and it will be easier to review the changes, which we are going to make to resolve your latest comments, when they are not mixed with tons of other changes. And also, to be honest, this thing has been here for a while - just want to cross this out from TODO list.

Issue #740, opened 2018-06-11
Atlas PLP fails due to changed PLP package API


pavgra:
```
Executing SQL took 2.45 mins
Fetching data from server
Fetching data took 13.8 secs
Fetching outcomes from server
Loading outcomes took 6.14 secs
Outcome is 0 or 1
Error in PatientLevelPrediction::setGradientBoostingMachine(ntrees = c(10,  : 
  unused arguments (max_depth = c(6), min_rows = c(20), learn_rate = c(0.1))
Execution halted
```

PR #741, opened 2018-06-11
PLP code generation fix

pavgra:
Closes https://github.com/OHDSI/Atlas/issues/740

### DIFF
--- a/js/components/plp-r-code.html
+++ b/js/components/plp-r-code.html
@@ -245,9 +245,9 @@
     <span class="token comment"># Create the model settings ----</span>
     modelSettings <span class="token operator"><-</span> PatientLevelPrediction::setGradientBoostingMachine(ntrees = c(<span class="token number" data-bind="text: $component.patientLevelPrediction().moNTrees"></span>),
                                          nthread = <span class="token number" data-bind="text: $component.patientLevelPrediction().moNThread"></span>,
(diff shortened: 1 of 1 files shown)


chrisknoll:
Sorry: I forget that when we reviewed the PLP script, was the version of the PLP library that is referenced in the script up to date as well? Should we update both the parameters as well as the referenced library are pointing to the latest version?  If it was already set, then disregard.

pavgra:
@chrisknoll , the latest released version fails at Redshift, so I would either wait for the next release or use latest commit instead

chrisknoll:
Of course, let's not update it if it's broken. I hope they know about the redshift issue.

Thanks for checking.

pavgra:
@cgreich , the issue was fixed, but there were no releases after the fix was introduced. So the only option is to replace the `ref = "2.0.0"` with commit hash

chrisknoll:
So, I don't recommend referencing a hash, so we can either wait for an official 2,0.5 or 2.1.0 of PLP or we can just push this PR as is and come back to update it later.

@jreps: do you know if there will be a release of PatientLevelPrediction soon?

pavgra:
@chrisknoll , I'd push it and then update the version as soon as the release happens

chrisknoll:
Ok sounds good to me. Go ahead and we'll revisit this when PLP gets updated.

PR #743, opened 2018-06-14
Fixed parameters for various ML models

JamesSWiggins:
Some of the ML model parameter names don't match the actual parameter names in current versions of PatientLevelPrediction.  For instance, setRandomForest specifies 'max_depth', but the actual parameter name is 'maxDepth'.


anthonysena:
@JamesSWiggins - thanks for making these changes! Can I ask a favor: could you rebase your changes off of the v2.4.3 branch in Atlas? @pavgra had also made some updates to the PLP code generator that overlap with some of the changes you made. I attempted to change the base branch here but it looks like it may be easier to issue a new PR based off of v2.4.3. Let me know if you have any questions or concerns.

PR #746, opened 2018-06-15
Fixed parameters for various ML models

JamesSWiggins:
Some of the ML model parameter names don't match the actual parameter names in current versions of PatientLevelPrediction.  For instance, setRandomForest specifies 'max_depth', but the actual parameter name is 'maxDepth'.


chrisknoll:
Thanks for your contribution!

PR #748, opened 2018-06-18
Structure refactoring/vocabularies

johnSamilin:
added components: tabs, panel
moved VocabularyProvider to providers/Vocabulary


johnSamilin:
Hi @chrisknoll,
thanks for pointing up on these errors, I've fixed them

chrisknoll:
Still getting a problem.  I search for 'infarc' and this endpiont is requested:
http://localhost:8080/WebAPI/vocabulary/OPTUM-PDW/search/infarc

However, the results panel says 'no concept sets found' (which i'm not sure why it's showing a concept sets message when I am doing a vocabulary search.

johnSamilin:
fixed

chrisknoll:
Ok, things look like they are working.

There's one thing I'd like to ask that you consider looking into. I'm not sure if it was a result of this refactoring, or some other refactoring that was done on the search page (and something to do with routes):

When performing the search, there's an odd flashing that happens while the route is updated and the search is performed.  You hit enter in the search box and the screen does a sort of flash, then the search box is updated with the value form the URL, and then the results present.  I think this is something new since the refactoring.

Something that is pre-existing that I would ask that you maybe look into:  whenever a route changes (because of a new search, or if you navigate to another left-nav item) there's a wierd flash as the route changes.  I think there's some default route behavior that clears the active page (and replaces it with a loading component) and then when the route engages, it replaces the main viewport with the 'active page' component.

Do you think you could look into why we have this "component-dance" going on? I think it should just be a matter of replacing one nav-component with a different one based on the requested route.  Admittedly, some of this is from the prior version of the code, but as long as we're opening it up, maybe you have an improvement to make on this?

johnSamilin:
This happens because of the token refreshing between different routes. There's a loader on the page until the backend answers. This can be fixed by hiding this loader, but the core logic will remain the same: route won't be changed for some amount of time. Is that ok?

chrisknoll:
Is it a simple change to see how it works with it implemented and then if it feels like it's laggy, we can revert?  This only is an issue when security is enabled, yes?  I think that I need to talk to @pavgra about possibly only pinging the server for a refresh until it needs to.  But, I'll talk to him about that offline.

PR #749, opened 2018-06-20
Improvements for auth handling to support different auth managers (regular and proxy-based)

pavgra:
Closes https://github.com/OHDSI/WebAPI/issues/479

Pls, leave Merge button for me - testing is still in progress

### DIFF
--- a/js/components/welcome.html
+++ b/js/components/welcome.html
@@ -41,7 +41,7 @@
 
 			<div data-bind="if: isLoggedIn() && !tokenExpired()">
                 <div data-bind="text: $component.status"></div>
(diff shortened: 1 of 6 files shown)


js/components/welcome.html
```diff
@@ -41,7 +41,7 @@
 
 			<div data-bind="if: isLoggedIn() && !tokenExpired()">
                 <div data-bind="text: $component.status"></div>
-				<span class="btn btn-sm btn-default" data-bind="click: $component.signout"><i class="fa fa-sign-out"></i> Sign Out</span>
+				<span class="btn btn-sm btn-default" data-bind="click: $component.signout, visible: $component.token()"><i class="fa fa-sign-out"></i> Sign Out</span>
```
johnSamilin: this does not make sense since it won't be rendered with expired token

Issue #750, opened 2018-06-20
add a name of a person who kicked off a job


gklebanov:
in jobs screen, add a name of a person who kicked off a job

PR #762, opened 2018-07-04
Add a name of a person who kicked off a job (#750)

mpozhidaeva:


chrisknoll:
When pulling up cohort defintions using this branch, I'm getting this error in the console:

```
POST http://localhost/atlas/undefinedlookup/identifiers 404 (Not Found)
```

This error happened previously when there was an async call to the services that the main startup of the app wasn't waiting for to complete.  However, this error seems to have returned.

Issue #697, opened 2018-05-19
[Bug] - utilization function of heracles


gowthamrao:
### Expected behavior
When selecting filters in panel, summary table should change

with no filters selected
![image](https://user-images.githubusercontent.com/13936600/40271968-544077fe-5b73-11e8-80f8-71021415fbea.png)

### Actual behavior

with filters selected
![image](https://user-images.githubusercontent.com/13936600/40271973-675fed38-5b73-11e8-8173-f66d19f4b582.png)

### Steps to reproduce behavior

@pavgra @chrisknoll

pavgra:
@anthonysena , this one was merged into RC 2.4.1 branch

gowthamrao:
@pavgra the issue applies to both visits and drugs. Looks like only visit was fixed?

pavgra:
@gowthamrao, should be fixed in 2.4.3 now

PR #763, opened 2018-07-05
Fixed name of query param in Drug Util report

pavgra:
Fixes https://github.com/OHDSI/Atlas/issues/697


js/modules/cohortdefinition/components/view/reporting/cost-utilization/base-drug-util-report.js
```diff
@@ -13,7 +13,7 @@ define(
   ],
   function (ko, view, BaseCostUtilReport, appConfig, BemHelper, config, costUtilConst, CohortResultsService) {
 
-    const DRUG_SOURCE_CONCEPT = 'drugSourceConcept';
+    const DRUG_SOURCE_CONCEPT = 'drugType';
```
agackovka: missleading name

Issue #764, opened 2018-07-10
Histogram data is calculated incorrectly when min X value is greater than zero


pavgra:
### Steps to reproduce behavior
Open Data sources -> Daashboard -> Age at first Observation

### Expected behavior
Proper distribution

### Actual behavior
Distribution shifted by the value of min age to right

PR #766, opened 2018-07-10
Fixed calculation of data for histograms in Data sources

pavgra:
Fixes https://github.com/OHDSI/Atlas/issues/764
(cherry picked from commit d0195b2 into master)


chrisknoll:
@pavgra , I've pushed a commit to change how histogram mapping works: it now iterates across the interval data (instead of trying to do a lookup to the interval index for the 'i' variable.  I've also made changes to lock in the x-axis for the age at first observation (on the dashboard) so that datasources where ages start at 18 or are less than 90 will have a consistent x-axis domain.  If you could try this verison out (this was done in master, I'm not messing with anything in the 2.4.3 branch). let me know if you run into any issues.

js/components/data-sources.js
```diff
@@ -1093,11 +1094,11 @@ define(['jquery', 'knockout', 'atlas-state', 'text!./data-sources.html', 'd3', '
 			var minValue = histogramData.min;
 			var intervalSize = histogramData.intervalSize;
 
-			for (var i = 0; i <= histogramData.intervals; i++) {
+			for (var i = 0; i < histogramData.data.intervalIndex.length; i++) {
 				var target = new Object();
-				target.x = minValue + 1.0 * i * intervalSize;
+				target.x = minValue + (1.0 * histogramData.data.intervalIndex[i] * intervalSize);
```
pavgra: @chrisknoll, intervalIndex contains list of ages, e.g. interval from 18 to 90, so it doesn't make sense to add `minValue` to the indexes

chrisknoll: In that case, the minValue will be 0.  The wider usecase of supplying a minValue is where the histogram data is comming in as 0-indexed, but you want the X-axis of the report to start at some non-zero x-coordinate value.

For example: if you recieved histogram data for 'year 1 after studyStart'...'year 2 after studyStart'..etc, but you know the study started in 2005 you would set the minValue to 2005, and use the histogram index as the 2005+index to give you the bar's position on the hisogram.  

In the cases of our age histograms, we know the indexes don't need to be offset, so the minValue is 0 so not to offset anything.

Note, this code is done in 'map histogram' so no assumptions can be made there about what the nature of the indexes are.  The minValue is just an option that lets you specify that the index value should map to another actual value on the X coordinates.  You can see in the code for 'age at first' we're setting the min/max based on the values of the indexes (because thsoe indexes have actual meanings about the ages.

johnSamilin:
@pavgra @chrisknoll haven't we moved data-sources component under `pages` folder?

chrisknoll:
@johnSamilin , that's a good piont, this PR is for master which should have the refactored code in it. let me check.

chrisknoll:
The refactored code is there, the /components/data-sources.js is the component for the <data-sources> tag so I believe the code is in order. Can you confirm?

chrisknoll:
So, I am a bit confused as to what I saw prior to my commit to this branch:  I had to make adjustments in components/data-sources.js that @pavgra submitted, and those changes  I made did have an effect, i was doing debugging in chrome, so the code was loaded and executed, but, looking closer at it now in master, it doesn't appear /components/data-sources.js is loaded, and the code that should be modified is in /pages/data-sources/components/reports/dashboard.js.  I'll push the changes pavel provided to that component.

Could we remove any un-used code since the refactoring?  if components/data-sources.js is gone, let's remove it from the codebase.

johnSamilin:
@chrisknoll sure, that's why I've added `console.warn('deprecated');` at the top of the refactored components (which are, for now, `components/data-sources`, `components/search` and `modules/WebAPIProvider/VocabularyProvider`).
I also removed them from `main.js` paths section, but kept for now in order to do cleanup after structure refactoring is done

chrisknoll:
Ok, I've applied the histogram changes to the proper files.
I did clean out the data-sources/template file from the project. If we need to refer to them, we can always look at a prior commit with the files in it to see what is there, but i wanted to remove it because doing find-in-files was pointing me to things that weren't actually active.

chrisknoll:
@johnSamilin : i actually have seen deprecated warnings under master for ```VocabularyProvider.js?_=ddpzsr:2 deprecated, moved to /providers/Vocabulary'``` and i'm not sure if you are saying that we shouldn't be seeing those warnigns (because those modules shoudl not be called). if we shoudln't see depricated warnings, then there's an area of code that is still calling VocabularyProvider under the old codebase...perhaps it's because we haven't refactored those moduels yet...so in that case, maybe we need to leave that code in place until we complete the refactoring.

johnSamilin:
@chrisknoll yes, exactly, that's the idea of why there's still two VocabularyProviders. Let's leave both for now, but keep in mind that VocabularyProvider should be in the `providers` folder in the end

chrisknoll:
Ok, then if this PR is good, someone apply an approval on it and I'll merge it in.  Since @pavgra opened it and I added to it, @johnSamilin  will you do the honors?

PR #767, opened 2018-07-11
Support domain type concept exclusions.

chrisknoll:
Adds a new UI element to toggle 'is any of' and 'is not any of' for domain type concepts. This allows you to do things like 'where type not in patient reported'.

### DIFF
--- a/js/components/cohortbuilder/components/CycleToggleInput.js
+++ b/js/components/cohortbuilder/components/CycleToggleInput.js
@@ -1,0 +1,31 @@
+define(['knockout', 'text!./CycleToggleInputTemplate.html'], function (ko, template) {
+
+	function CycleToggleInputViewModel(params) {
+		var self = this;
+		
+		function getOptionIndex(value, options) {
+			return (value == null) ? 0 : options.indexOf(self.options.filter((option) => option.value === value)[0]);
+		}
+		
+		self.value = params.value;
+		self.options = params.options;
+		
+		self.currentIndex = ko.pureComputed(() => getOptionIndex(self.value(), self.options));
+		self.selectedOptionText = ko.pureComputed(function() {
+			return self.options[self.currentIndex()].name;
+		});
+		
+		self.nextOption = function () {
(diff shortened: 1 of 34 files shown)


js/modules/cohortbuilder/components/ConditionOccurrenceTemplate.html
```diff
@@ -49,7 +49,7 @@
 			<tr data-bind="if: ConditionType() != null, visible: ConditionType() != null">
 				<td><i data-bind="click: function() { $component.removeCriterion('ConditionType') }" class="fa fa-times" /></td>
 				<td>
-					Condition Type is:
+					Condition Type <button class="btn btn-primary btn-xs" data-bind="click: function() { ConditionTypeExclude(!ConditionTypeExclude())}, text: ConditionTypeExclude() ? 'is not any of': 'is any of'"></button>
```
pavgra: The added button is pretty the same in all changed files: wouldn't it be reasonable to move it into a separate template which accepts boolean observable?

chrisknoll: We can do that.

chrisknoll:
@pavgra : I've moved the duplicated code into a new Input component: CycleToggleInput.

js/components/cohortbuilder/components/CycleToggleInput.js
```diff
@@ -0,0 +1,31 @@
...
+		
+		self.value = params.value;
+		self.options = params.options;
+		
+		self.currentIndex = ko.pureComputed(() => getOptionIndex(self.value(), self.options));
+		self.selectedOptionText = ko.pureComputed(function() {
+			return self.options[self.currentIndex()].name;
+		});
+		
+		self.nextOption = function () {
```
pavgra: Looks too similar to `js/components/cohortdefinitionviewer/components/CycleToggleInput.js`. Why not to move common parts out / have single parametrized component?

PR #780, opened 2018-07-23
Refactored cohort-definitions page

johnSamilin:
Part of #633


js/services/http.js
```diff
@@ -73,28 +12,6 @@ define(function(require, exports) {
...
-          break;
-        default:
-          this.handleUnexpectedError();
-          break;
-      }
-
-      return false;
-    }
-
     sendRequest(method, path, payload) {
```
pavgra: Why was this method left but not inherited?

js/services/http.js
```diff
@@ -127,26 +40,11 @@ define(function(require, exports) {
...
-    }
-
-    doDelete(path, payload) {
-      return this.sendRequest(METHODS.DELETE, path, payload);
-    }
   }
 
   const singletonApi = new Api();
-  singletonApi.setApiHost(config.api.url);
+  // singletonApi.setApiHost(config.api.url);
```
pavgra: What is this for?

js/services/http.js
johnSamilin: @pavgra because of this line:
`headers: this.isSecureUrl(path) ? this.getHeaders() : {},`

chrisknoll:
Ok, this comes at an interesting moment: I'm about to make some changes to the Inclusion Rule report as part of OHDSI/circe-be#47 which will involve some UI changes, that I'd like to base on this refactoring and not try to squeeze it in before.  So, I will try to review and get these refactor changes into master as soon as possible.  I'm also reviewing changes for OHDSI/Atlas#647 which have some changes related to cohort definitions, and I'm worried that those code changes will be in conflict with these changes here.

How would you like to proceed: should I finish the work on OHDSI/Atlas#647 and then you rebase/merge those changes into the refactoring of this PR, or do this PR first, and then refactor your diagnostics work into master after this PR is complete?

chrisknoll:
Re: the ```assets/``` folder:  I'm not sure if this is introduced per this PR or if we always had an 'assets' folder since the start of the refactoring, but I'm just questioning now the name of this folder:  I think of 'assets' as resources (like CSS, LESS, images, templates, etc) but it looks like those things that you placed into 'assets' are libraries (like, clipboard, colorbrewer, etc).  I other projects, I've seen these 'externally developed libraries' under a ```vendor/``` folder...not sure I really like that name, but 'assets' is not exactly what I expected.  Maybe ```libs/```?  Open to your thoughts on this. I apologize for nit-picking on a name, but I don't want to have confusion later about the purpose of the ```assets/``` folder is.

pavgra:
@chrisknoll , I've resolved conflicts and pressed approve button

Issue #719, opened 2018-05-25
Atlas - Cost and utilization - heracles - select analysis id to run


gowthamrao:
### Expected behavior
Able to select what reports to run for (cost and) utilization
Depending on selected reports, the analysis id's are populated
Less analysis id is expected to take shorter run time

### Actual behavior
all reports and all analysis id in (cost and) utilization
All analysis id's requires longer run time

### Details - allows to run only required analysis id
In 2.4 Utilization runs all of the following analysis id's. Would like the option to select 'report' and based on the selection will run analysis_id's

[analyses ids used in cohort reports.xlsx](https://github.com/OHDSI/Atlas/files/2040190/analyses.ids.used.in.cohort.reports.xlsx)

gowthamrao:
@pavgra 

heracles_periods has 'ww', 'mm', 'qq', 'yy' for weekly, monthly, quarterly, yearly. Would like the pop-up option to specify and ask if the user would like to run 'ww', 'mm', 'qq', 'yy' . This will allow end-user to restrict analysis to what is needed, and not run analysis not required - improving run-times.

PR #793, opened 2018-07-26
Added cost util period type selection

agackovka:
fixes https://github.com/OHDSI/Atlas/issues/719


js/components/visualizations/filter-panel/filter-panel.js
```diff
@@ -19,22 +19,25 @@ define(
...
-              }
-            });
-          }
-        });
+        this.filterList.subscribe(newFilter => {
+          Object.values(newFilter).map(filter => {
+            let prevVal;
+            if (filter.type === 'select') {
+              prevVal = filter.selectedValue();
+              filter.selectedValue.subscribe(() => {
```
johnSamilin: subscribers should be disposed

chrisknoll: There should be a few examples in the codebase where you can see how a 'dispose' method is declared on the model, which will be triggered when the view is cleaned up via knockout.  As @johnSamilin says, subscriptions should be cleaned up, but also ko.computed (but not ko.pureComputed), etc.

js/pages/cohort-definitions/components/reporting/cost-utilization/base-report.js
```diff
@@ -175,9 +176,18 @@ define(
         }
         this.loading(false);
       }
+      
+      async initializePeriods() {
+        try {
+          this.periods = await CohortResultsService.loadPeriods({ source: this.source, cohortId: this.cohortId, window: this.window });
+          this.filterList(this.getFilterList());
```
pavgra: I'd move this outside, as separate step

PR #797, opened 2018-07-30
Structure refactoring/incidence rates

johnSamilin:
part of #633


chrisknoll:
@johnSamilin ,

Something changed in the Incidence Rate API where when we used to get the report data back from this call:
```
IRAnalysisService.getReport(source.info().executionInfo.id.analysisId, source.source.sourceKey, this.selectedTarget(), this.selectedOutcome()).then((report) => {
```

But now the resulting data structure is soemthing resembling an XHR request:

image

The data is buried somewhere under the 'data' field, but the api is just supposed to return the data, not the low level details of the ajax request.  Can you change this to have the previous behavior?

johnSamilin:
@chrisknoll sure, fixed it

chrisknoll:
@johnSamilin , I pushed up some fixes:

1.  There was a reference to a css style found in another module so that was just an update.
2. More serious: pure computeds that need to access the view model (formerly, this was handled via a 'self' reference) were having 'this' result in 'undefined' when triggered by an asynchronous update in the ui (ie: one of the observables notified a change).  Converting the function declaration to arrow function allowed the 'this' ref to be resolved to the view model, but I wanted to call this out for future changes to make sure that functions that reference view models via 'this' are probably bound.  I like what we're doing with the refactoring, but this element of ES6 classes is making me a bit unhappy.

If you don't see any problems with these changes, I can approve and merge it in.

johnSamilin:
Yes, seems like I just missed those two functions. When I analyse the code of the component, I convert the callbacks from regular functions to arrow ones to preserve its context.
I'm not quite sure that I understood the second point correctly, you wouldn't want to use arrow functions or to bind class methods to `this`?

chrisknoll:
> I'm not quite sure that I understood the second point correctly, you wouldn't want to use arrow functions or to bind class methods to this?

No, the arrow-function style is fine with me.  It's just that we need to make sure that we're are using arrow functions where the function might be called async and 'this' becomes undefined.  So, yes, use arrow functions because they handle the 'this' binding to the view model automatically.

PR #800, opened 2018-08-01
Bundling of external libs

### LINKED ISSUE #628: ATLAS UI fails to load JS resources
### Expected behavior
All JS resources are loaded properly by browser.

### Actual behavior
In some corporate networks with proxy/firewall unpkg.com host is not in whitelist.
It's the reason that
https://unpkg.com/@ohdsi/atlascharts@1.1.0/dist/atlascharts.min.js file is not loaded and UI fails to start.

### Suggested fix
Include atlascharts.min.js file in local static resources. OR give a possibility to override require.js configuration file to re-set the library paths.

johnSamilin:
fixes #628 
requires node v8.9.4+


chrisknoll:
With this enhancement, we really need detailed instructions on how to perform the bundle. Can you add something to README.MD that can provide instructions on how bundling is performed?

pavgra:
@chrisknoll, just to be clear - nothing changes in app's behavior by default. Everything should continue to work as it did before. Just an additional option for combining external resources into a single local file appears. For that, sure, @johnSamilin will update README

chrisknoll:
Understood, @pavgra, that's what we agreed to: async loads during dev cycle, build/optimize/bundle for deployments.  

I appreciate @johnSamilin doing this, it would be good to know what the prescribed approach for creating the app bundle is.

PR #804, opened 2018-08-01
Added users import from Active Directory

wivern:
closes OHDSI/WebAPI#506


js/modules/WebAPIProvider/UsersAPI.js
```diff
@@ -0,0 +1,20 @@
+define(function (require, exports) {
```
pavgra: We agreed to services in `js/services` folder

js/modules/WebAPIProvider/UsersAPI.js
```diff
@@ -0,0 +1,20 @@
...
+
+  const $ = require('jquery');
+  const config = require('appConfig');
+  const authApi = require('webapi/AuthAPI');
+
+  function getAuthenticationProviders() {
+    return $.ajax({
```
pavgra: Use `services/http`

js/pages/configuration/users-import/components/atlas-roles.html
```diff
@@ -0,0 +1,23 @@
+<div>
+    <table width="100%" cellspacing="0" data-bind="dataTable:{
+                            data: roles(),
+                            options: {
+                                        dom: 'Bliprt',
```
pavgra: What is this for?

js/providers/Users.js
```diff
@@ -0,0 +1,76 @@
+define(function (require, exports) {
```
pavgra: Why is this separate from `js/modules/WebAPIProvider/UsersAPI.js`? + same comment about `services` folder

js/providers/Users.js
```diff
@@ -0,0 +1,76 @@
...
+  function getAuthenticationProviders() {
+    return $.ajax({
+      url: config.webAPIRoot + 'user/providers',
+      method: 'GET',
+      error: authApi.handleAccessDenied,
+    });
+  }
+
+  function searchGroups(provider, search) {
+  	return $.ajax({
```
pavgra: use `http` service

js/pages/configuration/users-import/components/atlas-roles.html
wivern: It`s a wondering feature of [datatables](https://datatables.net/reference/option/dom),
for example, B means Buttons

PR #809, opened 2018-08-03
Added flex to section/sections styles

agackovka:


js/components/modal-pick-options.less
```diff
@@ -5,15 +5,21 @@
 
   &__section {
     margin-bottom: 2rem;
+    flex-basis: 50%;
```
pavgra: flex: 0 0 50%;

js/components/modal-pick-options.less
```diff
@@ -5,15 +5,21 @@
...
+    color: black;
+    font-weight: 600;
   }
   
   &__sections {
     list-style: none;
     padding: 0;
+    display: flex;
+    flex-direction: row;
+    justify-content: space-between;
```
pavgra: +
flex-wrap: wrap;

PR #808, opened 2018-08-03
Structure refactoring/profiles

johnSamilin:
part of #633

PR #814, opened 2018-08-07
Issue 806 cannot navigate directly to concept

### LINKED ISSUE #806: Cannot navigate directly to concept
### Expected behavior
Navigate directly to a concept using the concept URL: http://localhost/atlas/#/concept/4294416 and the concept's information is properly displayed

### Actual behavior

Navigating directly to a concept using the code currently on master errors with 

`GET: http://localhost/atlas/undefinedconcept/4294416 404 (Not Found)`

And the page does not load

See above. As a note, I'm doing this without security enabled. I've traced through the code and have determined that part of the problem is that the sharedState (/components/atlas-state.js) has not been initialized at the time when js/pages/concept-sets/routes.js is loaded. This is the reason for the "undefined" portion of the offending REST call above. 

Previously, routes were loaded after the sharedState initialization which prevented this type of an issue. With the refactoring, it appears that this loading sequence is compromised thus resulting in the above error.

Tagging @pavgra, @wivern and @johnSamilin for input.

johnSamilin:
1. Get rid of unneccessary dependency injection (views can access global model methods & properties through $root, controllers have a reference to a model through `current component` binding)
2. Separated global model definition, app bootstrapping & router setup
3. Added AutoBind provider to automatically preserve context in class methods
4. Added `synchronization` step to an app's bootstrap process
5. Removed a lot of unused code
6. Routes fixed #806 
also, part of #694


pavgra:
@anthonysena , thank you a lot for testing it! @johnSamilin will take care of it. Just a couple of comments from my side:

> I moved this into the root js folder and updated the settings.js and main.js to reflect this

I want to push back on this and address the root of the issue to keep things structured and not ruin effort spent for the Atlas refactoring (versus storing all files in one folder)

> Defining a new cohort attempts to request "js/bootstrap-select-css.css" which is not found

Have seen that too. For some reason, the dependency was commented out in requirejs config.

anthonysena:
Thanks @pavgra - happy to put optional.js back under plugins if you guys can figure out why it is throwing the error. I tried to figure it out without success and opted for the approach that I committed. I'd be interested to see how you guys address the issue and keep the resource under the /plugins folder.

chrisknoll:
@johnSamilin : there's many conflicts, so please resolve.

Also:  related to the optional plugin, i was going to fix that separately, but you are tryign to do it here.

What I recommend is:

1. Put the plugin back in /plugins.
2. Add this block of code before the first requireJS call in main.js:
```
  requirejs.config({path: "optional":"plugins/optional"});
```
This will allow you to resolve the 'optional' plugin for the first require call, then subesequent reuqire calls will be based off of the udpated require.config() call that's made in the callback.

I tested this locally, and it seemed to eliminate the 'unable to load optional' error.  The reason this error was happenign was that there was a race conditon: you were loading the 'optional' plugin via 'plugins/optional' but then another dependency being requested wias 'config-local' which was accessing the 'optional' plugin which was causing a conflict.  The way you fixed it will work (now optional needs to be referenced as 'plugins/optional'), but I'd prefer it if all plugins were stored in 'plugins' and we create the path int he config to resolve 'optional' to the path to the optional plugin.

chrisknoll:
Ok, I pushed up my change which should give us best of both worlds: the plugin is in the /plugins folder, and we don't get an error on startup with conflict in the initial require() call.

We still have several conflciting files with master, please resolve these.

pavgra:
@johnSamilin, I've pulled your latest changes and iterated through the list of issues reported by @anthonysena. Results are below.

> If you navigate directly to a search (i.e. http://localhost/atlas/#/search/searchtermnotfound) the request fails since the vocabulary providers are not loaded.

Cannot reproduce now.

> You cannot import a concept set - all of the options fail.

Still reproducable. E.g. go to Search -> Import -> Concept Identifiers -> Enter "1112807" -> Import -> No concepts are displayed in new Concept Set. While it works at http://www.ohdsi.org/web/atlas/

> Opening an existing concept set also fails

Still reproducible:

image

> The page also calls "WebAPI/cohortdefinition/0/check" which throws an error.

Still reproducible. Open new Cohort page, change name, press "Save":

image

> After defining a new cohort, if you open the concept set editor, you get an error and you cannot see any of the concepts

Cannot reproduce now.

> I also noticed that when the cohortdefinition list screen is opened (atlas/#/cohortdefinitions) it appears to call the check service for each cohort displayed?

Cannot reproduce now.

chrisknoll:
Hey, everyone:
  Can someone please merge from master and resolve conflicts. One source of conflicts will be from the fix for the 'optional' module path that I put into a different PR: #808.  You can just take the change that was in that PR since it fixes the JS error.

Thank you!

anthonysena:
Still reviewing these changes - the issues I raised in my first review appear to be resolved so thank you for that! I just pushed a few small fixes that address 2 bugs: the 'included source codes' tab was throwing an error related to the fact that sharedState was being referenced in /js/const.js but it was never declared. Also, I made a change to /js/providers/Model.js since this was failing when looking up included concepts when loading up a cohort definition.

One new behavior I just noticed: I view a concept set or cohort definition, make no changes and attempt to close it and I'm prompted to save. I have to dig into what is happening with the dirtyFlag() for these objects but it appears that this is a new bug.

anthonysena:
OK - spent a good deal of time trying to debug this particular issue and am requesting help from @pavgra @johnSamilin. The bug I'm facing is that when you open a concept set, the dirtyFlag on the app model is being interpreted as being dirty when it should not and as a result, users will be prompted to save changes when they have not made any.

The /js/components/conceptset/conceptset-manager.js is referencing the params.model.currentConceptSetDirtyFlag which should be set at the app level. However, when debugging it appears that the object that is passed into the component as "params.model" is not the same as the GlobalModel defined in /js/providers/Model.js. 

Part of me wonders if this is due to the fact that /js/components/conceptset/conceptset-manager.js does not use the new ES6 approach to create a class which extends Component? Please let me know if you have any recommendations on this otherwise I'm not sure how to best proceed.

chrisknoll:
While I was working with @anthonysena , I noticed that we're using the spread operator ```...``` when passing in the paramaters to the component, and wasn't sure if that could be manipulating the object refereces in an unexpected way.  It seems sometimes the object reference was to a 'GlobalModel' but other times it was an untyped object..

I'm referring to here:
https://github.com/OHDSI/Atlas/blob/issue-806_Cannot_navigate_directly_to_concept/index.html#L131

The use of the spread operator has changed since master.  I'm not sure that this is the cause of some of the issues, so apologies if this is a red-herring.

chrisknoll:
@anthonysena , I pushed a change to make the conceptSetDirtyFlag an observable (instead of a simple property). i think what was happening was that, using just the simple refernce in the object, if you Have this type of code:

```
var a = model.dirtyFlag;

...

model.dirtyFlag = new dirtyFlag();

alrt(a == model.dirtyFlag) ; // will be false
```

What I believe was happening int he code was a reference to one dirty flag was captured via model.dirtyFlag.  Then the dirty flag was reassigned, and the former references to model.dirtyFlag maintained the first reference.  So, instead I made the dirtyFlag a n obserable() in the model, and getting/setting the dirty vlag via model.dirtyFlag(newVal) ensures that if you get a reference to dirtyFlag, de-referencing it gets the current reference, not a stale reference.

anthonysena:
Thanks @chrisknoll that resolved the issue with the dirtyFlag on the concept set! I made a small update to the code to fill in some gaps where there were a few places where we were still referring to a non-observable dirtyFlag.

Next issue: when on a concept set page, the call to resolve a concept set is not working. To reproduce the issue on this branch: create a concept set with the concept_id = 201826 => type 2 diabetes. Click on the 'descendants' checkbox. You should see the # of included concepts go up to 18 (your number might differ based on your vocab version). Tracing through this tonight, here's what I found - the checkboxes on this page are rendered by this function in /js/Model.js: 

https://github.com/OHDSI/Atlas/pull/814/files#diff-c89c9457c900ca77bee8ad1d7c19d785R394

Within this function on line 394, you will see that the click binding is set to call the 'resolveConceptSetExpression()' which is also defined in Model.js. Stepping into that function reveals another function called 'resolveConceptSetExpressionSimple(expression, success)': https://github.com/OHDSI/Atlas/pull/814/files#diff-c89c9457c900ca77bee8ad1d7c19d785R373. It looks like the anonymous function defined when the 'success' parameter is undefined has a bug, I'll try to look into this a bit more tomorrow but wanted to point this out in case others have a fix here.

js/services/http.js
```diff
@@ -26,7 +27,7 @@ define(function(require, exports) {
       if (payload && payload instanceof FormData) {
         params.body = payload;
       } else if (payload) {
-        params.body = JSON.stringify(payload);
+        params.body = ko.toJSON(payload);
```
chrisknoll: So, this is an interesting change:  one one hand, I'd argue that the http service does not need to have a dependency on knockout, and it should be expected that the callers passing in their payload perform the appropriate ko.toJS() on their object before sending it over to the http service.

johnSamilin:
@anthonysena indeed, previously this function (`resolveConceptSet`) used `ko.toJSON` to serialize data. With introduction of the `@ohdsi/ui-toolbox` package the data serialization task moved to `httpService`. It used `JSON.stringify`, which turned to be not the same as `ko.toJSON`. I guess now this bug should be fixed

js/services/http.js
johnSamilin: On the other hand, it's much better to have one single point of truth for data preparing

anthonysena: Thanks @johnSamilin - I had actually made a fix in the way that @chrisknoll described whereby I used ko.toJS() further upstream which fixed this. I'm concerned that changing this in the http service might have other unintended consequences?

anthonysena: Great - sounds good in terms of compatibility. To @chrisknoll's other point: not sure if knockout is already a dependency in the @ohdsi/ui-toolbox in which case using it in the http service works. However, if you are establishing the knockout dependency to address this problem, I'd suggest we instead fix this upstream in the Atlas code. When I used ko.toJS() to unwrap the observable before posting to the REST endpoint that fixed the problem without having to touch the existing http service.

chrisknoll: Hi, the ko.toJS (not toJSON) is to strip the observables away and populate just the raw values.  This isn't about strigify.  What you need to do, when submitting the code over HTTP, is to stringify.  All I was suggesting is that pepole should know if they use HTTP service, they should take care of de-observable their models and the HTTP service should not take any additional steps on behalf of the caller.  This makes it more generic, and less dependency laiden.

anthonysena: Agreed @chrisknoll - if the http service sticks with the original approach using JSON.stringify that works for me. In the specific case of the resolveConceptSetExpression function, adding the ko.toJS() was necessary to populate the raw values prior to POST-ing the data via the http service which solved the problem I reported last night.

johnSamilin: But ko.toJSON does just the same. Doing so in httpService is much easier and will lead us to less boilerplate. By the way it will reduce the chance of a wrong behaviour in the future beacuse it's dramatically easy to forget to call ko.toJS before sending the data to httpService

chrisknoll: That's true, @johnSamilin , that there's no harm in calling  ```ko.toJSON``` even on plain old javascript objects.  And it will keep people from making the mistake of not calling ko.toJS() before calling a http service method.

All I'm saying is:  1) if you do it your way, then if you wanted to export the HTTP module and use it in another app. you've made the other app dependent on KO.  What if you wanted to make this a stand-alone module for a React app? Wouldn't it be strange to have a KO dependency in a React app?   And 2) What if there are other frameworks at play that need special handling for their javascript objects prior to being sent over HTTP? Do we need to bundle all possible use-cases to handle the proper javascript structure in the HTTP service, or put it on the plate of the person calling the HTTP method?

The pro of your approach is it catches the specfic knockout usecase error. The con of it is that it may not be completely comprehensive (point 2 above) and it introduces a dependency that may add a burdeon of re-usability in different applications if we ever want to pull the HTTP service modules out into a shareable library.

Because of the tight deadlines we're facing right now, I don't think it's worth spending any more time changing the code. But, we should keep these points in mind in the future if either point 1 or 2 I made above become a real issue.

pavgra: @chrisknoll , we already have the http service as a standalone module (get a look at `const OHDSIApi = require('ohdsi-api').Api;`) which is used in Athena (React app). And the module obviously doesn't depend on Knockout. What you see here is child class which extends the module and is required to handle specifics of Atlas (so I'd call it "adapter" in terms of patterns). And as for me, it looks very reasonable to have application-specific `ko.toJSON` here. So I am confused where is the problem

chrisknoll: Thanks @pavgra : the piece i was missing is that the Http serivce in atlas was an extention to a shared library that was created specifically for atlas.  IN that case, my points about depenency and reusability are moot.

Thanks for the clarification.

js/providers/Model.js
```diff
@@ -461,7 +461,7 @@ define(
 				// a respository. If it is dirty, prompt the user to save and exit.
 				if (this.currentConceptSet()) {
 					if (this.currentConceptSetSource() == 'repository') {
-						if (this.currentConceptSetDirtyFlag && this.currentConceptSetDirtyFlag.isDirty() && !confirm("Concept set changes are not saved. Would you like to continue?")) {
+						if (this.currentConceptSetDirtyFlag() && this.currentConceptSetDirtyFlag().isDirty() && !confirm("Concept set changes are not saved. Would you like to continue?")) {
```
chrisknoll: Btw, this.currentConceptSetDirtyFlag() is guarunteed to be non-null (it's initalized at application startup) so, you shouldn't need to do the existance check, but  there's no harm in doing it. Just FYI.

anthonysena:
@johnSamilin I'd recommend you take a look at the changes in https://github.com/OHDSI/Atlas/pull/814/commits/8acf82ada692190bf97f4e7ecaf1f03a28a5840c since I refactored the way we initialize sources. Here's what changed:

- /js/Application.js: added a check to determine if authentication is enabled. When auth was disabled, the initSourceConfig() and the promise was resolved immediately which was causing sync issues.
- The code for caching data sources was broken and there was redundant code between /js/Application.js and /js/modules/WebAPIProvider/SourceAPI.js which is now encapsulated in a new method "setSharedStateSources"
- Data source caching is now set through /js/config.js and by default is set to 'false'. 

Let me know if you have any issues but this now resolved #806 from my side.

pavgra:
@anthonysena, thank you for the time you've spent for the PR! Great that this can be finally marked as fixed. I'd suggest proceeding with the merge, so that we don't get into one more iteration of conflicts. And @johnSamilin can work on any more changes, if required, in a new branch (PR).

PR #815, opened 2018-08-07
CommonEvidenceModel v1.0

anthonysena:
Front-end changes to support CEM v1.0 - needs to be tested with: 

https://github.com/OHDSI/WebAPI/pull/537


anthonysena:
A few notes on these Atlas related changes: 

- As mentioned in https://github.com/OHDSI/WebAPI/pull/537#issuecomment-411511336, the evidence diamon has been renamed to 'cem'.
- A new configuration section has been added to config.js to support the ways in which you may retrieve evidence details from CEM: https://github.com/OHDSI/Atlas/pull/815/files#diff-de645108406065aec943815ce4e8e0b4R90. 

Please let me know if you have any questions/concerns.

chrisknoll:
I think @pavgra  might have some comments about how these components are structured.

I believe these components can be considered 'stand alone' (ie: they aren't bound to one specific page of the Atlas UI).  So making a folder for these components under /components makes sense.  We're trying to use the class model @pavgra and @wivern and @johnSamilin introduced, perhaps they can give some pointers on how this code should be structured to follow those guidelines?

anthonysena:
Thanks @chrisknoll - appreciate the careful review and look forward to feedback from @pavgra, @wivern and @johnSamilin since this is my first attempt to follow the new structure.

chrisknoll:
Looks good, @anthonysena , do you think it's worth taking those 'options' (like options, columns etc) into a seperate module that's a local moduel to your component (you can access as ./options for example) and you can remove some of the verbose option code that is in your model into a separate file, cleaning it up a bit, and centeralizing some of these elements that could be encapsulated.

PR #840, opened 2018-08-16
Structure refactoring/estimation

johnSamilin:
part of #694


js/pages/estimation/cohort-comparison-manager.html
```diff
@@ -1,622 +1,620 @@
...
-
-    <div role="tabpanel" data-bind="css: { active: $component.tabMode() == 'results' }" class="tab-pane">
-      <div class="padded-wrapper" data-bind="visible:$component.resultsMode() == 'sources'">
-        <div class="heading">Available Data Sources</div>
-        <div data-bind="foreach:$component.sources" class="results-tab-pane">
-          <div class="row">
-            <div class="col-sm-8" data-bind="text:sourceName"></div>
-            <div class="col-sm-4 text-right">
-              <span class="btn-group">
-								<button data-bind="css:{disabled: $component.config.api.isExecutionEngineAvailable()==false}, click:function() {$component.executeCohortComparison(sourceKey)}" class="btn btn-sm btn-primary"><i class="fa fa-play"></i> Generate</button>
```
anthonysena: Getting an error on this line due to the fact that config.api.isExecutionEngineAvaiable() is not a function. I see that this is set in the /js/Application.js but not sure why the observable is no longer available at this point?

anthonysena:
I pushed some fixes to this branch - I cannot approve these at this time since importing concept sets is failing when attempting to call the REST end point /WebAPI/vocabulary/conceptSetExpressionSQL and I'm unsure what's happening. I'd also ask we consider holding off on any structure/refactoring around estimation and PLP since I plan to make some significant enhancement to both of these modules and at that time I can restructure things.

pavgra:
@anthonysena , if you could follow the new project structure and apply adopted ES6 (class based) approach, we can switch to other things, sure.

js/pages/estimation/cohort-comparison-browser.html
```diff
@@ -11,8 +11,5 @@
 		<faceted-datatable data-bind="visible: loading()==false" params="orderColumn: 3, reference: reference, columns: columns, options:options, rowClick:rowClick"></faceted-datatable>
 	<!-- /ko-->
 
-	<div data-bind="visible:loading()" class="loading">loading</div>
-	<div data-bind="if: !reference().length, css: classes({ element: 'empty-state', extra: 'flexed' })">
-		<empty-state params="message: 'No PLE analyses created yet'"></empty-state>
-	</div>
```
johnSamilin: why empty state is being removed?

js/pages/estimation/cohort-comparison-manager.html
```diff
@@ -556,8 +556,8 @@
         <ul class="nav nav-pills">
           <li role="presentation" data-bind="css: {active: $component.expressionMode() == 'print' }, click: function() {$component.expressionMode('print') };"><a>Print Friendly</a></li>
           <li role="presentation" data-bind="css: {active: $component.expressionMode() == 'rcode' }, click: function() {$component.expressionMode('rcode') };"><a>R Code</a></li>
-          <li role="presentation" data-bind="css: {active: $component.expressionMode() == 'import' }, click: function() {$component.expressionMode('import') };"><a>Import</a></li>
-          <li role="presentation" data-bind="css: {active: $component.expressionMode() == 'export' }, click: function() {$component.expressionMode('export') };"><a>Export</a></li>
+          <li role="presentation" data-bind="visible: false, css: {active: $component.expressionMode() == 'import' }, click: function() {$component.expressionMode('import') };"><a>Import</a></li>
```
johnSamilin: why would we hide those tabs?

johnSamilin:
@anthonysena I'm already finishing PLP refactoring, could you make those enchancements after it's reviewed & merged?

anthonysena:
@johnSamilin - I will attempt to make my enhancements after merging these since you've done most of the refactoring. I'd kindly ask that you test each module with the developer console open and correct any errors for both PRs. That includes: 

1. Creating a new estimaton/plp specification, filling in all inputs and saving it
2. Opening an existing estimaton/plp specification, modifying some of it and saving it
3. Copy a specification (PLP only) and the deleting it
4. Verifying that the print friendy and code export tabs work

I know your team is testing these things with security enabled which is great - I'm happy to repeat this testing locally where I have security disabled and correct any errors. 

Yesterday when testing I was encountering an error when choosing a concept set whereby the call made to the REST endpoint: /WebAPI/vocabulary/conceptSetExpressionSQL via the new VocabularyProvider failed. If we follow the testing steps above for both PRs, these issues will be caught and will save some time prior to PR review.

Please let me know if you have any questions or concerns. Appreciate all of your efforts around this. Thanks!

anthonysena:
@johnSamilin and @pavgra - Dug into this some more and found the problem. The problem I ran into is pretty deep in the stack so I'm going to detail my findings. First, the rest endpoint /WebAPI/vocabulary/getConceptSetExpression produces TEXT_PLAIN instead of APPLICATION_JSON. 

https://github.com/OHDSI/WebAPI/blob/master/src/main/java/org/ohdsi/webapi/service/VocabularyService.java#L693

Noting this behavior, I traced the attempted request sent to the REST endpoint via ATLAS I discovered that an "ACCEPT" header item was added to only accept application/json which then causes the server to send back an HTTP 406 UNACCEPTABLE error.  I further traced this down to the shared UI component here:

https://github.com/OHDSI/UiToolbox/blob/master/src/api/index.js#L17

Note that in this module the header is set to only accept application/json. Should we consider also allowing for text/plain as an acceptable media format? There are a few other REST endpoint on the WebAPI side that utilize this media response type. Please let us know your thoughts.

chrisknoll:
Sena, I believe all of our services are going to be producging APPLICATION/JSON. So I think the bug is in th getConceptSetExpression service where it should be reporting it returns JSON.

There are other cases, tho, like export a CSV or XLS, where the server will not return JSON, so from a 'generic toolbox' perspective, it should be flexible to allow reqeusts for different content types....but on the Atlas calling side, it should specify applciation/json or we should default to application/json. I don't know any case where we're just producing text...

johnSamilin:
@anthonysena @chrisknoll 
I've tested PLE using the list of steps you provided, for both secure and public modes. PLP will be tested in another PR.
I also did some changes to httpService: added a new property, `plainTextService`, which will send `Accept: text/plain` header and properly handle the response. For this I had to change `@ohdsi/UIToolbox` a little bit, see `https://github.com/OHDSI/UiToolbox/pull/1/files`

PR #847, opened 2018-08-17
Structure refactoring/prediction

johnSamilin:


chrisknoll:
@johnSamilin , 
Before we do our own review, please confirm you've tested the following use cases:
1. Creating ne analysis
2. Updating analysis
3. Deleting Analysis
4. Cohort Selection (both target and outcome selection)
5. Included Conceptset and Excluded Conceptset selection

Throughout verifying these functions, please have your browser console open, and verify no errors occur in the log.

johnSamilin:
@chrisknoll I've tested for both public and secure modes. The only errors in the console are for components `plp-roc', `plp-spec-editor` and `plp-inspector` (related to `d3`). Keeping in mind that they are always hidden, I'm not sure whether we need to fix it now

chrisknoll:
Thanks for testing this.  I agree the hidden components probably don't need to be addressed unless they break other functions   in the app.  I can review this after Sept 5th, unless someone else can pull this and review it first.

js/pages/prediction/components/plp-inspector.js
```diff
@@ -1,4 +1,4 @@
-define(['knockout', 'jquery', 'text!./plp-inspector.html', 'appConfig', 'd3', 'css!./styles/plp.css'],
+define(['knockout', 'jquery', 'text!./plp-inspector.html', 'appConfig', 'd3', 'css!../../../styles/plp.css'],
```
pavgra: @anthonysena , since the `styles` directory is in the app root, let's use an absolute path (and, in general, three folders back doesn't looks to be very solid approach, easy to break accidentally)

anthonysena: Agreed - made that change. Thanks!

pavgra: @anthonysena , thank you. I also see such thing in `js/pages/prediction/components/plp-roc.js`

anthonysena: Missed that - think that should resolve this. Let me know if you have any other concerns.

anthonysena:
@pavgra @johnSamilin - I was hoping we might pull this branch into master. I did another test of this branch and it appears that merging in some of the changes from master have introduced some regressions. I am now getting an error when selecting a concept set in the PLP designer. Could you take a look? Also, I'm noticing an odd behavior that when ATLAS starts up, the ATLAS logo is zoomed in and then zooms out to its normal size. Not sure if this is a byproduct of the CSS bundling changes or not but wanted to see if was only a problem on my side or if you can reproduce it.

Please let me know so we can get this merged in as I'd like to have these changes in master so I can ensure that the new PLP editor follows the same structure. Thanks!

johnSamilin:
@anthonysena I've fixed this error, it turnes out one component (`plp-editor`) needed some changes.
I also fixed size of the logo, yes, it was because of the css bundling

anthonysena:
Thanks @johnSamilin - the concept set selection is now working. However, I'm now unable to get the cohort selector to work? Here are the errors I've received so far:

image

Error 1: This resource is not found: http://localhost/atlas/js/modules/plp/main.css?_=k3uuxt 
Error 2: Line 40 in the plp-spec-editor.js throws an error since cohortDefinition is undefined.

Could you take a look? Thanks.

johnSamilin:
@anthonysena I fixed that and checked the process of creation of the PLP analysis. Everything seems to be ok now

johnSamilin:
@anthonysena this is already in master https://github.com/OHDSI/Atlas/blob/master/js/settings.js#L128

PR #849, opened 2018-08-17
Fixed Included Concepts is not updated when Descendants checkbox is turned on / off

### LINKED ISSUE #848: Count in Concept set -> Included Concepts is not updated when Descendants checkbox is turned on / off
count-not-updated

wivern:
Fixes #848

### DIFF
--- a/js/components/conceptsetInclusionCount/conceptsetInclusionCount.js
+++ b/js/components/conceptsetInclusionCount/conceptsetInclusionCount.js
@@ -23,6 +23,11 @@ define([
 
 			this.getInclusionCount = this.getInclusionCount.bind(this);
 			this.conceptSetExpression.subscribe(this.getInclusionCount);
+			this.conceptSetExpression().forEach((item) => {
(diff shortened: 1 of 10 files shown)


js/components/conceptsetInclusionCount/conceptsetInclusionCount.js
```diff
@@ -23,6 +23,11 @@ define([
 
 			this.getInclusionCount = this.getInclusionCount.bind(this);
 			this.conceptSetExpression.subscribe(this.getInclusionCount);
+			this.conceptSetExpression().forEach((item) => {
+				item.includeDescendants.subscribe(this.getInclusionCount);
+				item.includeMapped.subscribe(this.getInclusionCount);
+				item.isExcluded.subscribe(this.getInclusionCount);
+			});
```
pavgra: @wivern , cannot we apply more universal `ko.computed(() => ko.toJSON(root)).subscribe(this.getInclusionCount)` instead of 6 lines above?

PR #861, opened 2018-08-23
Fixes Job Polling (Issue #825)

### LINKED ISSUE #825: Cohort Reporting is not polling for status properly
### Expected behavior
As you run a cohort reporting analysis (either full or Utilization), the status of the job should progress through Starting ... Running..Complete.

### Actual behavior
Job status does not seem to progress through 'Starting'.

1. Create a cohort that has some episodes in it.
2. Go to cohort reporting, and select full analysis.  Note: when this error occurred, there already was a prior analysis executed.
3. Monitor jobs through the page or on the notification dialog...the jobs never show 'complete' even after the job completes on the back end.

Note: I'm putting this as an Atlas issue, because I believe it's an issue with polling for the job status and updating the UI.  The back end seems to report that the job is complete.  note: if you refresh the browser completely, the page shows the jobs are no longer running and you can view the reports.

anthonysena:
Fixes #825 and adds in the jQuery UI dependencies that were missing and throwing errors in the developer console.


js/components/user-bar.js
```diff
@@ -48,6 +57,10 @@ define([
 			this.updateJobStatus = this.updateJobStatus.bind(this);
 			this.clearJobNotifications = this.clearJobNotifications.bind(this);
 			this.clearJobNotificationsPending = this.clearJobNotificationsPending.bind(this);
+
+			if (!appConfig.userAuthenticationEnabled) {
+				this.startPolling();
```
pavgra: @anthonysena / @chrisknoll , how is it supposed to work in secured app mode?

anthonysena: The subscription here: https://github.com/OHDSI/Atlas/pull/861/files#diff-704b7a14b7c7b575513079f09614f96dR39 will handle the situation where security is enabled. The bug that this change addresses is that we were not polling for updates when security is disabled. When security is enabled, the subscription will start polling for updates once the user is logged in.

pavgra: @anthonysena , ouch, my bad - will be more attentive next time. Thanks for pointing me

PR #871, opened 2018-08-27
fixed reloading of the report

### LINKED ISSUE #864: Data Sources: after dataset is changed report is not reloaded on the page.
### Expected behavior
Report is updated.

### Actual behavior
Report is NOT updated.

1. Select a source
2. Select a report -> Condition.
3. Make sure that report is loaded.
4. Choose different data set in drop down

johnSamilin:
fixes #864


chrisknoll:
If this PR is being replaced with another branch, please clean up this branch if it is no longer used.

PR #872, opened 2018-08-27
ATLP-466. Fixed table format issues

aklochkova:


js/providers/Model.js
```diff
@@ -549,7 +551,7 @@ define(
 										cdsi.sourceId = sourceInfo.id.sourceId;
 										cdsi.status = ko.observable(sourceInfo.status);
 										var date = new Date(sourceInfo.startTime);
-										cdsi.startTime = ko.observable(date.toLocaleDateString() + ' ' + date.toLocaleTimeString());
+										cdsi.startTime = ko.observable(momentApi.formatDateTime(date));
```
pavgra: how is it going to work if you removed (renamed) the method?

js/modules/WebAPIProvider/MomentAPI.js
```diff
@@ -22,7 +22,7 @@ define(function (require, exports) {
     return m.isValid() ? m.format(DURATION_FORMAT) : EMPTY_DATE;
   }
 
-  function formatDateTime(timestamp, outFormat) {
+  function formatDateTimeWithFormat(timestamp, outFormat) {
```
pavgra: `formatBlahBlahFormat` sounds not so good.  what was the reason for change?

PR #881, opened 2018-08-30
Css bundling

johnSamilin:
This PR introduces bundling of the css files


chrisknoll:
Hi, I have some concerns with these changes, but I'm currently OOO and will not be able to dig into this:

The main issue is how changes were made to some of the CSS (such as jquery) to point to an absolute path for the images instead of using a path relative to the CSS file.  This breaks the idea of being able to reuse components outside of the atlas context (if we want to use those components in other application contexts).  I believe these changes are necessary due to how the bundling is handled, but I think the CSS assets can be copied relative to the bundle location, and then the relative path will work for either case that it's loaded locally or via CDN.

chrisknoll:
Could you please resolve the conflicts? Thank you!

johnSamilin:
@chrisknoll resolved

build/index.js
```diff
@@ -24,38 +41,90 @@ function downloadLib(name, path) {
...
-    requirejs.optimize(settings, () => console.log('success'), (er) => console.error(er));
+  return {
+    libs,
+    cdnLibs,
+    paths,
+  };
+}
+
+function optimizeJs() {
+  const { libs, cdnLibs, paths } = downloadLibs(settings.paths);
```
pavgra: The function is called `optimizeJs` but not `downloadLibsAndOptimizeJs`. Let's keep things atomic and split the functionality.

build/index.js
```diff
@@ -24,38 +41,90 @@ function downloadLib(name, path) {
...
+          },
+        );
+      });
   });
+
+  return promise;
+}
+
+function optimizeCss() {
+  const { libs, cdnLibs, paths } = downloadLibs(settings.cssPaths, false);
```
pavgra: Same comment as above

js/settings.js
```diff
@@ -70,30 +70,38 @@ const settingsObject = {
       'jqueryui/autoGrowInput': 'assets/jqueryui/autoGrowInput',
     }
   },
-  paths: {
+  paths: {    
+    "datatables.net-buttons": "assets/jquery.dataTables.buttons.min",
```
pavgra: Why was the reordering done? Seems for me like only unnecessary diff

js/styles/jquery.dataTables.min.css
```diff
@@ -1 +1 @@
...
...Tables_processing,.dataTables_wrapper .dataTables_paginate{color:#333;}.dataTables_wrapper .dataTables_scroll{clear:both}.dataTables_wrapper .dataTables_scroll div.dataTables_scrollBody{*margin-top:-1px;-webkit-overflow-scrolling:touch}.dataTables_wrapper .dataTables_scroll div.dataTables_scrollBody th>div.dataTables_sizing,.dataTables_wrapper .dataTables_scroll div.dataTables_scrollBody td>div.dataTables_sizing{height:0;overflow:hidden;margin:0 !important;padding:0 !important}.dataTables_wrapper.no-footer .dataTables_scrollBody{border-bottom:1px solid #111}.dataTables_wrapper.no-footer div.dataTables_scrollHead table,.dataTables_wrapper.no-footer div.dataTables_scrollBody table{border-bottom:none}.dataTables_wrapper:after{visibility:hidden;display:block;content:"";clear:both;height:0}@media screen and (max-width: 767px){.dataTables_wrapper .dataTables_info,.dataTables_wrapper .dataTables_paginate{float:none;text-align:center}.dataTables_wrapper .dataTables_paginate{margin-top:0.5em}}@media screen and (max-width: 640px){.dataTables_wrapper .dataTables_length,.dataTables_wrapper .dataTables_filter{float:none;text-align:center}.dataTables_wrapper .dataTables_filter{margin-top:0.5em}}
```
pavgra: What and why was changed?

js/settings.js
johnSamilin: the order here is important

pavgra: Why? If there exist some rules for ordering, why weren't they described?

johnSamilin: It's because of the cross-dependencies. X requires Y, Y requires jquery.
To be honest, this exact order has been found by guessing

johnSamilin:
seems like you were right: `settings.js` diff was not necessary

PR #888, opened 2018-09-05
Structure refactoring/concept sets

johnSamilin:
part of #694


js/pages/concept-sets/components/tabs/conceptset-expression.js
```diff
@@ -0,0 +1,115 @@
...
+      } else {
+        this.toggleOffSelectAllCheckbox('.conceptSetTable', '#selectAllDescendants');
+      }
+      if (mappedCount == this.selectedConcepts().length) {
+        this.toggleOnSelectAllCheckbox('.conceptSetTable', '#selectAllMapped');
+      } else {
+        this.toggleOffSelectAllCheckbox('.conceptSetTable', '#selectAllMapped');
+      }
+      // Create event handlers for all of the select all elements
+      $(document).off('click', '#selectAllExclude');
```
pavgra: let's migrate such things to bindings

js/pages/concept-sets/components/tabs/conceptset-expression.js
```diff
@@ -0,0 +1,115 @@
...
+        }
+        if (conceptSetItem.includeDescendants()) {
+          descendantCount++;
+        }
+        if (conceptSetItem.includeMapped()) {
+          mappedCount++;
+        }
+      });
+      if (excludeCount == this.selectedConcepts().length) {
+        this.toggleOnSelectAllCheckbox('.conceptSetTable', '#selectAllExclude');
```
pavgra: let's not use IDs and try to migrate such things to bindings

js/pages/concept-sets/conceptset-manager.html
```diff
@@ -0,0 +1,190 @@
...
+    </div>
+  </div>
+  <tabs params="
+    componentParams: $component.componentParams,
+    selectTab: selectTab,
+    selectedTab: selectedTab,
+    tabs: $component.tabs
+  "></tabs>
+</div>
+<!--
```
pavgra: Why was this left?

js/pages/concept-sets/conceptset-manager.js
```diff
@@ -0,0 +1,465 @@
...
+			this.saveConceptSet("#txtOptimizerSavingNewName", conceptSet, selectedConcepts);
+			this.optimizerSavingNew(false);
+		}
+
+		cancelSaveNewOptimizedConceptSet() {
+			this.optimizerSavingNew(false);
+		}
+
+	}
+		/*
```
pavgra: Why was this left?

js/pages/concept-sets/components/tabs/conceptset-compare.html
```diff
@@ -0,0 +1,133 @@
...
+    },
+  }"
+></atlas-modal>
+
+<conceptset-save params="
+  conceptSetName: compareNewConceptSetName,
+  onSave: compareCreateNewConceptSet,
+  show: saveConceptSetShow"
+></conceptset-save>
+<!--
```
pavgra: Why was this left?

js/pages/concept-sets/components/tabs/conceptset-compare.js
```diff
@@ -0,0 +1,360 @@
...
+            var valid = true; //d.INVALID_REASON_CAPTION == 'Invalid' ? 'invalid' : '';
+            return '<a class=' + valid + ' href=\'#/concept/' + d.conceptId + '\'>' + d.conceptName + '</a>';
+          },
+        },
+        {
+          title: 'Class',
+          data: d => d.conceptClassId,
+        },
+        {
+          title: '<i id="dtConeptManagerRC" class="fa fa-database" aria-hidden="true"></i> RC',
```
pavgra: Move to template

js/pages/concept-sets/components/tabs/included-conceptsets.js
```diff
@@ -0,0 +1,129 @@
...
+      this.model = params.model;
+      this.ancestorsModalIsShown = ko.observable(false);
+      this.ancestors = ko.observableArray([]);		
+      this.showAncestorsModal = conceptSetService.getAncestorsModalHandler({
+        model: params.model,
+        ancestors: this.ancestors,
+        ancestorsModalIsShown: this.ancestorsModalIsShown,
+      });
+      this.searchConceptsColumns = [{
+        title: '<i class="fa fa-shopping-cart"></i>',
```
pavgra: Move to template

anthonysena:
@pavgra @johnSamilin  A few bugs I've noted for the Concept Sets section:

1. **Included concepts** tab and **Included Source Codes** show the loading icon and also an empty table while the results are pulled from WebAPI:

image

2. Table headings are missing in the **included concepts** tab:

image

3. There was an error on the Explore Evidence page related to the way the conceptSetService.getGenerationInfo() returns its information now. Before that was an array but it appears that instead it has been modified to include a .json property that contains the contents? This seems like a bug. By the way, do you guys have CEM set up so that you could test this?
4. The "Compare" tab fails to function; the POST to perform the comparison between 2 concepts fails and I'd guess it has something to do with the way the concept set is being sent to the server.

From the cohort definition side, here are some bugs and steps to reproduce:

Do the following: Copy to Concept Set Repository: create a cohort definition, navigate to the concept sets tab. Click "new concept set" and give it a name in the section that opens below. Next, navigate to search and add in some concepts (I used metformin for testing). Navigate back to the concept set definition in the cohort definition editor. Click 'descendants' on the concepts you selected. 

Bugs:

- After the steps above, I noticed that the count next to the 'Included Concepts' tab was not consistent with the contents of the table revealed under that tab. If I clicked through to the other tabs in that section and came back to 'Included Concepts' it was then showing a list that was consistent with the number in the tab. Unsure whether this is new based on the changes in this PR or if that's an on-going issue.
- I attempted to use the 'Copy to Concept Set Repository' button and received an error on the PUT request to WebAPI. Perhaps this should be a POST since we're creating a new entity from a copy of the current concept set?

Stopping here for now - please let me know if you have any questions on the items above.

chrisknoll:
I see that a change was made to pass a template ID for use in the facet data table, but is it possible to leverage the functionality described in the section 'Passing markup into components' here:
https://knockoutjs.com/documentation/component-custom-elements.html

pavgra:
@johnSamilin, I'm still receiving exception while doing the concepts comparison:
image

pavgra:
@johnSamilin ,

concepts-issue

anthonysena:
We're pretty close on this. From my testing, I noted a few other bugs:

- This isn't a bug specific to changes in this branch but it is a bug that we should fix here. On the concept set -> Export page, I noticed that the JSON that is under the 'Concept Set Expression JSON' heading is incorrect. It should resemble something like this:

```
{
	"items": [{
			"concept": {
				"CONCEPT_ID": 315558,
				"CONCEPT_NAME": "Atherosclerosis of arteries of the extremities",
				"STANDARD_CONCEPT": "S",
				"STANDARD_CONCEPT_CAPTION": "Standard",
				"INVALID_REASON": "V",
				"INVALID_REASON_CAPTION": "Valid",
				"CONCEPT_CODE": "51274000",
				"DOMAIN_ID": "Condition",
				"VOCABULARY_ID": "SNOMED",
				"CONCEPT_CLASS_ID": "Clinical Finding"
			},
			"isExcluded": false,
			"includeDescendants": true,
			"includeMapped": false
		}
	]
}
```
Note the properties:
```
"isExcluded": false,
"includeDescendants": true,
"includeMapped": false
```
are missing in the current JSON representation that is in ATLAS 2.5.1.

- The 'Included Source Codes' tab on the 'Concept Set' page is not working right. When you navigate to that tab, an empty table is shown...then after a few seconds the "loading" icon appears and eventually the results are displayed. The loading icon should come up immediately when clicking on that tab. 
- I cannot navigate directly to a concept again - if you go to: http://localhost/atlas/#/concept/46368009 using the code in this branch the app fails to load.

I would also ask that you please test the concept set import via the cohort definition editor - I've been attempting to use that as part of this testing and it appears to work but wanted to make sure this was part of your testing routing too.

If you need clarification on these items, please let me know.

johnSamilin:
Hi Anthony,
the second point is WebAPI-related.
as for the third one, Model's `loadIncluded` method is being triggered right after ConceptSetService loads info. The call chain is the following: route change -> `Model.loadRepositoryConceptSet` -> `ConceptSetService.loadConceptSet` -> `Model.loadIncluded`. So, I believe, this delay is somehow critical.
The latter issue is fixed

anthonysena:
Thanks @johnSamilin - I'll review and test those changes.  A few questions:

> the second point is WebAPI-related.

In what way? Was there a change on WebAPI that changed the way that concept sets are retrieved to show in the export section? I'm not sure why that would change as the concept set structure has been stable for quite a while.

> The call chain is the following: route change -> Model.loadRepositoryConceptSet -> ConceptSetService.loadConceptSet -> Model.loadIncluded. So, I believe, this delay is somehow critical.

The delay is understood - those calls can take time to complete. That said, we should then have variables that indicate when each section is loading so that we can have the loading display ready if/when a user clicks on that. In ATLAS 2.5.0, this works well so I'd suggest taking a look at that code for reference.

Also, one more question: when debugging on this branch, I noticed a number of warnings related to the use of depreciated components/modules. Should we clean those up? They are a bit distracting when debugging.

johnSamilin:
@anthonysena 
indeed, the JSON issue was related to front-end, I've fixed it
also I fixed other points from the list

anthonysena:
Thanks @johnSamilin - in reviewing this branch today. I'll list them here but let's decide how best to proceed forward:

- When viewing concept sets in a cohort definition, the 'source codes' tab does not appear to work - the table is always empty.
- Also in the context of a cohort definition concept set, I'm now unable to import a concept set from the repository. Getting the following errors:

image

- When optimizing a concept set, the modal dialog background does not disappear. To reproduce: create a new concept set with redundant concepts (e.g. metformin RxNorm Ingredient and from the Included Concepts tab, add a few concepts to the concept set). Click the 'Optimize' button and then use the 'Save Options' -> 'Create New Concept Set'. The operation completes (a new concept set is created) but the background doesn't disappear requiring you to refresh the browser.

I think if we can address the first 2 items above, we can merge this into master and address the 3rd item as a separate patch since we can work around that for now. Tagging @pavgra to get his thoughts on this plan.

pavgra:
@anthonysena , thank you for helping us not to commit a bunch of new bugs into the master! I fully support your plan.

johnSamilin:
@anthonysena not sure whether it's frontend-related. I Imported this cohort definition: 
[test-cohort.txt](https://github.com/OHDSI/Atlas/files/2423900/test-cohort.txt)
and got sourcecodes table filled:
capture123
I also fixed an import issue, the process was wierd

anthonysena:
I fixed one error related to selecting a concept set inside of the cohort definition manager which I've pushed to this branch. Without this patch, I was unable to open a concept set inside of a cohort definition.

I'm still facing an issues. For example: open a cohort definition, go to the concept sets, select a concept set to open it. At the bottom of the screen, select 'delete'. I was then faced with the following error:

image

I'd kindly ask that you do additional testing on the concept set portion of the cohort definition editor as this is a critical component of these changes. To be clear, please make sure that you have the developer console open and no errors appear when doing the following inside of a cohort definition:

- Editing an existing concept set: open a concept set, navigate to the vocabulary, add concepts to it and then navigate back to the concept set editor to ensure the concepts are added.
- Remove concepts from a concept set: use the 'shopping cart' icon to remove items and save the cohort definition
- Review the included source codes and concepts and ensure at least 1 item shows in the list. In most cases there will be > 1 but if there are 0, there is an issue.
- Import a concept set from the repository and overwrite an existing concept set. I was also seeing DataTable related errors on this operation.
- Copy a concept set from a cohort definition into the repository. This appeared to work at first but eventually threw an error in the console.
- Close the concept set
- Open another concept set. Delete it and save the cohort definition.

Thanks!

johnSamilin:
@anthonysena in my turn, I'd kindly ask that you do additional testing too, because the fix that you introduced (https://github.com/OHDSI/Atlas/pull/888/commits/97672c83a344fb6a21615eb1ff6c83e19151648f) led to adding empty ConceptSets (with id=undefined) to the cohort definition

johnSamilin:
This change is because inputTypes/ConceptSet doesn't expect the input parameters to be observables.
I think this branch is ready, though I have no idea where it will break next

anthonysena:
@johnSamilin  did another test starting at the vocabulary search. I attempted to go Search -> Import -> Concept Identifiers and then paste in this list:

`199074,433753,257007,443800,436665,4084966,29735,314658,4288310,380094,255573,443617,192367,134718,433811,438134,24609,437784,139099,80004,440358,316084,436940,319843,440374,193739,375292,321596,436676,194997,73754,141932,374366,313459,197236`

Got this error:

image

Traced it a bit to find that the js/pages/vocabulary/components/import.js expects the model passed into the component so that it can get the current concept set but the component never receives the model. Since this is using a new approach with the AutoBind(component) approach, I stopped here since I'm not sure how this parameter is expected to arrive inside of the component.

I'd like to help address some of these issues but must admit I'm still learning some of the new patterns introduced in the code. Please let me know how this would be addressed for my own education. Thanks!

pavgra:
@johnSamilin , not sure whether it is still actual, but you can get a look at the conceptSets import handling here
https://github.com/OHDSI/Atlas/blob/feature_analyses/js/pages/characterizations/components/feature-analyses/feature-analysis-view-edit.js#L197

johnSamilin:
@pavgra thanks, but this issue is not even related to concept-sets refactoring
@anthonysena fixed and checked all the ways to do import

anthonysena:
Thanks @johnSamilin - I'll test this out today and let you all know my findings.

anthonysena:
Nearly everything worked - 1 bug remains:

- Open a cohort definition -> Concept Sets Tab -> Open a concept set and note the ID of the concept set. Click "import" and select a repository concept set. The concept set is imported but it shares the same ID the the open concept set tab which will result in issues with the cohort generation process. The IDs will need to be unique and I think the approach that @pavgra shared is applicable here.

Thanks!

PR #891, opened 2018-09-06
refactored jobs page

johnSamilin:
part of #694


js/pages/jobs/job-manager.js
```diff
@@ -0,0 +1,75 @@
...
+			this.canReadJobs = ko.pureComputed(() => {
+				return authApi.isPermittedReadJobs();
+			});
+
+			if (this.canReadJobs()) {
+				this.updateJobs();
+			}
+		}
+
+		updateJobs() {
```
pavgra: why not in service?

PR #897, opened 2018-09-10
For issues 896 => fix authApi not defined error after authentication.

chen-regen:
Hi Anthony @anthonysena, please review. This is for issue 896. Thank you!

### DIFF
--- a/js/pages/home/routes.js
+++ b/js/pages/home/routes.js
@@ -1,6 +1,7 @@
 define(
   (require, factory) => {
     const { Route } = require('providers/Route');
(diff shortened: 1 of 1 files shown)


js/pages/home/routes.js
```diff
@@ -1,6 +1,7 @@
 define(
   (require, factory) => {
     const { Route } = require('providers/Route');
+    var authApi = require('webapi/AuthAPI');
```
pavgra: Pls, use `const` instead of `var`

chen-regen: Done.

PR #900, opened 2018-09-12
created t&c modal window

### LINKED ISSUE #899: License agreement for SNOMED
Since Atlas provides access to SNOMED vocabulary, it is required for users to accept terms & usage conditions of SNOMED. So we need to introduce a popup similar to the one at [Snomed browser](http://browser.ihtsdotools.org) and require a user to click "Accept" before he / she can proceed to work with Atlas.

To make things more general and customizable proposed design is the following:
- On startup Atlas checks for local terms & conditions config file (which contains `header`, `description`, `content`, `acceptanceExpiresInDays`)
- If the file is presented, Atlas looks up localStorage to check whether the user has accepted the terms & conditions before and whether the acceptance is not expired
- If the user hasn't accepted the terms & conditions before or the acceptance has expired, Atlas shows the popup with content based on the config file
- If the user accepts the terms & conditions, Atlas saves it to localStorage and doesn't show the popup again until the acceptance is expired; otherwise Atlas closes the current tab.

Default config file should cover the [following popup](http://browser.ihtsdotools.org/?):
image

tagging @cgreich, @gklebanov

johnSamilin:
fixes #899


js/config/local.config.js
```diff
@@ -0,0 +1,9 @@
+define([], function () {
+	var config = {};
+	config.userAuthenticationEnabled = true;
```
pavgra: This was disabled by default. Same as props below. Why was the behavior changed?

js/components/terms-and-conditions/terms-and-conditions.js
```diff
@@ -0,0 +1,63 @@
...
+        }
+      });
+      this.title = termsAndConditionsConfig.header;
+      this.description = termsAndConditionsConfig.description;
+      this.content = termsAndConditionsConfig.content;
+      this.isAccepted = ko.observable(true);
+      
+      setTimeout(() => {
+        this.isAccepted(this.checkAcceptance());
+      }, 300);
```
pavgra: Looks very dirty. Not a solution

js/config/app.config.js
```diff
@@ -1,7 +1,7 @@
-define(['optional!config-local'], function (localConfig) {
+define(['optional!config/local.config'], function (localConfig) {
```
pavgra: Since this is not a major release, pls add fallback for old `config-local.js` file

PR #901, opened 2018-09-12
fixed flushing of the component params

johnSamilin:
fixed flushing component params when navigating between pages

### DIFF
--- a/js/providers/Model.js
+++ b/js/providers/Model.js
@@ -352,6 +352,13 @@ define(
 			*/
 			}
 
+			handleViewChange(view, componentParams = {}) {
+				if (componentParams !== false) {
+					this.componentParams(componentParams);
+				}
+				this.currentView(view);
+			}
+
 			// for the current selected concepts:
 			// update the export panel
 			// resolve the included concepts and update the include concept set identifier list
--- a/js/providers/Router.js
+++ b/js/providers/Router.js
@@ -24,7 +24,6 @@ define(
       run() {
         const routerOptions = {
           notfound: () => this.handleNotFound(),
-          on: () => this.getModel().componentParams({}),
         };
         this.router = new Router(this.aggregateRoutes());
         this.router.qs = this.qs;
@@ -49,7 +48,8 @@ define(
         }, {});
         const routesWithRefreshedToken = Object.keys(routes).reduce((accumulator, key) => {
 					accumulator[key] = (...args) => {
-            this.setCurrentView('loading');
+            // TODO:
+            //this.setCurrentView('loading');
 						if (this.onLoginSubscription) {
 							this.onLoginSubscription.dispose();
             }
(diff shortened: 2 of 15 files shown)


pavgra:
@johnSamilin, spread operator used for `<div id="currentComponent" class="flexed" data-bind='component: {name: $data.currentView, params: { ...$data.componentParams(), model: $data } }'></div>` caused creation of new `params` object each time when some route params changed, so page compoennt was re-created. I've introduced fix for that by using dedicated observable. Pls go through current router params usages and updated object paths.

js/pages/estimation/cohort-comparison-manager.js
```diff
@@ -48,10 +48,10 @@ define([
 	) {
 		class CohortComparisonManager extends AutoBind(Component) {
 			constructor(params) {
-				super();
-				this.cohortComparisonId = params.currentCohortComparisonId;
-				this.cohortComparison = params.currentCohortComparison;
-				this.cohortComparisonDirtyFlag = params.dirtyFlag;
+				super(params);
```
chrisknoll: Please see comment in https://github.com/OHDSI/Atlas/issues/910 for how we should handle routerParams as a direct observable, and not extract each param from the routerParams() object.

js/providers/Model.js
```diff
@@ -352,6 +354,13 @@ define(
 			*/
 			}
 
+			handleViewChange(view, routerParams = {}) {
+				if (routerParams !== false) {
```
chrisknoll: I don't understand the routerParams !== false here.  do you need to pass 'false' as the routerParams value to get that block to execute?  Instead should we just default routerParams to null and if we don't have a value, then don't set the routerParams observable?

js/providers/Router.js
```diff
@@ -49,7 +48,7 @@ define(
         }, {});
         const routesWithRefreshedToken = Object.keys(routes).reduce((accumulator, key) => {
 					accumulator[key] = (...args) => {
-            this.setCurrentView('loading');
+            this.getModel().loading(true);
```
chrisknoll: I can't quite tell fro code, but is this setting a global 'loading' observable that will set a top-level 'loading' message?  I seem to recall that when the 'top' level loading is presented, the currentView component is removed from the dom. By removing it from the dom, it destroyes and re-creates the component needlessly.  If this is true, I think each component should be responsible for manging their own 'lodading' messaging and this shouldn't be managed at the route-processing level (we shoudln't see a flash of 'loading' show up when we go too cc/analysis/detail or cc/analysis/executions.

pavgra: @chrisknoll, this is setting loading icon in navbar instead of top-level loading panel to remove the re-initiation of page-component:
image

And yes, components should take care of loading panel on their own.

js/providers/Model.js
johnSamilin: That's true, route params should be flushed between top-level page navigations unless you implicitly say the router not to do so (pass false as the second param).

PR #912, opened 2018-09-18
Page route status fix

### LINKED ISSUE #910: Dirty indicators no longer show up
### Expected behavior
When navigating to the IR, PLP, or PLE nav in Atlas, if you do not have an existing item open, you should get the list of items that are available for that module (the list of IR analysis, for example).  

When you select something to open, it should then change the left nav icon to a color: green for 'open' and red for 'open and unsaved'.  If you select the left nav with an entity open, it should navigate directly to the edit screen for what the user has selected to edit.   If the entity is 'dirty', the icon in the left nav should appear 'red'.

### Actual behavior

When selecting an IR, PLP, or PLE analysis, the indicator in the left nav no no longer indicates that the item is 'dirty' (a green icon means not dirty, a red icon means unsaved changes). Additionally, after opening an item, clicking back on the nav is supposed to send you back to the item that is open, but as of the 'pages refactoring' the left nav is static, and clicking on the left nav always send you back to the list view of the module.

Because of the lost functionality, if you open an IR analysis, and make unsaved changes, then click on the left nav, you are directed back to the IR Analysis list (browse) you can click on some other IR Analysis, and there's no inidication that unsaved changes were lost, that you had an IR Analysis open, etc.

Go to IR analysis List.
Click New Analysis. 
Note: Icon in left nav is not green
Make changes, do not save
click Left nav.  Note: it takes you to the analysis list (this is wrong, it should have not navigated anywhere because you have an analysis open
Click some other IR analysis item.  Your unsaved changes are lost, and no notification.

### Implementation notes;

From the 2.4.3 branch, you can see the left nav is created statically and looks like this:
https://github.com/OHDSI/Atlas/blob/v2.4.3/index.html#L47

```
			<a class="list-group-item" data-bind="css: irStatusCss, attr: {href: irAnalysisURL}"><i class="fa fa-fw fa-bolt" aria-hidden="true"></i>&nbsp;Incidence Rates</a>
			<a class="list-group-item" href="#/profiles"><i class="fa fa-fw fa-user" aria-hidden="true"></i>&nbsp;Profiles</a>
			<a class="list-group-item" data-bind="css: ccaCss, attr: {href: ccaURL}"><i class="fa fa-fw fa-balance-scale" aria-hidden="true"></i>&nbsp;Estimation</a>
			<a class="list-group-item" data-bind="css: plpCss, attr: {href: plpURL}"><i class="fa fa-fw fa-heartbeat" aria-hidden="true"></i>&nbsp;Prediction</a>
```

The css and attr bindings are used to control what the user is navigated to when the left nav is clicked. You should also notice: not all links work this way.

### Background on the difference in behavior:
The Concept Set and Cohort Definition modules work differently than the other modules:  Because of the way the vocabulary manipulates concept sets in the 'open' entity, if you have a cocnept set open or cohort definition open, you have a 'breadcrumb' at the top of the main UI that will let you return to the open cohort def/concept set.  Note, this meant that you could only have only either a cohort defintiion OR a concept set open to manage, but not both.  Complicated logic was implemented in the app to make this functionality work.  You will see things like the 'source' of the concept set being 'definition' or 'repository' and other things like this.

When we added other modules (IR was the first one), we quickly realized that the breadcrumb behavior was not maintainable and very complicated to implement.  In addition, it did not make sense to have to close your IR analysis just because you opened up a concept set.  Therefore, we coded the app to maintain a state of the open entitities, and had the left nav show colors based on the 'dirtyness' of the entity and the URL to navigate to the module would change to the 'active' entity so if you click on it, you go back to the item you have open.

When the code was refactored into the pages structure, and the left nav is now dynamic based on what page modules have been defined, we lost the functionality of maintaining dirty state of the entity and also the navigation to the open entity in the left nav.

johnSamilin:
fixes #910


chrisknoll:
The incident rate status indicator works: goes green when open, red when unsaved, and green when saevd.  Navigating between home and IR page keeps the unsaved data in memory.

However:
image

This error happens when I navigate to 'profiles' page from Incidence Rate view.  Soemthing about routerParams was null so you can't subscribe to it.

johnSamilin:
@chrisknoll fixed by merging master branch in

chrisknoll:
Still a problem:
image

To reproduce:  
1. Open IR analysis.
2. click on 'search' left nav
3. Observe error in console.

chrisknoll:
Ok, we'll merge this in.  @johnSamilin , can you let us know the other branch that has the additional fixes so we can get those fixes incorporated into master?

PR #918, opened 2018-09-19
fixed expired session handler

### LINKED ISSUE #884: Unauthorized request doesn't break the session
### Expected behavior
When API answers with 401 status code, page-wide empty state should appear
### Actual behavior
When API answers with 401 status code, there's only an error in console
* Start atlas in secure mode
* Go to Estimation page
* Open PLE analysis
* Wait for the session to expire
* Change tab

johnSamilin:
fixes #884


pavgra:
@johnSamilin , this doesn't address the issue with receiving 401 as a result of some AJAX happening at a page (e.g. polling), not when you navigate between some pages:

image

PR #923, opened 2018-09-19
fixed exception on charts (bad data)

### LINKED ISSUE #914: Units, Values, Concepts charts shouldn't be empty in Measurement report of Achilles
### Expected behavior
Units, Values, Concepts charts shouldn't be empty in Measurement report of Achilles

### Actual behavior
The charts are empty
image

1. Go to Data sources
1. Select a source
1. Open Measurement report
1. Click on a concept

johnSamilin:
closes #914 
also need improvements into @ohdsi/atlascharts package

### DIFF
--- a/js/modules/databindings/chart.js
+++ b/js/modules/databindings/chart.js
@@ -20,7 +20,12 @@ define(['knockout'], (ko) => {
   ko.bindingHandlers.chart = {
 		update: function (element, valueAccessor) {
       const chart = valueAccessor();
-			draw(chart.data(), element, chart.minHeight, chart.format, chart.renderer);
+      try {
+        draw(chart.data(), element, chart.minHeight, chart.format, chart.renderer);
+      } catch(er) {
+        console.error('Error when rendering chart');
+        draw(null, element, chart.minHeight, chart.format, chart.renderer);
+      }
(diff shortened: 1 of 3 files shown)


js/modules/databindings/chart.js
```diff
@@ -20,7 +20,12 @@ define(['knockout'], (ko) => {
   ko.bindingHandlers.chart = {
 		update: function (element, valueAccessor) {
       const chart = valueAccessor();
-			draw(chart.data(), element, chart.minHeight, chart.format, chart.renderer);
+      try {
+        draw(chart.data(), element, chart.minHeight, chart.format, chart.renderer);
+      } catch(er) {
+        console.error('Error when rendering chart');
```
pavgra: Why wouldn't you add exception object to logger?

johnSamilin: because first, it tells nothing (it doesn't know anything about chart's name), and second, the actual bug is in the `atlascharts` (it doesn't check data for validity)

chrisknoll:
@johnSamilin : if the atlascharts throws an exception, tho, would it be different?  We can throw exceptions in the charting library, we just haven't yet.

Issue #938, opened 2018-09-25
Age of First Observation chart shows the shifted data


mpozhidaeva:
### Expected behavior
The chart shows correct data with correct axis
### Actual behavior
Age of First Observation chart shows the shifted data (in case with negative "age")
### Steps to reproduce behavior
Generate and open the dashboard report

Original issue: https://odysseusdataservices.atlassian.net/browse/ATLP-525

mpozhidaeva:
Some examples of charts changes:
1)**Before all changes** (master branches of WebAPI and Atlas, atlascharts - 1.5.0-alpha-0, not unique mapHistogram() function in different .js files)

Cohort report: observation period
![cohort-observ-1](https://user-images.githubusercontent.com/31308645/46421216-5a312180-c73a-11e8-9bee-3cba9102f75c.png)

![cohort-observ-2](https://user-images.githubusercontent.com/31308645/46421334-93699180-c73a-11e8-9f33-91f07ed40133.png)

Cohort report: person
![cohort-person](https://user-images.githubusercontent.com/31308645/46421355-9c5a6300-c73a-11e8-9122-1fcab49bc99e.png)

Data source report: dashboard

Test input data:

|  intervalIndex | percentValue  |  countValue | 
|---|---|---|
| 4  |0.0018599| 10272|
| 3  |0.00203| 9397|
| 1  |0.00252|  12761|
| 0  |0.012509|  63266|
| -2 |1.799999| 901|
| 2  |0.00197| 9975|
| -1 |0.0002| 989|
| 5  |0.0002| 400|

![ds-dashboard](https://user-images.githubusercontent.com/31308645/46421390-abd9ac00-c73a-11e8-9ceb-fc339a736b6c.png)

2)**The same code/data + attempt to use mapHistogram() from atlascharts**

![ds-dasboard](https://user-images.githubusercontent.com/31308645/46422162-674f1000-c73c-11e8-8a19-9dbacf369cc6.png)

2)**After changes** (atlascharts - 1.5.0-beta-0)
Cohort report: observations

![cohort-obs-1](https://user-images.githubusercontent.com/31308645/46422267-a54c3400-c73c-11e8-921f-11ea2aed3cb8.png)

![cohort-obs-2](https://user-images.githubusercontent.com/31308645/46422286-abdaab80-c73c-11e8-9b51-cc6dec1e3efa.png)

Cohort report: person

[...]

PR #941, opened 2018-09-26
Fixed age of observation chart with negative ages

mpozhidaeva:
Fixes https://github.com/OHDSI/Atlas/issues/938


chrisknoll:
Hi @mpozhidaeva ,
 Please see the implementation of mapHistogram here:
```
https://github.com/OHDSI/Visualizations/blob/12f327dcf6acd439d60c4c3827444d7895570e92/atlascharts/histogram.js#L26
```

Can you check if this implementation will work for you. The version in AtlasCharts should have the correct form.

PR #942, opened 2018-09-26
fixed bundling

johnSamilin:


chrisknoll:
Did you ever figure out why that d3-chromatic file was not downloading by the build script?

johnSamilin:
Because jsdelivr answers with status code 301

chrisknoll:
Ok, in the future could the bundling process handle those responses? I hate having the repo filled wit copies of CDN'd libs.

PR #948, opened 2018-09-28
Finalized list and view-edit pages for Feature Analyses

pavgra:
For https://github.com/OHDSI/WebAPI/issues/522. To be merged after https://github.com/OHDSI/WebAPI/pull/608 is finalized.


chrisknoll:
Question: in IR, and TxP, we have indicators showing that there is an open 'project' in the nav, and clicking on the left nav will redirect you to the edit page for the project/design.  I don't think I see that behavior implemented here (ie: there shoudl be an indicator on the nav, and it should be green if 'open' and red if 'dirty'.

One thing about CC, however, is that you can have _two_ potential things open: a feature design or a characterization design.  This puts a new twist on the modality of 'you have a design open'...but, if the characerization page, with the 2 tabs, can decide which component to show for each sub-tab: if you have an open CC, show the CC designer (view-edit). If you have a feature design open, show the feature design component.  The nav indicator will just be red if you have either a cc or feature design unsaved, green if you have either a cc or feature open, else, none.

PR #953, opened 2018-10-02
Estimation & Prediction Specification Editors

anthonysena:
ATLAS code for the new editors - must be tested with the est-plp-symposium branch of WebAPI under review at OHDSI/WebAPI#620. Aims to address OHDSI/WebAPI#568.

This is not ready for merging - I will still make commits to this branch but feel free to start any reviews.


js/pages/estimation/cca-manager.js
```diff
@@ -0,0 +1,580 @@
...
+			this.editorComponentParams({ 
+				comparison: comparison,
+			});
+			this.managerMode('editor')
+		}
+
+		deleteFromTable(list, obj, index) {
+			// Check if the button or inner element were clicked
+			if (
+				obj.target.className.indexOf("btn-remove") >= 0 ||
```
johnSamilin: split into two separate methods

js/pages/estimation/cca-manager.js
```diff
@@ -0,0 +1,580 @@
...
+			}
+		}
+
+		newAnalysis() {
+			this.loading(true);
+			this.estimationAnalysis(new EstimationAnalysis({id: 0, name: 'New Population Level Estimation Analysis'}, this.estimationType, this.defaultCovariateSettings));
+			// EstimationAnalysis takes time to load - use the setTimeout({}, 0) 
+			// to allow the event loop to catch up.
+			// http://stackoverflow.com/questions/779379/why-is-settimeoutfn-0-sometimes-useful
+			setTimeout(() => {
```
johnSamilin: It's a hack. Better use promise

js/pages/estimation/cca-manager.less
```diff
@@ -0,0 +1,156 @@
+.panel-default>.panel-heading {
```
johnSamilin: we added less loader not for just a cool extension.
first of all, less means that you follow it's semantics of folding the rules
then, the Component class provides a field `classes`, which is a BEM class names factory, which, in conjunction with the LESS styles should provide easy and safe way to style a component. It's styles should always be local to the component

js/pages/estimation/components/CohortMethodAnalysisEditor.html
```diff
@@ -0,0 +1,318 @@
+<div class="paddedWrapper">
+    <div>
+        <div class="row">
+            <div class="analysisNameContainer"><label class="analysisName">Analysis name:</label></div>
+            <div><input class="form-control analysisName" style="width: 500px;" type="text" data-bind="textInput: $component.analysis.description" /></div>
```
johnSamilin: why 500px?

js/pages/estimation/components/CohortMethodAnalysisEditor.html
```diff
@@ -0,0 +1,318 @@
...
+            </div>
+            <div class="panel-body">
+                <div>
+                    <outcome-model-args-editor params="matchStratifySelection: $component.matchStratifySelection, outcomeModelArgs: $component.analysis.fitOutcomeModelArgs"></outcome-model-args-editor>
+                </div>
+            </div>
+        </div>    
+    </div>
+</div>
+<div data-bind="modal: showConceptSetSelector" class="modal fade" tabindex="-1" role="dialog">
```
johnSamilin: please use `atlas-modal` component

js/pages/estimation/components/CohortMethodAnalysisEditor.js
```diff
@@ -0,0 +1,126 @@
+define([
+	'knockout', 
+	'text!./CohortMethodAnalysisEditor.html',	
+	'providers/Component',
+	'utils/CommonUtils',
+	'../options',
+	'../inputTypes/ConceptSet',
+	'databindings',
```
johnSamilin: what does this import do here?

js/pages/estimation/estimation-browser.js
```diff
@@ -0,0 +1,120 @@
...
+						return momentApi.formatDateTimeUTC(d.modifiedDate);
+					}
+				},
+				{
+					title: 'Author',
+					data: 'createdBy'
+				}
+			];
+			
+			EstimationService.getEstimationList()
```
johnSamilin: move to `onPageCreated` lifecycle method

js/pages/estimation/inputTypes/ComparativeCohortAnalysis/CreateStudyPopulationArgs.js
```diff
@@ -0,0 +1,25 @@
+define(function (require, exports) {
+
+    var ko = require('knockout');
+
+	function CreateStudyPopulationArgs(data) {
```
johnSamilin: why not create a class for each of the input types to get rid of `self` pattern?

js/pages/estimation/inputTypes/NegativeControl.js
```diff
@@ -0,0 +1,11 @@
+define(function (require, exports) {
+
+    var ko = require('knockout');
+
+	function NegativeControl(data) {
```
johnSamilin: please specify a default value instead of modifying an input parameter

johnSamilin: btw, what is the purpose of this function?

js/pages/estimation/options.js
```diff
@@ -0,0 +1,295 @@
+define(
+    (require, exports) => {
+  
+    var options = {};
```
johnSamilin: why not `const.js`?

js/pages/prediction/components/modelSettings/LassoLogisticRegressionSettings.js
```diff
@@ -0,0 +1,24 @@
...
+			super(params);
+
+			this.variance = {
+				name: 'variance',
+				value: this.modelSettings.variance,
+			};
+		}
+	}
+
+	return commonUtils.build('LassoLogisticRegressionSettings', LassoLogisticRegressionSettings, view);
```
johnSamilin: why not kebab-case?

js/pages/prediction/const.js
```diff
@@ -2,8 +2,27 @@ define(
   (require, exports) => {
     const pageTitle = 'Prediction';
 
+    const apiPaths = {
```
johnSamilin: this should be called `paths`

js/pages/estimation/cca-manager.js
```diff
@@ -0,0 +1,580 @@
...
+					}
+					if (xref.propertyName === constants.conceptSetCrossReference.positiveControlCovariateSettings.propertyName.excludedCovariateConcepts) {
+						targetPcsCovariateSettings.excludedCovariateConceptSet(selectedConceptSet);
+					}
+				}
+			});
+		}
+
+		
+		init() {
```
johnSamilin: each component which is being used as a top-level view (in other words, goes to `router.setCurrentView`), should extend `Page` class. This class has a lifesycle method called `onPageCreated`

js/pages/estimation/cca-manager.less
chrisknoll: Without proper guidance on how to apply BEM principles and how the LESS loader implements them, it's extremely difficult to just look at a piece of code and understand what the proper way to apply those principles are.  What we were given was 'here's the BEMHelper...here's a .less loader,...now go', which is grossly insufficient on how to apply those tools properly.  

having just started using BEMHelper (via $component.classes('xxx')), I can tell you that it's very challenging to move from the 'global style' approach of applying classes to the 'define each individual element', so what's presented in this PR is what we could get to work following the patterns that are better understood.

I think it would be useful to dedicate a Atlas/WebAPI meeting to walk through BEM, the BEMHelper and LESS syntax, and this is something that should have been done as part of introducing the less loader and BEM helper. So, please hold your scorn when you don't see something is being applied when there's zero documentation or training on how to apply those methods properly.

js/pages/estimation/inputTypes/ComparativeCohortAnalysis/CreateStudyPopulationArgs.js
chrisknoll: In this case, the self reference isn't necessary because there are no functions/behavior being defined in this class (it's all state data attached to the instance).

I'm not sure a class is even necessary here because, per javascript implementation of classes, classes just manipulate the prototype chain.  I think returning a constructor function isn't a terrible practice since it is clear that without the class directive, there's no behavior being 'templates' to the instance. 

If the idea behind ES6 classes is: 'use them always, don't event think about it', then we should make that clear, but this code is written to not write something 'just for the sake of writing something'.

js/pages/estimation/cca-manager.less
pavgra: @chrisknoll , even w/o touching BEM, the presence of `.panel-default>.panel-heading` in LESS file looks like the person, who coded it, totally misses the preprocessor idea

js/pages/estimation/components/ComparisonEditor.html
```diff
@@ -0,0 +1,95 @@
...
+    <div>
+        <label>Choose your negative control outcomes:</label>
+        <div class="input-group">
+            <input class="form-control" type="text" disabled data-bind="value: comparison.negativeControlOutcomesConceptSet().name" />
+            <span class="input-group-btn">
+                <button class="btn btn-primary" data-bind="click:chooseNegativeControlOutcomesConceptSet"><i class="fa fa-folder-open"></i></button>
+                <button class="btn btn-danger" data-bind="disable:comparison.negativeControlOutcomesConceptSet().id == 0,click:clearNegativeControlOutcomesConceptSet" type="button"><i class="fa fa-times-circle" aria-hidden="true"></i></button>
+            </span>
+        </div>
+        <!--<div data-bind="visible: comparison.comparator().id > 0 && comparison.comparator().name == null" class="errorInline"><i class="fa fa-exclamation-circle" aria-hidden="true"></i> The comparator cohort specified (#<span data-bind="text:comparison.comparator().id"></span>) was deleted. Please select a different cohort.</div>-->
```
pavgra: Why commented code was left?

js/pages/estimation/components/PositiveControlSythesisSettingsEditor.html
```diff
@@ -0,0 +1,121 @@
...
+                    <span><i class="fa fa-cogs"></i> Prior</span>
+                </div>
+                <div class="panel-body collapse">
+                    <cyclops-prior-editor params="prior: $component.settings().prior"></cyclops-prior-editor>
+                </div>
+            </div>
+        </div>
+    </div>
+</div>
+<div data-bind="modal: showCovariateSelector" class="modal fade" tabindex="-100" role="dialog">
```
pavgra: There is a separate component for modals

js/pages/estimation/components/StratifyArgsEditor.js
```diff
@@ -0,0 +1,29 @@
...
+	options,
+) {
+	class StratifyArgsEditor extends Component {
+		constructor(params) {
+            super(params);
+
+			this.stratifyArgs = params.stratifyArgs;
+            this.options = options;
+
+            this.hasCovariateIds = ko.pureComputed(() => {
```
pavgra: Issues w/ indentation all over the place

js/pages/estimation/estimation-browser.js
```diff
@@ -0,0 +1,120 @@
...
+				},
+				{
+					title: 'Type',
+                    data: d => d.type,
+                    visible: false,
+				},
+				{
+					title: 'Name',
+					data: d => {
+						return '<span class=\'linkish\'>' + d.name + '</span>';
```
pavgra: Why don't you use regular, normal links? What is more, there is a link builder in DatatableUtils for this

js/pages/estimation/options.js
```diff
@@ -0,0 +1,317 @@
...
+
+    options.positiveControlSynthesisArgs = {
+        modelType: [{
+            name: "Poisson",
+            id: 'poisson',
+        }, {
+            name: "Survival",
+            id: 'survival'
+        }],
+        minOutcomeCountForModelOptions: ['100', '75', '50', '25', '10'],
```
pavgra: Use lodash

js/pages/prediction/components/PopulationSettingsEditor.html
```diff
@@ -0,0 +1,68 @@
...
+<div>
+    <label>Add exposure days to start?</label>
+    <div class="input-group">
+        <select class="form-control" data-bind="options: $component.options.yesNoOptions, optionsText:'name', optionsValue:'id', value: $component.populationSettings.addExposureDaysToStart"></select>
+    </div>
+</div>
+<div>
+    <label>Define the time-at-risk window end:</label>
+    <div class="input-group">
+        <span contenteditable="true" class="numericInputField dropdown" data-bind="htmlValue: $component.populationSettings.riskWindowEnd, eventType:'blur', ko_autocomplete: { value: $component.populationSettings.riskWindowEnd, source: $component.options.dayOptions, minLength: 0, maxShowItems: 10, scroll: true }" />&nbsp;days from cohort start date
```
pavgra: Usage of `span` for what is input in fact, is a bit...hm..not very logical as for me and spoils sematics. Pls get a look how it is properly done in Cohort Pathways

js/pages/prediction/prediction-browser.js
```diff
@@ -0,0 +1,120 @@
...
+			});
+			this.canCreatePrediction = ko.pureComputed(() => {
+				return (config.userAuthenticationEnabled && authApi.isAuthenticated() && authApi.isPermittedCreatePlps()) || !config.userAuthenticationEnabled;
+			});
+
+			this.options = {
+				Facets: [
+					{
+						'caption': 'Last Modified',
+						'binding': function (o) {
```
pavgra: Doesn't copy-pasting this block of code all over the place tell that it should be moved into a shared function? Anyway, that was done, pls use `DatatableUtils.getFacetForDate`

js/pages/prediction/prediction-browser.js
```diff
@@ -0,0 +1,120 @@
...
+				},
+				{
+					title: 'Type',
+                    data: d => d.type,
+                    visible: false,
+				},
+				{
+					title: 'Name',
+					data: d => {
+						return '<span class=\'linkish\'>' + d.name + '</span>';
```
pavgra: Why linkish??

js/pages/prediction/prediction-manager.js
```diff
@@ -0,0 +1,619 @@
...
+					obj.target.className.indexOf("btn-remove") >= 0 ||
+					obj.target.className.indexOf("fa-times") >= 0
+				) {
+					this.deleteFromTable(this.modelSettings, obj, rowIndex);
+				} else {
+					this.editModelSettings(data);
+				}		
+			}
+	
+			this.covariateSettingRowClickHandler = (data, obj, tableRow, rowIndex) => {
```
pavgra: Why handlers are not at class level?

js/pages/prediction/components/PopulationSettingsEditor.html
chrisknoll: The reason for the us of span is that it can grow/shrink based on the content of the DOM.  Perhaps there has been recent advances in CSS and DOM technology, but this is why the people on the HTML specification committee added the 'contenteditable' attribute to the element.

And as far as 'proper', please please do not use inflammatory language when suggesting an alternative.  The Pathways implementation desired the static with in the input fields because of their arrangement.  If we did it 'properly' in cohort definition editor here:
image

vs if we went with pathways:
image

Which  feel the compressed input fields makes sense here for cohort definitions. Does it make sense in PLE? Why not ask the question instead of passing judgement?

js/pages/prediction/prediction-browser.js
chrisknoll: 'all over the place' is a bit of hyperbole.  This utility class was just merged into master 2 weeks ago, and, I don't know about you, when I copy in some code, I don't go ahead and do a full codebase search to see if there is a utility class that performs the work.  I think we all need to come to an understanding that when something is added to the codebase, it isn't automatically encoded into the brains of everyone developing code (especially when much of this code is done in parallel.  

I think the underlying issue here is that we don't have an effective mechanism for introducing new efficiencies in the codebase in a way that people are aware of when and how to use it.  This is the case with BEMHelper + Less. This is the case with the Page life cycle.  This is the case for the new utility class functions.  

We have the Atlas/WebAPI calls to bring awareness of these things to the community. We've been having them for several weeks now, and no one brought up these items to raise awareness.  So, for the next few WebAPI calls, let's make sure that we raise these items for awareness...bring up the code, show how they are used in practice.

pavgra: A bit, yes. But:
- the construction was definitely copy-pasted from some other place; so even at this point it would be logical to move it into an util method and re-use
- it was copied at least twice in the PR (hard to see precise - too many not expanded files); the second time where I would personally think of moving copy-pasted stuff into a shared place

> I think the underlying issue here is that we don't have an effective mechanism for introducing new efficiencies

That's true, no doubts. Neither Page, nor BEMHelper are easy to get w/o intro that's why you haven't seen me mentioning them in the review. But LESS, being around for a long while, I thought is a bit more straightforward thing.

chrisknoll: Great.  absolutely agree that if it looks like we're copying things, we should think about encapsulation and re-use. That is a fair assessment.

But let's agree that it's fair to call out areas of where we could be leveraging technology/frameworks where they aren't use in practice, and that if we aren't using them, it's a symptom that it needs to be addressed at the Atlas/WebAPI calls and instruction on it's use.

Let's also agree that until that communication occurs, no one is 'on the hook' for doing it 'correctly'.  That's why it would be unfair to criticize your team for not following existing patterns in the code base (which we have never gave you formal introduction to the code) while at the same time it _is_ unfair to criticize design choices have been made without understanding the context and reasoning behind those decisions.  I'm sure we have a lot to learn from each other from our respective experiences.  So let's enter into these debates in the spirit of mutual understanding.

pavgra: > Let's also agree that until that communication occurs, no one is 'on the hook' for doing it 'correctly'

I thought we've discussed UI refactoring all together and received a lot of valid feedback from you, guys, same as you reviewed our PRs so that I assumed the communication took place (which is not the same thing with WebAPI, where we haven't done such broad communication yet). Anyway, this comment was about at least 3 blocks of the exactly same code blocks, so isn't it valid? And I just pointed that currently the function already exists so no need to code it from the ground.

chrisknoll: > Anyway, this comment was about at least 3 blocks of the exactly same code blocks, so isn't it valid? 

Yes, it's valid statement, which is where I started with:

> Great. absolutely agree that if it looks like we're copying things, we should think about encapsulation and re-use. That is a fair assessment.

Where it got on to the 'communication' topic was that we should have known that there was some util function that we should all be using for that column. That led into the other thread, but I think we're all on the same page now.

js/pages/prediction/components/PopulationSettingsEditor.html
pavgra: Why I named it not proper? Because I can fall into such situation:
2018-10-05 15_11_42-_atlas_ cohort definitions unsaved

Why do I receive this? Because `contenteditable` is rich-text editor, which is not what required for a single digit. Your screenshot is a bit unfair: although you are right about fixed width, there is no need to do such wide fields. So smth like 5-digit wide should look OK.

chrisknoll: >  Your screenshot is a bit unfair:

It's just an input box, what your screenshot is showing is that the user is free to type in whatever they want into it. What your screen shot is not showing is that the numeric validation rejects that as an input and everything snaps back down to size.  Unfair-says-what?

As far as my example being unfair,  took the exact dimensions from the TXP pathways UI to capture the _exact_ same range of numeric inputs for those parameters.  I don't know how i could have been more fair.

We're getting off track here, the commentary was just how your feedback is being delivered:  making statements about what is 'proper' without understanding what the functionality requirements are.  So, I suggest we answer the question 'What is the desired behavior here' first, next ask 'Is the approach X applicable for this situation', and then no one feels like their efforts are degraded.

chrisknoll:
image

In the above, the choice is equipoise, and looks like a 25,75 input value, but the bounds label is (1-100%)

chrisknoll:
Hello, @pavgra  and @johnSamilin ,
 I appreciate your feedback on these items, and although I had every intent on addressing the concerns, It's now 1am on Sunday, and I realize that it is not going to happen.

Here's the status:

I spent much of the time looking through the .less concerns (thinking it was brought up first, so I'll tackle that first), and one aspect I wasn't clear on was how separate components may want to reuse definitions from a single .less file.  In this specific case: we have the PLP editor and PLE editor sharing a common component 'cyclops'. Within cyclops, there is 2 UI compoennts called a 'ControlEditor' and a 'PriorEditor'.  These guys have input fields (a numericInputField with a modifier of dropdown that adds a down arrow selector), and so, (we really don't have adequate documentation on how we're supposed to use less + BEM),   it looks like the apporoach was to use global styles and load the style sheet via 'less!../cyclops'.  And, if the styles are global, it works just like if you loaded it via some css loader.  So, I attempted to encapsulate a central BEM helper in the cyclops' 'options' module and I would like to get your feedback on that approach here: https://github.com/OHDSI/Atlas/tree/est-plp-symposium-refactoring

Maybe this would be a good post-symposium exercise on leveraging less and BEM usage in atlas.

PR #956, opened 2018-10-02
Cohort Pathways Implementation

chrisknoll:
In addition:
Fix execution scrolling problem (both CC and TxP)


chrisknoll:
I should also add: i modified the modal component to allow you to embend the template HTML directly inside the ```<modal>``` tag, so you do not need to pass in a template_id or set up a ```<script>``` tag anywhere with the modal content.

js/components/cohort/linked-cohort-list.js
```diff
@@ -62,6 +75,12 @@ define([
 				return `<a data-bind="click: () => $component.params.${action}('${d[identifierField]}')">Remove</a>`;
 			}
 		}
+		
+		getEditCell() {
```
pavgra: Why this is not used anywhere?

js/components/cohort/linked-cohort-list.less
```diff
@@ -21,4 +21,14 @@
 			padding-left: 1rem;
 		}
 	}
+	
+	&__table.edit span {
```
pavgra: Why not using modificator to be compliant with BEM?

js/main.js
```diff
@@ -86,11 +87,13 @@ require([
 			'components/terms-and-conditions/terms-and-conditions',
 		],
 			(
+				ko,
 				Application,
 				Model,
 				Router,
 				sharedState,
 			) => {
+				//ko.options.deferUpdates = true;
```
pavgra: Do we need this?

js/pages/pathways/components/browser.js
```diff
@@ -0,0 +1,109 @@
...
+	constants,
+	config,
+	PathwayService,
+	authApi,
+	Component,
+	commonUtils,
+	datatableUtils
+) {
+	class PathwaysBrowser extends Component {
+    constructor(params) {
```
pavgra: indentation issue

js/pages/pathways/components/browser.js
```diff
@@ -0,0 +1,109 @@
...
+      super(params);      
+      this.loading = ko.observable(false);
+      this.config = config;
+      this.analysisList = ko.observableArray();
+      this.isAuthenticated = authApi.isAuthenticated;
+      this.canRead = ko.pureComputed(() => {
+        return (config.userAuthenticationEnabled && this.isAuthenticated() && authApi.isPermittedReadIRs()) || !config.userAuthenticationEnabled;
```
pavgra: `this.isAuthenticated()` is redundant - this check is encapsulated in `AuthAPI.isPermitted`

js/pages/pathways/components/browser.js
```diff
@@ -0,0 +1,109 @@
...
+        return (config.userAuthenticationEnabled && this.isAuthenticated() && authApi.isPermittedCreateIR()) || !config.userAuthenticationEnabled;
+      });
+      
+      // startup actions
+      this.refresh();
+    }
+
+		refresh() {
+			this.loading(true);
+			PathwayService
```
pavgra: let's use async / await

js/pages/pathways/components/browser.js
```diff
@@ -0,0 +1,109 @@
...
+				.list()
+				.then(res => {
+					this.analysisList(res.content);
+					this.loading(false);
+				});
+		}
+
+		newAnalysis() {
+			commonUtils.routeTo('/pathways/0/design');
```
pavgra: let's move link building to util method (since it appears in multiple places)

js/pages/pathways/components/manager.js
```diff
@@ -0,0 +1,165 @@
...
+			if (analysisId !== undefined) {
+				this.analysisId(parseInt(analysisId));
+				this.load(this.analysisId() || 0);
+			}
+			this.setupSection(section);
+			this.executionId(subId);
+		}
+
+		selectTab(index, { key }) {
+			commonUtils.routeTo(`/pathways/${this.componentParams.analysisId()}/${key}`);
```
pavgra: The same url builder would be useful here

js/pages/pathways/components/manager.js
```diff
@@ -0,0 +1,165 @@
...
+
+		load(id) {
+			if(this.dirtyFlag().isDirty() && !confirm("Your changes are not saved. Would you like to continue?"))
+				return;
+			
+			if (id < 1) {
+				this.setupDesign(new PathwayAnalysis());
+			} else {
+				this.loading(true);
+				PathwayService
```
pavgra: async / await here and below

js/pages/pathways/components/tabs/pathway-executions.js
```diff
@@ -0,0 +1,216 @@
+define([
+	'knockout',
+	'../../PathwayService',
```
pavgra: I would use absolute path here to exclude issues if file is moved

js/pages/pathways/components/tabs/pathway-results.less
```diff
@@ -0,0 +1,168 @@
...
+		.title {
+			font-size: 2.0rem;
+		}
+		.sub-title {
+			font-size: 1.2rem;
+		}
+	}
+	
+	&__detail-table {
+		tr td:nth-child(2), tr td:nth-child(3), tr th:nth-child(2), tr th:nth-child(3), {
```
pavgra: Cannot we apply classes here?

pavgra:
Permissions check is incorrect. Should use module's PermissionService

image

pavgra:
Analysis params section was not styled as we agreed to
image

pavgra:
Design links are not clickable / permissions are not properly checked
image

pavgra:
Layout for sunburst viewer blocks should take 100% width and three blocks should take the whole space (left and right ones, I believe, should take equal space). Also, they should be consistent in term of borders. I would suggest applying panels for the three blocks (similarly to how charts in Achilles / Heracles reports are laid out). What is more, things in the right side (details viewer) are too closely located - need to add more "air" there (paddings / margins), same as some styling for the table (basic bootstrap classes should work)

image

chrisknoll:
> Analysis params section was not styled as we agreed to
I seem to have misplaced our specification on how that should be styled. Could you send it to me?  But note: some of those controls (like the numeric dropdown selector) are reused components from other parts of the app (where numeric fields are input), so be aware that these inputs are consistent with other inputs on the app.

> Design links are not clickable / permissions are not properly checked

I'm going to need some help with this, I've used the implementation from characterization almost exactly (except for the permission methods coming from the permission service for pathways).  If you have a secured environment with permissions set up, it would be helpful if you could troubleshoot exactly what is leading to the links being active.

pavgra:
So, @chrisknoll , you want to work on the styling, right? Will you be able to address the code related comments?

chrisknoll:
From the results report perspective, this is what the latest push yields:
image

all 3 elements are arranged along a center-line which makes them not stetch top-to-bottom and is easier to eye-track from left to rght, so I think this is an improvement.  The eyesore is how the title of the plot is not exactly centered under the sunburst.  the div's space is centerd, but the content isn't aligned the way i'd like.  I tried messing with varous CSS controls on it, but I coudln't quite get there.

js/pages/pathways/PathwayAnalysis.js
```diff
@@ -0,0 +1,34 @@
+define(function (require, exports) {
```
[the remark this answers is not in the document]
chrisknoll: No, I'm not in favor of that.  There's just 2 or 3 classes invovled (a data class, 2 service classes) and there's no need to divide those into separate pacakges.

We're seeing this attitude take over on the WebAPI side, and I'm going to strongly push against this highly granular separation of classes into 'pattern' packages, but this is a Post-symposium discussion item, so I won't bog down the PR here.

js/pages/pathways/components/tabs/pathway-design.less
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  &__col-param-remove {
+    width: 10%;
+  }
+
+  //
+
+	&__settings .numericInputField {
+		padding-left: 3px;
```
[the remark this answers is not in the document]
chrisknoll: Without them the UI element looks wrong.

js/pages/pathways/PathwayAnalysis.js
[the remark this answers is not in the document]
pavgra: For a single model class (`PathwayAnalysis`) I'm OK with leaving it in root, but for 2 services I'll create appropriate `services` folder

js/pages/pathways/components/tabs/pathway-design.less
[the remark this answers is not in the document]
pavgra: @johnSamilin is saying that typically very specific paddings (not just 5 / 10 / 15 px, etc) mean smth wrong with markup approach

pavgra:
@chrisknoll , I am working on permissions and styling issues now so let's be careful not to overlap.

Into the log of things which I cannot get why were not addressed from UX point of view and which I have to work on:
image
- how a user, just looking at the page, can understand that the name in second column is editable? I cannot imagine how...
- Action columns are not of equal size
- Do you treat "Edit" column size as reasonable?

pavgra:
Security was tested and necessary fixes were done. Redirect after creation of new analysis was fixed.

@chrisknoll / @pbr6cornell , hopefully the styling changes make sense:

**Before**. Pathway analysis design page.
design-before

**After**. Pathway analysis design page.
design-after

**Before**. Result page.
sunburst-before

**After**. Result page.
sunburst-after

pbr6cornell:
The styling does look nicer, thank you for that work Pavel.

pavgra:
Linking to https://github.com/OHDSI/WebAPI/issues/522

PR #980, opened 2018-10-05
Fix notifications popup issues

vantonov1:


pavgra:
@vantonov1 , 
1. Fix indentation - tabs should be used
1. Use `/services/http.js` instead of `$.ajax`
1. Why is `jobListing` still here if we talked about relying on back-end status? What is going to happen if I refresh page - all notifications gone?

vantonov1:
3 jobListing is the model for the notifications table, and where are subscribers to jobListing updates
On page refresh notifications are reloaded from backend

pavgra:
@vantonov1 , if back-end is the single source of truth - why do have lines like `sharedState.jobListing.queue(job);`?

4. Why commented out lines were not removed?
```
// this.pollForInfo();
// executionId: String(this.model.currentCohortDefinition().id()) + String(this.getSourceId(source.sourceKey)),
``

**Comment by vantonov1** [comment]
Not exactly. Backend is polled, it happens not often, so first notification (with PENDING state usually) is added on UI, and then updated from backend

4 Yes, it have to be removed

File: js/components/negative-controls.js [hunk 1]
```diff
@@ -425,10 +423,13 @@ define(['knockout',
...
 			$.when(negativeControlsJob)
-				.done(function (jobInfo) {
-					pollTimeout = setTimeout(function () {
-						sharedState.jobListing.queue(job);
-						self.pollForInfo();
+				.done(function (info) {
+                    job.name = info.data.jobParameters.jobName;
+                    job.executionId = info.data.executionId;
+                    job.status(info.data.status);
+                    sharedState.jobListing.queue(job);
```
pavgra: 1. The 4 lines of code above look exactly the same way as they do in `negative-controls.js`. Why not to move it to util method?
1. Why `queue` here is called outside of timeout function (which seems to be the proper way), while above - inside the timeout?

js/pages/incidence-rates/ir-manager.js
```diff
@@ -85,7 +85,7 @@ define([
 			this.selectedAnalysisId.subscribe((id) => {
 				authAPI.loadUserInfo();
 			});
-			
+
```
pavgra: Dummy diff

js/pages/incidence-rates/ir-manager.js
```diff
@@ -253,54 +253,55 @@ define([
...
-			this.queueJob(sourceItem);
-			var executePromise = IRAnalysisService.execute(this.selectedAnalysisId(), sourceItem.source.sourceKey);
-			executePromise.then(() => {
+            var jobDetails = this.createJobDetails(sourceItem);
+            var executePromise = IRAnalysisService.execute(this.selectedAnalysisId(), sourceItem.source.sourceKey);
+			executePromise.then((info) => {
+                jobDetails.name = info.data.jobParameters.jobName;
+                jobDetails.executionId = info.data.executionId;
+                jobDetails.status(info.data.status);
+                sharedState.jobListing.queue(jobDetails);
```
pavgra: Get a look at the comment above regarding shared util method

js/pages/incidence-rates/ir-manager.js
```diff
@@ -337,27 +338,31 @@ define([
...
-								this.pollForInfo();
-							});
+                            var jobDetails = this.createJobDetails(sourceItem);
+                            var executePromise = IRAnalysisService.execute(this.selectedAnalysisId(), sourceItem.source.sourceKey);
+                            executePromise.then((info) => {
+                                jobDetails.name = info.data.jobParameters.jobName;
+                                jobDetails.executionId = info.data.executionId;
+                                jobDetails.status(info.data.status);
+                                sharedState.jobListing.queue(jobDetails);
+                                this.pollForInfo();
```
pavgra: The same as above

pavgra:
> Not exactly. Backend is polled, it happens not often, so first notification (with PENDING state usually) is added on UI, and then updated from backend

@vantonov1 , wouldn't it be better just throw event / call a method to pull the list of notifications when required but not to maintain two lists (versions) of the notifications?

vantonov1:
it costs additional server call. After success call to start a job, server returns just enough info to update notifications list - we just add new job to the list, there is no 2 versions. I dont think it is worth to ignore returned info, and ask again

chrisknoll:
I'm in agreement with @vantonov1 : this is how it works (or used to work? Haven't checked if refactor changed this behavior) in cohort definition generation.  The reason:  if we rely on the polling to update the status of the generation, and the polling is at 10s intervals, if you post the generate request at second 1 of 10, you have to wait 9 whole seconds to see the UI respond to your generation request. What do users do when they don't see an immediate result of their button click? they click it again. How many clicks can you fit in 9 seconds?  A lot (you should see me play starcraft).  So, instead, the approach is to get the button click, disable the button, do the async post, get the response, read the status of the job or just populate the job list with a fake (but have the jobID from the initial POST to get the identifier of the job) and then subsequent polls will deliver the up-to-date state.

pavgra:
@chrisknoll , if you read my previous message, you'd see that no you do NOT

> have to wait 9 whole seconds

because I suggested

> call a method to pull the list of notifications when required

But, I am OK with this:
> I dont think it is worth to ignore returned info, and ask again

chrisknoll:
Apologies, @pavgra , I did not recognize that consideration. However, I will say that I've seen tricky issues with if you have a poll on a setInterval, and then due to a user action you do a invoke-and-wait call to the same action that the setInterval does, you sometimes get this out of order refresh where the setInterval comes back with some slightly different data than your user-initiated call returns and it gets all flaky.   So, my underlying concern was just have a single source of refresh (the polling) while being able to update the UI with some state that confirms to they user that we got the request and are working on it.  Certainly doing @vantonov1 approach means there is no network latency involved.

> But, I am OK with this:

Does that mean you are OK to approve the PR or did you want me to pull and review?

vantonov1:
Im going to make all suggested changes today

js/pages/cohort-definitions/cohort-definition-manager.js
```diff
@@ -1083,46 +1056,21 @@ define(['jquery', 'knockout', 'text!./cohort-definition-manager.html',
...
-						sharedState.jobListing.queue(jobDetails);
-					},
-						error: (xhr, status, error) => {
-							this.createReportJobFailed(true);
-							var createReportJobErrorPackage = {};
-							createReportJobErrorPackage.status = status;
-							createReportJobErrorPackage.error = xhr.responseText;
-							this.createReportJobError(JSON.stringify(createReportJobErrorPackage));
+				this.createReportJobFailed(false);
+				return httpService.doPost(config.api.url + 'cohortanalysis', JSON.stringify(cohortJob))
```
pavgra: Pls define `httpService.doPost(config.api.url + 'cohortanalysis', data)` as an appropriate service method

js/pages/cohort-definitions/cohort-definition-manager.js
```diff
@@ -774,29 +760,16 @@ define(['jquery', 'knockout', 'text!./cohort-definition-manager.html',
...
-					viewed: false,
-					url: 'cohortdefinition/' + this.model.currentCohortDefinition().id() + '/generation',
-				});
-
-				$.ajax(route, {
-					error: this.authApi.handleAccessDenied,
-						success:  (data) => {
-						job.status(data.status);
-						sharedState.jobListing.queue(job);
+				httpService.doGet(route)
```
pavgra: Pls define `httpService.doGet(route)` as an appropriate service method (e.g. `generateCohort(cohortId, sourceKey, includeFeatures)`

js/components/user-bar.js
```diff
@@ -6,120 +6,146 @@ define([
...
-							context: job,
-							success: (progressData) => {
-								if (job.progress() != progressData.progress) {
-									var currentStatus = job.getStatusFromResponse(progressData);
-									job.status(currentStatus);
-									job.viewed(false);
-									this.jobListing.valueHasMutated();
-								}
+		updateJobStatus(endpoint) {
+			httpService.doGet(endpoint)
```
pavgra: Pls define `httpService.doGet(endpoint)` as an appropriate service method (e.g. `JobService.list`)

js/components/user-bar.js
```diff
@@ -6,120 +6,146 @@ define([
...
+				var unviewedNotificationCount = this.jobListing().filter(j => {
+					return !j.viewed();
+				}).length;
+				return unviewedNotificationCount;
+			});
+
+			this.clearJobNotificationsPending = this.clearJobNotificationsPending.bind(this);
+
+			this.updateJobStatus = this.updateJobStatus.bind(this);
 			this.startPolling = () => {
```
pavgra: Move methods out of constructor (https://github.com/OHDSI/Atlas/issues/990) to class-level ones

js/components/user-bar.js
```diff
@@ -6,120 +6,146 @@ define([
...
+			this.jobListing = state.jobListing;
+
+			this.jobNotificationsPending = ko.computed(() => {
+				var unviewedNotificationCount = this.jobListing().filter(j => {
+					return !j.viewed();
+				}).length;
+				return unviewedNotificationCount;
+			});
+
+			this.clearJobNotificationsPending = this.clearJobNotificationsPending.bind(this);
```
pavgra: use `class UserBar extends AutoBind(Component)` to get rid off manual bindings

pavgra:
@vantonov1 , exception while trying to execute cohort

image

pavgra:
@vantonov1 , markup is broken during application load:

image

pavgra:
@vantonov1 , unnecessary paddings in Notifications popup:

image

pavgra:
@vantonov1 , markup is broken during application load:

image

pavgra:
@vantonov1 , unnecessary paddings in Notifications popup:

image

pavgra:
image

1. Links for PLP (at least) don't work
1. There is no caption that only Latest N are shown

pavgra:
@vantonov1,

1) Why most recent jobs are at the end of the list?
2) I've started Cohort Characterization - CC kicks-off cohorts generation before the analysis itself, but those cohorts are not standalone entities, there is no reason for showing them independently of CC (as a separate lines)

image

Also, it would make sense to add "Duration" and "End time" columns between "Job name" and "Status". Pls address

js/components/user-bar.js
```diff
@@ -56,71 +47,99 @@ define([
...
+		getJobURL(n) {
+			switch (n.jobInstance.name) {
+				case "generateCohort":
+					return 'cohortdefinition/' + n.jobParameters.cohort_definition_id + '/generation';
+				case 'irAnalysis':
+					return 'iranalysis/' + n.jobParameters.analysis_id + '/generation';
+				case 'negativeControlsAnalysisJob':
+					return 'conceptset/' + n.jobParameters.concept_set_id + '/evidence';
+				case 'generateCohortCharacterization':
+					return 'cc/characterizations' + n.jobParameters.cohort_characterization_id;
```
pavgra: `/` is missed - link doesn't work

js/components/user-bar.html
```diff
@@ -33,46 +35,45 @@
...
+                <tr>
+                    <td class="pad-5"><i class="fa fa-tasks"></i> <a data-bind="click: () => { $component.data().jobNameClick($data); }, text: $data.name" class="linkish"></a></td>
+                    <td class="pad-5" data-bind="html: status"></td>
+                </tr>
+                <!-- /ko -->
+                </tbody>
+            </table>
+        </div>
+    </div>
+    <div style="justify-content: flex-end; display: flex;">
```
pavgra: No inline styles

js/components/user-bar.js
```diff
@@ -1,126 +1,125 @@
...
 							}
-						});
-					}
+						} else {
+							job = {
+								type: n.jobInstance.name,
+								name: n.jobParameters.jobName,
+								status: ko.observable(n.status),
+								executionId: n.executionId,
+								viewed: ko.observable(false),
```
pavgra: This causes notification icons to be highlighted on each page load

pavgra:
@vantonov1 , 

1. Generate cohort
1. Wait until the cohort is generated and notifications are refreshed from server

image

pavgra:
@vantonov1 , there is following block of code in `cohort-definition-manager.js`:

```
// check if we can tell if the job to generate the reports is already running
					if (this.model.currentCohortDefinition()) {
					var listing = sharedState.jobListing;
						var tempJob = listing().find(this.isActiveJob)
					if (tempJob) {
						if (tempJob.status() == 'STARTED' || tempJob.status() == 'STARTING' || tempJob.status() == 'RUNNING') {
								this.currentJob(tempJob);
								this.generateReportsEnabled(false);
							return "generating_reports";
						}
					}
				}
```

As soon as we kick off 10 jobs after Cohort reporting (Heracles) it will stop working properly (the required job won't be in the shared state anymore -> need to query for the job status explicitly)

PR #982, opened 2018-10-05
Fixed save btn disabled state in concept set

aklochkova:


js/pages/configuration/configuration.js
```diff
@@ -106,21 +106,21 @@ define([
     updateVocabPriority() {
       var newVocabUrl = sharedState.vocabularyUrl();
       var selectedSource = sharedState.sources().find((item) => { return item.vocabularyUrl === newVocabUrl; });
-      this.updateSourceDaimonPriority(selectedSource.sourceKey, 'Vocabulary');
```
chrisknoll: Out of curiosity: why not ```this```?  I feel like I've been burned in the past when trying to invoke one method from another and without this it states that the function isn't found.  In this case, what error did the ```this``` reference cause?

aklochkova: This fix went into https://github.com/OHDSI/Atlas/pull/951/files

pavgra:
@aklochkova , so why your change would fix the issue? You just did double inversion. So, nothing changed, the same problem as before:
concept-set-double-save-issue

Issue #973, opened 2018-10-04
Data Source order on Cohort Generation tab


fdefalco:
### Expected behavior
Data sources listed in alphabetical order

### Actual behavior
Data sources listed out of order

### Steps to reproduce behavior
Go to generation tab on cohort definition

PR #991, opened 2018-10-07
Fixed source ordering

pavgra:
Fixes https://github.com/OHDSI/Atlas/issues/973


js/modules/WebAPIProvider/SourceAPI.js
```diff
@@ -183,6 +184,8 @@ define(function (require, exports) {
       }
     });
 
+    sourceList = lodash.sortBy(sourceList, ['sourceName']);
```
chrisknoll: use native sorting here.

Issue #1010, opened 2018-10-09
New PLP download doesn't work


pavgra:
Download button doesn't open new window (not related to security). Popup blocking notification is also not shown.

![plp-download-doesnt-work](https://user-images.githubusercontent.com/10010071/46693955-56d5e400-cbd9-11e8-8e8a-2956597fae3a.gif)

Issue #1009, opened 2018-10-09
New PLP / PLE update doesn't work


pavgra:
When you click Save on existing entity, Atlas sends a request for creating new thing but not updating the existing one

![plp-update-doesnt-work](https://user-images.githubusercontent.com/10010071/46693751-b8e21980-cbd8-11e8-9f84-a86de46e3f65.gif)

Issue #983, opened 2018-10-05
Popup blocker warning for PLP package download


AnthonyMolinaro:
### Expected behavior
Download the R package for your PLP study or see an error message if something went wrong.

### Actual behavior 
Using Chrome my popup blocker blocked the download and it wasn't obvious this was happening.

### Steps to reproduce behavior
1. Turn on popup blocker.
2. Create a PLP study.
3. Name the R package and click download.

chrisknoll:
Sites use adblocker detection to show a message on screen, maybe we can add a panel to the top/bottom of the atlas main container and if we detect a pop-up blocker is in place, we can show the panel indicating that the blocker should whitelist atlas?

pavgra:
@chrisknoll, blocked download popups are typically due to improperly implemented downloading mechanisms - when the browser thinks that smth is being downloaded w/o user's intent. So that's the other story relatively to adblockers

PR #1011, opened 2018-10-09
Fixes UI security methods. Fixes PLE / PLP update & download methods

pavgra:
Front-end fix for https://github.com/OHDSI/WebAPI/issues/642
Fixes https://github.com/OHDSI/Atlas/issues/1009
Fixes https://github.com/OHDSI/Atlas/issues/1010
Fixes https://github.com/OHDSI/Atlas/issues/983

### DIFF
--- a/js/pages/estimation/PermissionService.js
+++ b/js/pages/estimation/PermissionService.js
@@ -1,0 +1,40 @@
+define([
+	'webapi/AuthAPI',
+], function (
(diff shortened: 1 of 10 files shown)


chrisknoll:
Download behavior is nice.  popup-block proof. I'm sorry we didn't know about this feature earlier, but we had to resort to the download-via-popup because we didn't have an alternative.

Issue #1004, opened 2018-10-09
Reshuffle contents of "data", "providers" and "modules" folders according to the refactoring proposal


pavgra:
`To be refactored` section from https://github.com/OHDSI/Atlas/issues/990

Issue #990, opened 2018-10-06
Coding conventions


pavgra:
WORK IN PROGRESS

## Background

A lot of work was done on UI refactoring, many initials ides were rethought during the evaluation, a lot of new things were brought into and finally, I feel that we are ready to write down and agree on new coding conventions to make contribution requirements transparent and our codebase - consistent.

### General folders approach

/split-by-type/split-by-domain/split-by-type/...

### High-level project structure

**/components**

**/config**

**/data**

**/extensions**
  /bindings
  /ko-extenders
  /plugins

**/pages**

**/resources**
    /libs
    /styles
   /images
   /fonts

**/services**
  \+ appropriate data classes

**/utils**

#### To be refactored

/providers -> to be removed by putting stuff into context-specific locations
/modules/circe/components -> /components/circe
/modules/conceptpicker -> /components/conceptpicker
/modules/conceptsetbuilder -> /components/conceptsetbuilder 
/modules/databindings -> /extensions/bindings
/modules/job -> /services/job
/modules/plp -> /pages/plp/services
/modules/WebAPIPRovider -> /services

### Page structure

TBD

### Dependency management

NPM

### Build process

Webpack

pavgra:
@chrisknoll , @anthonysena , @fdefalco , @johnSamilin , @wivern , would be great to hear your feedback and get to the agreement

johnSamilin:
Totally agree with these proposals

chrisknoll:
Thanks for letting me take the time to review the codebase as it has evolved so I can take an objective view about what some of these patterns mean from a code structure perspective and what challenges they might introduce that we should consider when applying them.


## General Patterns

>Folder nesting should follow /split-by-type/split-by-domain/split-by-type/... approach

[...]

PR #1021, opened 2018-10-17
Project structure refactoring

pavgra:
Fixes https://github.com/OHDSI/Atlas/issues/1004


pavgra:
@chrisknoll , obviously those numbers are due to folders renaming. Regarding the `src`, the reason was the following: we have `resources` which contain `style` and `data`. Those are not JS files, so I was concerned with the paths like `js/resources/style` and thought that having `src/resources/style` would be more semantically correct

chrisknoll:
> @chrisknoll  , obviously those numbers are due to folders renaming. 

What's not obvious is why some of the files in this PR are indicating they are just renames (contributing 0 lines of change) while others appear as deletes.  Is it really obvious to you which of those deletes are a result of a folder rename and which are true deletes without looking through the 1180 files changed to gather which files were really gone and which files were not?  How can you tell?

Secondly, the semantics of where resources folder should go (and the introduction of the /src folder) is not in the scope of what your PR is addressing.  In our past experiences, a few things have slipped by under the fog and haze of thousands of lines of changes where the PR was indicating one scope of change, but buried deep in the massive amount of files changed was some other change. To give you a specific example:  atlas folder restructuring introduced a page life-cycle and component hierarchy which was unrelated to cleaning up the folder structure.

So, if you are willing to wait, we'll digest the entirety of this PR to understand what the actual scope of changes are in this PR.  Or, if you prefer speed, put together a PR that only addressed the elements in the ```To be refactored section``` from #990, which I presume will consist of just several file renames and a few dozen lines of code changes to update the moduleIDs that are called in the ```define()``` or ```require()``` calls.

pavgra:
@chrisknoll , ok, I will open several sequential PRs:
- `To be refactored` section from #990
- moving ko-extenders, requirejs plugins, etc into `extensions`
- assembling `extensions` folder

Will this work for you?

chrisknoll:
This is way better @pavgra, thank you for the consideration.

js/Model.js
```diff
@@ -7,17 +7,17 @@ define(
 		'services/CohortDefinition',
 		'services/ConceptSet',
 		'services/role',
-		'providers/AutoBind',
-		'providers/Vocabulary',
+		'utils/AutoBind',
+		'services/Vocabulary',
```
chrisknoll: I'm seeing the Vocabulary service as 'services/Vocabulary' but later (I will mark it below) i see a 'services/VocabularyProvider'.  Is this something that should be merged into a single VocabularyAPI or VocabularyService?

js/components/circe/components/CohortConceptSetBrowser.js
```diff
@@ -1,4 +1,4 @@
-define(['knockout', 'text!./CohortConceptSetBrowserTemplate.html', 'vocabularyprovider', 'appConfig', 'conceptsetbuilder/InputTypes/ConceptSet', 'webapi/AuthAPI', 'webapi/MomentAPI', 'components/ac-access-denied', 'databindings', 'css!./style.css'], function (ko, template, VocabularyProvider, appConfig, ConceptSet, authApi, momentApi) {
+define(['knockout', 'text!./CohortConceptSetBrowserTemplate.html', 'services/VocabularyProvider', 'appConfig', 'conceptsetbuilder/InputTypes/ConceptSet', 'services/AuthAPI', 'services/MomentAPI', 'components/ac-access-denied', 'databindings', 'css!./style.css'], function (ko, template, VocabularyProvider, appConfig, ConceptSet, authApi, momentApi) {
```
chrisknoll: Here I am seeing 'services/VocabularyProvider', vs 'services/vocabulary', and wonder if these things are redundant?

chrisknoll:
Overall, things look good and in order.  I wanted to know your thoughts about our naming conventions about those modules in the 'services' folder:  some end in 'API' some are just a domain name (like Vocabulary) others suffix 'Provider'... Do we want to apply those modules that are interfaces to some module library (ie: MomentAPI) vs. those modules which are javascript proxies back to the WebAPI services (ie: source service, vocabulary service, etc)?  It feels like there's no convention in place about how those modules should be named, and I thought this would be a good opportunity to address that.

pavgra:
@chrisknoll , 

1) I am not 100% for one option or another speaking of naming convention, but some specific postfix may simplify search by file type: so that you search by "Service"  or "API" and get list of all the services across the project. I would vote for "*Service"
1) Regarding MomentAPI / CookieAPI / etc, those have side-effects so I would rather treat them as services than utils
1) Yes, you are right - Vocabulary & VocabularyProvider should be merged, but I decided not to do it now to keep on track of just structure reshuffling and not receive too big diff again

Issue #1024, opened 2018-10-17
UI generates wrong link to cohort definition


pavgra:
### Steps to reproduce behavior

1. Go to Cohort Definitions, choose any Cohort.
2. Press button with the tip "Get a link to this Cohort Definition.
3. Check the link.

### Actual behavior
1. The link has double slash.

### Expected behavior
1. the link doesn't have double slash.

![image](https://user-images.githubusercontent.com/10010071/47100845-6e7b2100-d206-11e8-9710-e3a9808b3cae.png)

alondhe:
Seeing this still in public Atlas, are there plans to update that version to include @pavgra's bugfix?

PR #1025, opened 2018-10-17
Fixed incorrect link to cohort definition generated by UI

pavgra:
Fixes https://github.com/OHDSI/Atlas/issues/1024


js/pages/cohort-definitions/cohort-definition-manager.js
```diff
@@ -459,7 +459,7 @@ define(['jquery', 'knockout', 'text!./cohort-definition-manager.html',
 			
 			this.cohortDefinitionLink = ko.computed(() => {
 				if (this.model.currentCohortDefinition()) {
-					return this.config.api.url + "/cohortdefinition/" + this.model.currentCohortDefinition().id();
+					return this.config.api.url + "cohortdefinition/" + this.model.currentCohortDefinition().id();
```
wivern: may be add common-utils function that combines url from set of strings with checking whether each part ends with slash or not?
I believe this would be useful

PR #1031, opened 2018-10-17
Restrict deletion of sources with top-priority daimons

pavgra:
Part of https://github.com/OHDSI/WebAPI/issues/587

### DIFF
--- a/js/pages/configuration/const.js
+++ b/js/pages/configuration/const.js
@@ -1,9 +1,11 @@
 define(
   (require, exports) => {
     const pageTitle = 'Concept Sets';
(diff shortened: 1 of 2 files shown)


js/pages/configuration/const.js
```diff
@@ -1,9 +1,11 @@
 define(
   (require, exports) => {
     const pageTitle = 'Concept Sets';
+    const priotirizableDaimonTypes = ['Vocabulary', 'Results', 'CEM'];
```
wivern: CEM -> CDM?

PR #1034, opened 2018-10-18
Fixes for performance of cohort export to 2.4.5

vantonov1:


js/modules/cohortdefinitionviewer/main.js
```diff
@@ -1,6 +1,7 @@
 define(function (require, exports) {
 	
-	var ko = require('knockout')
+	var ko = require('knockout');
+	const commonUtils = require('utils/CommonUtils');
```
pavgra: How this could ever work if the utils file hasn't existed in 2.4.5?

PR #1069, opened 2018-10-27
closes #878 - Left-hand menu bar resolution

### LINKED ISSUE #878: Left-hand menu bar resolution
### Expected behavior
The menu bar on the far left will adjust with computer screen's resolution

### Actual behavior
Menu bar adjusts, but not fully enough to reach all items

image

View ATLAS on low-resolution screen

fdefalco:


pavgra:
As for me, the existing menu is much more convenient and nice looking. @chrisknoll / @anthonysena , your opinion?

chrisknoll:
I haven't pulled this PR to check, do you want to provide a screenshot and point out the issues?

fdefalco:
> As for me, the existing menu is much more convenient and nice looking. @chrisknoll / @anthonysena , your opinion?

The issue is that the size of the padding in the existing menu (10px top 10px bottom) makes it that some of the items in the left hand menu are inaccessible on some screens.  The screenshot on the referenced issue shows the start of the problem (configuration and feedback become inaccessible). The addition of the new side menu options (characterizations and cohort pathways) push an additional two menu items off the screen. 

https://github.com/OHDSI/Atlas/issues/878

pavgra:
So, @fdefalco , if the issue occurs only at smaller screens, why don't you apply the fix only to those screens but push it for everyone? Why not to use `@media`?

fdefalco:
That's a good suggestion.  It appears that when the refactor was performed, the change to the menu item padding was done for everyone.  I'll assign this to the original author and ask that it be done using @media as you suggest.  The intent is to allow all menu items to remain visible.

chrisknoll:
I think @fdefalco misunderstood my question: I was asking if @pavgra could provide a screenshot from this PR so that I can understand what he didn't like about the proposed solution. He said:

> As for me, the existing menu is much more convenient and nice looking. @chrisknoll / @anthonysena , your opinion?

And I wanted to understand what he considered 'more convenient and nice looking'.

pavgra:
@chrisknoll ,

Existing:
image

vs

Proposed:
image

pavgra:
@chrisknoll , addressed by https://github.com/OHDSI/Atlas/pull/1087

chrisknoll:
So we should look at #1087 instead?

chrisknoll:
Fixed in #1087

PR #1071, opened 2018-10-29
Implements distribution criteria analyses for Cohort Characterization

wivern:
Implements stage 1 from https://github.com/OHDSI/WebAPI/issues/495#issuecomment-412667979

### DIFF
--- a/js/components/cohortbuilder/WindowedCriteria.js
+++ b/js/components/cohortbuilder/WindowedCriteria.js
@@ -15,7 +15,7 @@ define([
 			this.StartWindow = new Window(data.StartWindow);
 			this.EndWindow = ko.observable(data.EndWindow && new Window(data.EndWindow));
 			this.RestrictVisit = ko.observable(data.RestrictVisit || false);
-			this.IgnoreObservationPeriod = ko.observable(data.IgnoreObservationPeriod || true);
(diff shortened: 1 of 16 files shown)


wivern:
I`ve fixed the "ignore observation period" checkbox, but could not reproduce the issue with columns ordering.

js/components/cohortbuilder/WindowedCriteria.js
```diff
@@ -15,7 +15,7 @@ define([
 			this.StartWindow = new Window(data.StartWindow);
 			this.EndWindow = ko.observable(data.EndWindow && new Window(data.EndWindow));
 			this.RestrictVisit = ko.observable(data.RestrictVisit || false);
-			this.IgnoreObservationPeriod = ko.observable(data.IgnoreObservationPeriod || false);
+			this.IgnoreObservationPeriod = ko.observable(data.IgnoreObservationPeriod || true);
```
chrisknoll: This can't default to true.  It must default to false.  The backwards compatabilty is that we do not ignore observation period. the reason why we made this 'ignore observation period' is that if the value is absent in the JSON, it defaults to false.   For CC, if you want the behavior to default to ignore observation period, then you need to do the new WindowedCriteria() and then update the ignoreObservationPeriod(true).

Issue #1063, opened 2018-10-25
error while browsing vocabulary hierarchy


fdefalco:
### Actual behavior
![image](https://user-images.githubusercontent.com/6586384/47507770-d05c0c00-d840-11e8-8bc1-7795616c49df.png)

### Steps to reproduce behavior
browse to concept (https://epi.jnj.com/atlas/#/concept/45888235)
browse to related concept from related concepts tab
https://epi.jnj.com/atlas/#/concept/42742329
switch to hierarchy tab

jhardin29:
Hi @fdefalco I think this is what I was trying to convey in post https://github.com/OHDSI/Atlas/issues/1048.

aklochkova:
Note that Table Qualifiers in Source configuration matter to reproduce this bug. While it works correctly for `CDM: synpuf_50_test
Results: synpuf_50_test_results
Vocabulary: synpuf_50_test` after this PR https://github.com/OHDSI/Atlas/pull/1073 it fails for `CDM: synpuf_53_test
Vocabulary: synpuf_53_test
Results: synpuf_53_test_results`. So I redesigned code here https://github.com/OHDSI/Atlas/pull/1094. Now it includes all cases of Table Qualifiers.

PR #1073, opened 2018-10-29
Set 'loadingRelated' in proper places while browsing vocabulary hierarchy

aklochkova:
Fixes https://github.com/OHDSI/Atlas/issues/1063

### DIFF
--- a/js/pages/concept-sets/concept-manager.js
+++ b/js/pages/concept-sets/concept-manager.js
@@ -417,7 +417,7 @@ define([
 			}
 			this.model.currentConcept(data);
 			// load related concepts once the concept is loaded
-            this.loadingRelated(true);
+			this.loadingRelated(true);
 			this.metarchy = {
 				parents: ko.observableArray(),
 				children: ko.observableArray(),
--- a/js/pages/concept-sets/concept-manager.js
+++ b/js/pages/concept-sets/concept-manager.js
@@ -1,7 +1,7 @@
 define([
 	'knockout',
-	'text!./concept-manager.html',
-	'components/Component',
+	'text!./concept-manager.html', 
+	'pages/Page',
 	'utils/AutoBind',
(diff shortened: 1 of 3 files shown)


js/pages/concept-sets/concept-manager.js
```diff
@@ -398,6 +398,7 @@ define([
 		}
 
 		async loadConcept(conceptId) {
+			this.currentConceptArray = ko.observableArray();
```
pavgra: I would not reassign observable but rather use `removeAll` method

js/pages/concept-sets/concept-manager.js
```diff
@@ -12,14 +12,14 @@ define([
...
-	Component,
+	Page,
 	AutoBind,
 	vocabularyProvider,
 	commonUtils,
 	sharedState,
 	httpService,
 ) {
-	class ConceptManager extends AutoBind(Component) {
+	class ConceptManager extends AutoBind(Page) {
```
pavgra: Why haven't you created `async onPageCreated() {` and moved initial loading there then?

js/pages/concept-sets/concept-manager.js
```diff
@@ -28,12 +28,13 @@ define([
...
 			this.sourceCounts = ko.observableArray();
 			this.loadingSourceCounts = ko.observable(false);
-			this.loadingRelated = ko.observable(false);
+			this.loadingRelated = ko.observable(true);
 
 			this.currentConceptId = params.model.currentConceptId;
 
 			this.subscriptions.push(
 				this.currentConceptId.subscribe((value) => {
+					this.loadingRelated(true);
```
pavgra: Why have you moved the loading flag out of the appropriate function? Why haven't you put it as a first line in the `loadRecordCounts`?

Issue #1061, opened 2018-10-25
Round results to two decimal places


fdefalco:
### Expected behavior
Numbers should have a format which limits the number of digits displayed after the decimal.

### Actual behavior
![image](https://user-images.githubusercontent.com/6586384/47505200-b409a080-d83b-11e8-80ce-b3d0c9935b0f.png)

### Steps to reproduce behavior
Generate cohort characterization and review age feature.

PR #1079, opened 2018-10-30
Rounded 'avg' and 'stdDev' for cohort characterization

aklochkova:
Fixes https://github.com/OHDSI/Atlas/issues/1061


js/pages/characterizations/components/characterizations/characterization-view-edit/characterization-results.js
```diff
@@ -438,11 +442,11 @@ define([
                 });
                 columns.push({
                     title: 'Avg',
-                    render: (s, p, d) => d.avg[i],
+                    render: (s, p, d) =>  this.roundDecimal2(d.avg[i]),
```
pavgra: Pls use existing libs. Specifically, we have `numeral(val).format('0.00')`

PR #1094, opened 2018-11-06
 Moved checking for 'undefined' to the right place while browsing vocabulary hierarchy

aklochkova:
Fixes https://github.com/OHDSI/Atlas/issues/1063

### DIFF
--- a/js/pages/concept-sets/concept-manager.js
+++ b/js/pages/concept-sets/concept-manager.js
@@ -402,7 +402,6 @@ define([
 		}
 
 		async loadConcept(conceptId) {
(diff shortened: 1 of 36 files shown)


js/pages/concept-sets/concept-manager.js
```diff
@@ -402,7 +402,6 @@ define([
 		}
 
 		async loadConcept(conceptId) {
-			this.currentConceptArray().length = 0;
```
pavgra: Why this was deleted?

pavgra:
@aklochkova , please describe a test case

PR #1095, opened 2018-11-07
Full project bundling

johnSamilin:
Instead of an old bundling process where the script used to download the individual libs and compiled them to one file, this PR introduces using of npm as a provider for 3rd party libraries.
Individual files which used to be stored in the `assets` folder now replaced by `node_modules`.
`npm run build` bundles the whole project and it's external dependencies (not including style files) in one file.
The compilation is done by requirejs optimizer with the help of babel (because `esprima` parser, which is used by requirejs, can't understand es6 features). 
This PR contains tons of changes, all of them are there to make optimization possible:
* there should not be anonymous modules (ie `define((require, exports)`) because it leads to collisions
* some of the relative paths are being changed to absolute ones
* `optional` loader cannot be used durung compile time, besides, it's useless
* `document` is unavailable during compile time, so some additional checks added

To use bundled project, run `npm run build` and then change `index.html`:
`<script data-main="js/main" src="js/require.js"></script>` -> `<script data-main="js/assets/bundle/bundle" src="js/require.js"></script>`

Note: anyway, `npm install` must be executed in order to install 3rd party libraries, even if you're not going to compile the project


pavgra:
@johnSamilin ,

> optional loader cannot be used during compile time, besides, it's useless

So what are the changes introduced in the file then? Why it was left if it cannot be used?

> not including style files

How the styles now are built?

>  then change index.html: <script data-main="js/main" src="js/require.js"></script> -> <script data main="js/assets/bundle/bundle" src="js/require.js"></script>

Maybe we could think of a way, which doesn't require changes of files (resulting in git diff)?

chrisknoll:
I also have concerns with this:

> there should not be anonymous modules (ie define((require, exports)) because it leads to collisions

This is necessary to support circular references.  In the case of the Circe object model, there's containers of things that have situations where A (the container) depends on B (to be contained) but B can contain A (So B references A).  The requireJS docs to talk about how to handle circular deps using the define() syntax: https://requirejs.org/docs/api.html#circular.  So, possibly I need to review that code to ensure it works (or if you have already seen that issue and have resolved it in this PR, I haven't checked)

But, if we're using the requirejs optimizer to do the build, this should work out of the box, so I'm confused about why this would cause a conflict...

> some of the relative paths are being changed to absolute ones
Also, concerned with this.  We should use relative paths whenever some module is referencing some other module that is 'owned' by the same module group.  If it's referencing some completely distinct top-level module, then the absoulte path does make sense. I'd have to check exactly which ones are being changed.

But, if the issue is with some requireJS issue and relative paths, here is some information on that:
https://stackoverflow.com/questions/25897658/can-requirejs-optimizer-work-with-relative-paths
https://raw.githubusercontent.com/jrburke/r.js/master/build/example.build.js

> optional loader cannot be used during compile time, besides, it's useless

This is what lets us override the default config at runtime.  Otherwise, you'll have to resort to doing a full build in order to change some configuration (which I think defeats the purpose: building and configuration should be separate things).  You should be able to exclude certain modules int he build config, so that when the app loads, it will pull the entire app down, but then make the +1 request to get the configuration (from config-local). You should be able to bundle the optional plugin into the main build, or, you can make it run as it does in non-bundled mode where it will download the plugin, then download the optional resource.  So, you wouldn't use optional loader during build, but you would use during runtime.

I think it is worth pointing out that requirejs/amd will allow you to run in a 'mixed' mode where modules that were bundled will be loaded up front, while anything we exclude from the bundling will be downloaded dynamically at run-time.  


> document is unavailable during compile time, so some additional checks added

Makes sense, I was surprised that we depend on document at all, but good to know.

@pavgra : the switching of the 'main' module in the seems to be the 'standard practice' when going from dev mode to optimized build, but I agree, that's a big eyesore. 
 
Looking at this article: http://www.webdeveasy.com/optimize-requirejs-projects/, it looks like the output of the optimizer overwrites the main.js file with the built file, so you don't need to change the index.html.  I believe this is described here: https://requirejs.org/docs/optimization.html#wholeproject.

js/components/cohortbuilder/CriteriaTypes/Criteria.js
```diff
@@ -1,10 +1,9 @@
-define(['require', 'knockout', '../InputTypes/Range','conceptpicker/InputTypes/Concept', '../InputTypes/Text', '../CriteriaGroup'], function (require, ko, Range, Concept, Text, CriteriaGroup) {
+define(['knockout', '../InputTypes/Range','conceptpicker/InputTypes/Concept', '../InputTypes/Text', '../CriteriaGroup'], function (ko, Range, Concept, Text, CriteriaGroup) {
 
 	function Criteria (data, conceptSets) {
 		var self = this;
 		var data = data || {};
 		
-		var CriteriaGroup = require("../CriteriaGroup");
```
chrisknoll: This is the circular reference i was referring to. Without doing it this way, the CriteriaGroup will be null as the require parameter, and you need to do a require() again later. I will pull the branch and verify, however.

js/components/user-bar.js
```diff
@@ -10,7 +10,7 @@ define([
 	'services/JobDetailsService',
 	'services/MomentAPI',
 	'lodash',
-	'less!./user-bar.less'
+	'less!components/user-bar.less'
```
chrisknoll: I don't understand the absolute path here. The user-bar.less file belongs to the user-bar module.

js/components/welcome.js
```diff
@@ -4,7 +4,7 @@ define([
     'appConfig',
     'services/AuthAPI',
 	'utils/BemHelper',
-    'less!welcome.less'
+    'less!components/welcome.less'
```
chrisknoll: Same comment here: welcome.less belongs to welcome module, it should use a relative path.

package.json
```diff
@@ -1,12 +1,13 @@
 {
   "name": "atlas",
-  "version": "2.6.0",
+  "version": "2.6.1",
```
chrisknoll: This is definitely not a patch-fix version update.

chrisknoll:
Additional thoughts on this:

> The compilation is done by requirejs optimizer with the help of babel (because esprima parser, which is used by requirejs, can't understand es6 features).

I also don't understand this:  the bundling should just take all the source files together and concatenate them together.  within the one bundle, there are mutiple calls to define() which will have the optimizer-generated moduleIds that will be available once the main (bundled) module is loaded.  No transpile is required for this.  As a secondary step, we can _minify_ the bundle so that it is compressed, but again, this isn't transpling, and there shouldn't' be anything about 'es6 feature support' here.

I did play with doing a rjs build optimize for the cohort builder, which is a very old branch, but I pushed it here:
https://github.com/OHDSI/Atlas/tree/cohort-builder-rjs/build

This had no problem doing a optimize with define(require) syntax in it, and I'm wondering if I'm missing some underlying issue that is causing the failure that you are seeing.

pavgra:
@chrisknoll ,

> the switching of the 'main' module in the seems to be the 'standard practice' when going from dev mode to optimized build

in Arachne we don't have such a need, that's why the comment took place. however, Arachne requires to be built always (just in dev mode there are different settings) so that's a bit different story

> the bundling should just take all the source files together and concatenate them together

not that simple. but I will leave the ability to describe tricky under-the-hoods of rjs to @johnSamilin

js/components/welcome.js
chrisknoll: Ah, I think the issue with the less!welcome.less is that it should ahev been less!./welcome.less. Could you try that?

js/components/user-bar.js
chrisknoll: This should have worked with the ./ path to the .less file.

johnSamilin:
> No transpile is required for this

Again, r.js uses esprima, which does not understand object spread operator. So without transpiling it's impossible to run optimization.
The second reason for transpiling (ie using `@babel/preset-env`) is that uglify2 does not work with es6 features (ie `const`)

johnSamilin:
> This is necessary to support circular references

Isn't that bad architecture solution?

I've already read all the information about requirejs optimization that's available in the internet.
r.js can optimize module paths in require statements, but seems that it's not working with loaders (ie `less!./style.less`)

chrisknoll:
> Again, r.js uses esprima, which does not understand object spread operator. So without transpiling it's impossible to run optimization.

What is the output of the rjs build when you set this flag in the build config:
```
	optimize:'none',
```

Based on this: https://github.com/requirejs/r.js/issues/971:

> The latest release of r.js (2.3.5) includes the latest esprima (4.0) to parse JS files, so I expect that parsing the file for combining into bundles work

So, I believe you that you're running into issues, but it seems like you should be OK to build using rjs with spread operators, if you set optimize:'none' and do minification as a secondary step.

johnSamilin:
> optimize:'none',

Un-minified bundle will be like 7MB in size.



> The latest release of r.js (2.3.5) includes the latest esprima (4.0) to parse JS files, so I expect that parsing the file for combining into bundles work

I'm not saying esprima doesnt handle array spread (http://esprima.org/demo/parse.html?code=const%20a%20%3D%20%5B1%2C2%2C3%5D%3B%0Aconst%20b%20%3D%20%5B%20...a%20%5D%3B), it's object spread (http://esprima.org/demo/parse.html?code=const%20a%20%3D%20%7B%20foo%3A%20'bar'%20%7D%3B%0Aconst%20b%20%3D%20%7B%20...a%20%7D%3B)

johnSamilin:
> So, I believe you that you're running into issues, but it seems like you should be OK to build using rjs with spread operators, if you set optimize:'none' and do minification as a secondary step.

Btw, esprima is being used before the minification step, when finding nested dependencies, so disabling it won't solve the problem

chrisknoll:
Hey, @johnSamilin , I'm sorry I haven't been able to be more real-time on this thread, I'm in an all day meeting today, tomorrow, so I will reply where I can.  Few things:

> This is necessary to support circular references
> Isn't that bad architecture solution?

I tried to explain how when dealing with nested containers where the container A depends on the items it contains B, but B the item can have a container A...I don't think that's "bad architecture", it's just a necessary requirement of functionality.  But, if you'd be willing to look at that specific section of code to remove the circular dependencies, I believe it would involve an exercise in 'flattening' the object model so all deps are at the same level, but if you have experience in dealing with these types of issue, I'm very happy to work through it with you, and perhaps a separate issue can be raised to deal with that.  It would be a great learning experience for me.

> optimize:'none',
> Un-minified bundle will be like 7MB in size.

I wasn't suggesting that we leave un-optimized source in a bundle, but I was suggesting that we let rjs do it's bundling without optimizing (like with uglify) and then leverage a separate 3rd party optimizer. What I hadn't realized (and thank you for educating me) was that in order to parse the deps out of a source script, rjs uses esprima.  So there's no escaping esprima when going through the rjs biuld.

Ok, so maybe we have 2 options here:

1. limit the use of certain ES6 features in our codebase to be toolchain friendly.  We already do this to some extent where we avoid ES6 features that don't have something like 95% browser support (ex: we avoid using 'default parameter values' in function declarations) and we can extend this to avoid features that aren't friendly to the optimizer.  I think if the issue is just related to 'object spread', let's remove that from the codebase.  I've seen object spread has been problematic in our components since sometimes it's a viewmodel that's being created via object spread, which makes a component completely redraw itself when something in the object changes vs. the model should be a static structure with dynamic content.  So getting rid of object spread there would solve that.

or 2.  make the toolchain do transpile individual source files  to optimize friendly form -> rjs to build bundle -> external minifier to compress code.

chrisknoll:
Just an additional thought: I would ultimately like to extract the cohort builder UI into a stand-alone module (just like we have atlascharts) and I've already done some legwork trying to get an rjs optimized build for just that subset of modules, and so perhaps the issues related to optimizing the cohortbuilder parts of the code can be moot if I get that subsytem of code into a separate repository.

pavgra:
> limit the use of certain ES6 features in our codebase to be toolchain friendly

I think that tools should work for us, not against us, so I would not give away cool convenient code sugar just because some optimizer, which is one of many, doesn't support it

chrisknoll:
> I think that tools should work for us, not against us, so I would not give away cool convenient code sugar just because some optimizer, which is one of many, doesn't support it

Fair point.  So, I trust you and @johnSamilin  are working to offer the best possible solution while trying to take considerations to my eccentric development needs.  Just to summarize what I've been trying to enforce:

> To start working, you just clone repo into a web server's webroot, and you can start developing code.  The only 'tool' needed is git to clone the repo, but you don't even need that if you grab the sourcecode from the git repository.

One way we are diverging from this is that now we have added ```npm``` to the mix, so to get started, you 1) get source code, 2) install npm, 3) run npm install to load deps into /node_modules.  We aren't there yet, but it could go even further to require you to start up a local 'hot-reload' supported HTTP service to develop code.  These things worry me as additional points of failure, and my push-back stems from keeping it as simple as possible.  The other half of developing code is having a back-end to point to so you can see your changes work, which is another technical step to get you started.

But I completely understand why the work in this PR is necessary: NPM solves dependency management.  The build solves the 'off-line' mode.  It improves runtime performance by limiting the server round-trips.

I'd just like to achieve that without making concessions about how we write our code (relative vs. absolute paths or you can't have circular refs (they can be necessary) or the module format can't do ```define(function(require, exports) {} )```.  

If we could try to come to a solution or find the work-around to deal with this last concern, it'd be much appreciated. but I understand sometimes we just don't have a choice.  Please be patient with me as I ask questions/offer suggestions that will help us not write specialized code to deal with the tool-chain.  The beauty of this PR and branch is that I can try some things myself so I won't bother you with some ideas, but I also know you have expertise with the tool, so I might just ask you a question out of lazyness.

chrisknoll:
A bit of a tangent, but I'd like to capture it here while we're working on this effort:

The idea here is to take all the 3rd party libraries and bundle them in to one huge 7MB application bundle.  This goes completely against the dynamic loading nature of the AMD architecture.  The set of libraries that (I assume) will be bundled into the final build:

- knockout
- jquery
- d3
- d3's 7ish componet libraries that are redundant to the main d3 library
- atlascharts
- underscore
- datatables
- numeric
- jqueryui
- probalby 5 or 6 more libraries that I can't think of.

When going right to the home page, we don't need d3, we don't need underscore, we don't need numeric, ... the list of libraries we need is probably shorter: knockout and jquery.  If we then navigate over to something else (like datasources) then we need datatables, so that should result in some dynamic load.  

So my question is: why are we trying to make an effort to lump all our dynamically-loadable modules into one huge file to download, when the technology we're leveraging here (asynchronous module definitions) supports lazy loading?  I feel it would be enough to have all the NPM dependent modules exuded from the build, and just compile together the individual Atlas source files into the bundle.  I apologize if the reason is something obvious I should recognize; I am just trying to reconcile the dynamic loading we have today with the 'load everything up front' approach we're trying for in this build process.

johnSamilin:
Well, maybe I undesrstood some of the output wrong, I've reverted anonymous modules and absolute paths and it worked. Not sure what produced the errors that made me do those changes

chrisknoll:
Hey, I'm sorry my branch wasn't a complete example.  There are additional path-config changes that are necessary so that the loader resolves the module to the dist build (which I didn't include in the branch).  I'm nearly 100% certain tho, that the optimizer did do its job.  This was just a quick POC that optimizer could handle certain module definition syntax without error.  I don't propose that we continue on my old branch to get that to work, we should focus on this current branch (related to this PR) and push this forward.  I'm just trying to minimize the source code changes that are necessary to get us to the point of an optimized output.

johnSamilin:
@chrisknoll I've reverted anonymous modules and absolute paths, do you have any other comments regarding this PR?

chrisknoll:
Checking it this morning.  Only concern will be the removal of the local require() to handle the nested criteria via circular references.  But I'll see if it works properly in build and non-built mode.

chrisknoll:
@johnSamilin , I got it pulled, i was looking through the package.json deps, and was curious why we are using the ```*```, and was wondering if we shouldn't be more specific about the version of the dependencies. For example, atlascharts is ```"@ohdsi/atlascharts": "*",``` and we have semver compliance with that library, so I'd expect we'd say ^1.0 or something like that for that library.

.gitignore
```diff
@@ -1,6 +1,5 @@
 .idea
 /web.config
-js/config-local.js
```
chrisknoll: I noticed this too late, but this should not have been ignored. This had my local configuration setup in it, and when I pulled down the branch, this file was blown away. :(

chrisknoll:
Starting up this branch seemed to have security enabled.  my config-local.js was blown away and so now i don't see how we're supposed to override local configuration changes.

I think you understand how local configuration should work but just to be clear:  config-local should not be added to the final bundle.  the optional! plugin can, but when you download the application (in bundled mode), you should get the full applicaiton bundle, and then there's a +1 request to config-local.js to load that module dynamically.  This should be possible under the AMD module loading capability.  Please take a look at this behavior, and I'll try to figure out if I have a backup of my config-local.js file somewhere.

johnSamilin:
@chrisknoll I've fixed packages versions and also found a way to prepare `optional` loader for the bundling (https://requirejs.org/docs/plugins.html#apiload `config.isBuild`)

js/config/app.js
```diff
@@ -8,9 +8,8 @@ define(function () {
   };
   appConfig.cacheSources = false;
   appConfig.pollInterval = 60000;
-  appConfig.useBundled3dPartyLibs = false;
 	appConfig.cohortComparisonResultsEnabled = false;
-	appConfig.userAuthenticationEnabled = false;
+	appConfig.userAuthenticationEnabled = true;
```
chrisknoll: We shouldn't switch this default value from false to true, it's an incompatible change.

chrisknoll:
Error opening up a cohort definition which has a date criteria specified

image

image

Date picker is a databinding which depends on jqueryui.

chrisknoll:
I haven't dug into PLE/PLP yet.   I am very familiar with the dependencies of cohort definition and incidence rates so that's where I started by tests.  I'll check those functions (perhaps @anthonysena could also help do a regression test on those functions) and once I feel everything is relatively bug-free, I will check on the bundled mode.

chrisknoll:
Running the bundler now. seeing this in the output:
image

Is that a problem?

chrisknoll:
Ok, good news is: bundle completed and web application launched!  

I noticed that .less files are not being compiled into the bundle and so you get this:

image

276/333 of the requests were for `.less` files.  I'm wondering if we use a less plugin that works with the optimizer, if that could be improved.

chrisknoll:
Sourcemaps are working good:

image

Notice that the text plugin is bundling those strings inline and they show text as an actual resource. Would be asesome if less plugin worked like this.

chrisknoll:
Great, sorry I have been in meetings all day, I will verify tonight. Does this address the issue I saw where the unbundled mode makes repeated requests to ``white-page.less``?  I will check for this, but I wanted to ask you to see if you had a fix for that.

johnSamilin:
> Great, sorry I have been in meetings all day, I will verify tonight. Does this address the issue I saw where the unbundled mode makes repeated requests to `white-page.less`? I will check for this, but I wanted to ask you to see if you had a fix for that.

No, it just bundles all the .less
I can't reproduce this issue now, so I hope it won't appear

johnSamilin:
I've briefly tested and it seems ok

Issue #1101, opened 2018-11-08
Can't create a Source in empty database


aklochkova:
applies to v. 2.4.5
### Steps to reproduce behavior

1. database is empty
2. login and see warning 'the current webapi has no sources defined. please add one or more on configuration page.'
3. go to configuration page and press 'New Source'

### Expected behavior
page with new Source is opened
### Actual behavior
warning 'the current webapi has no sources defined. please add one or more on configuration page.' appears

pavgra:
@aklochkova , attempt to create a DS fails with following error:
```
[...]

PR #1102, opened 2018-11-08
Fixed observables for source-manager

aklochkova:
Fixes https://github.com/OHDSI/Atlas/issues/1101

### DIFF
--- a/js/modules/WebAPIProvider/VocabularyProvider.js
+++ b/js/modules/WebAPIProvider/VocabularyProvider.js
@@ -13,39 +13,43 @@ define(function (require, exports) {
 	var defaultSource;
 	var domainPromise = $.Deferred();
 	var domains = [];
-
-	sourceAPI.getSources().then(function (sources) {
-		if (sources.length == 0) {
-			sharedState.appInitializationStatus('no-sources-available');
-			return;
-		}
-		// find the source which has a Vocabulary Daimon with priority = 1
-		var prioritySources = sources.filter(function (source) {
-			return source.daimons.filter(function (daimon) {
-				return daimon.daimonType == "Vocabulary" && daimon.priority == "1"
-			}).length > 0
-		});
-		if (prioritySources.length > 0)
-			defaultSource = prioritySources[0];
-		else // find the first vocabulary or CDM daimon
-			defaultSource = sources.filter(function (source) {
-				return source.daimons.filter(function (daimon) {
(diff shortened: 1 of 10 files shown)


js/modules/WebAPIProvider/VocabularyProvider.js
```diff
@@ -35,14 +35,16 @@ define(function (require, exports) {
...
 		// preload domain list once for all future calls to getDomains()
-		$.ajax({
-			url: config.webAPIRoot + 'vocabulary/' + defaultSource.sourceKey + '/domains',
-		}).then(function (results) {
-			$.each(results, function (i, v) {
-				domains.push(v.DOMAIN_ID);
-			});
-			domainPromise.resolve(domains);
-		});
+		if (defaultSource !== undefined){
```
pavgra: What happens if we start Atlas w/o sources and add a source via UI? The default vocab is not loaded?

chrisknoll:
Sorry, question: do we need to have getX and getXPromise in this module API?  Seems stange.  we understand that these service APIs are synchronous, so getDomains should return a promise.  Could we not reshuffle the code like this?

js/modules/WebAPIProvider/VocabularyProvider.js
```diff
@@ -118,7 +124,7 @@ define(function (require, exports) {
 		return deferred.promise();
 	}
 
-	function getDomains() {
+	function getDomainPromise() {
```
chrisknoll: Just need to revert this rename and just have 'getDomains' exposed int he api below.

js/modules/WebAPIProvider/VocabularyProvider.js
```diff
@@ -13,37 +13,43 @@ define(function (require, exports) {
...
-		});
-	})
+	
+	getDomains();
+	
+	function getDomains() {
+        sourceAPI.getSources().then(function (sources) {
+            if (sources.length == 0) {
+                sharedState.appInitializationStatus('no-sources-available');
+                return;
```
chrisknoll: I believe at this point, when we have no sources, we can resolve the domains promise to either [] or null.  Return [] if domains is expected to return a list.

chrisknoll:
So, I have some additional updates to this I'd like to make but realized that this is for a 2.4.5 branch, and I'm only really interested in updates to master.  Is there a separate PR for this for the master branch, or is that coming later?

PR #1123, opened 2018-11-16
Fixed bootstrap behavior about appInitializationStatus

aklochkova:
Fixes https://github.com/OHDSI/Atlas/issues/1101


js/services/SourceAPI.js
```diff
@@ -15,13 +15,11 @@ define(function (require, exports) {
 	function getSources() {
 		var promise = $.ajax({
 			url: config.webAPIRoot + 'source/sources/',
-			error: function (error) {
+			error: function () {
 				sharedState.appInitializationStatus('failed');
```
pavgra: Use constant

js/services/SourceAPI.js
```diff
@@ -94,7 +95,13 @@ define(function (require, exports) {
 
     return servicePromise;
   }
-
+  function setAppInitStatus(sources){
+      if (sources.length !== 0){
+          sharedState.appInitializationStatus('running');
```
pavgra: GlobalModel.applicationStatuses.running

js/services/SourceAPI.js
```diff
@@ -94,7 +95,13 @@ define(function (require, exports) {
...
+      } else {
+          sharedState.appInitializationStatus('no-sources-available');
```
pavgra: Constant

pavgra:
@johnSamilin , please review logic related to initializationStatus

js/Application.js
```diff
@@ -221,19 +215,23 @@ define(
...
 						} else {
-							sourceApi.initSourcesConfig().then(() => {
-								console.info('Init sources from server');
-								resolve();
-							});
+							sourceApi.initSourcesConfig()
+								.then(function (appStatus) {
+									sharedState.appInitializationStatus(appStatus);
+								})
+								.then(() => {
```
pavgra: Why do you need `then` here if previous step (`sharedState.appInitializationStatus(appStatus);`) doesn't return promise?

js/pages/configuration/sources/source-manager.js
```diff
@@ -265,6 +267,10 @@ define([
       this.loading(true);
       sourceApi.saveSource(this.selectedSourceId(), source)
         .then(() => sourceApi.initSourcesConfig())
+        .then(function (appStatus) {
+            sharedState.appInitializationStatus(appStatus);
```
pavgra: Where is return for promise chain? If there is no such - why separate step?

js/pages/configuration/sources/source-manager.js
```diff
@@ -311,6 +317,9 @@ define([
       this.loading(true);
       sourceApi.deleteSource(this.selectedSourceId())
         .then(sourceApi.initSourcesConfig)
+        .then(function (appStatus) {
+            sharedState.appInitializationStatus(appStatus);
```
pavgra: Same as above

js/services/VocabularyProvider.js
```diff
@@ -11,39 +11,47 @@ define(function (require, exports) {
...
+						$.each(results, function (i, v) {
+							domains.push(v.DOMAIN_ID);
+						});
+						resolve(domains);
+					});
+				}
+			})
 		});
-	})
+		return domainsPromise;
```
pavgra: Code duplication

js/services/Vocabulary.js
```diff
@@ -12,39 +12,47 @@ define(function (require, exports) {
 	loadedPromise.resolve();
 
 	var defaultSource;
-	var domainPromise = $.Deferred();
 	var domains = [];
```
pavgra: Why global variable is still used?

js/services/Vocabulary.js
```diff
@@ -12,39 +12,47 @@ define(function (require, exports) {
...
-
-		// preload domain list once for all future calls to getDomains()
-		$.ajax({
-			url: config.webAPIRoot + 'vocabulary/' + defaultSource.sourceKey + '/domains',
-		}).then(function (results) {
-			$.each(results, function (i, v) {
-				domains.push(v.DOMAIN_ID);
-			});
-			domainPromise.resolve(domains);
+	getDomains();
```
pavgra: Why do you need to call it?

js/Application.js
johnSamilin: all the next `.then` will trigger if previous `.then` returns undefined

pavgra: I haven't said that they won't. I was confused why the step was moved into separate `then` (earlier it wasn't)?

js/services/Vocabulary.js
chrisknoll: This triggers the domain initialization when the module is first loaded so that it's 'ready' when we get to a place in the UI that wants the domain list (but don't want to make the users wait).

I found the other PR for 2.4's branch (or was it 2.5?) and I'd like to push some changes to this PR's branch to try to account for the origional design goal: to load the domain list once per application start (once per module load, which modules are only loaded once per app launch).  I'll push a commit to this PR's branch as a suggestion, and if you guys like, take it, if not, i'll revert it.

Stand by, should only take me a few minutes.

js/services/Vocabulary.js
chrisknoll: Yes, we will want a global variable to hold the promise.  I am pushing up a commit to demonstrate this.

chrisknoll:
Ok, I've resolved conflicts between master and this branch.  summary of my commit:

1. global domainsPromise:  We only want to fetch the list of domains once per application launch.  It's a little bit of a server round trip for a list of domains which should be static, even across CDM versions.  But, if the user does have an issue with an out-of-date domains list, just reload the application and it will init again.  If we want to think about a more robust 'refresh domain' capability, we can think about that in 3.0.

The logic change is just to check for the promise being created at the start of getDomains(), and if it's not, create it.


2. services/VocabularyProvider module: From what I could tell, this file was duplicated, and we were left to make changes to both files for fixes.  That's a nightmare.  So, instead, i've removed VocabularyProvider, and set up a map in settings.js to map lookups to services/VocabularyProvider to services/Vocabulary, which saves us the trouble of finding all the moduleID refernces to services/VocabularyProvider.  later, we can go back and upadate the old refs across the different dependencies so taht we don't need this map configuration, but it works.

let me know if anyone has any objections and I can revert this.

js/services/Vocabulary.js
pavgra: @chrisknoll ,

1. You open Atlas app first time after deployment, there are no sources in the system
1. Your getDomains() function is automatically called, resulting in empty domains array
1. You go to Configuration tab, add Data source
1. Go to Cohort Definition / other place which relies on the domains list and observe incorrect behavior. Why is that a problem to load things on-demand and cache them. But on-demand, not on app startup?

chrisknoll: Yes, I pointed this out as a limitation:  if it only checks once per app load, then if you add sources later, you don't see the domains refresh (without further work to clear caches and whatnot).  The way to solve this would be to reset the domainPromise in the vocabAPI when sources get upadted.  But it just seems a lot of work for a 2.x series which will be probably quite different in 3.x, and the cost of reloading the app seems preferable to constantly polling the services + querying the vocabulary for the domains every time you want to get the domains (it's not a single service call it's 2: 1 to refresh the sources, and 2 to get the list of domains). Just doesn't feel worth it.

chrisknoll: Ok, so I got an idea, I'll push it up into a new commit but the idea is:  while we don't have sources, resolve a domainsPromise to []. Once we have a source with vocab, we can cache the domainPromise that resolves to the list of domains.  Shoudln't take long, stand by.

chrisknoll:
Ok, I pushed the update, it was relatively simple. Changes:
1. Remove dcall to getDomains() at startup.
2. If domainsPromise exists, re-use the promise for subsequent calls (caching the result list)
3. If we don't have a promise with domains yet, load the sources and search for a vocabulary daimon. If none found, resolve the promise to [].  If we do find a source with vocabulary, resolve the promise to the set of domains, and store the promise in the module-global 'domainsPromise'.  Subsequent calls will use this result without further interaction to the web services.

PR #1156, opened 2018-11-21
Fixed display of name in Concept Set header

### LINKED ISSUE #1131: Screen displays  “New Concept Set” for existing concept set
image

johnSamilin:
fixes #1131

### DIFF
--- a/js/pages/concept-sets/conceptset-manager.html
+++ b/js/pages/concept-sets/conceptset-manager.html
@@ -1,4 +1,4 @@
-<heading-title params="name: conceptSetName(), icon: 'shopping-cart', theme: 'dark'"></heading-title>
+<heading-title params="name: $root.currentConceptSet().name, icon: 'shopping-cart', theme: 'dark'"></heading-title>
 
 <div data-bind="css: classes()">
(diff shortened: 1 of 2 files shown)


js/pages/concept-sets/conceptset-manager.html
```diff
@@ -1,4 +1,4 @@
-<heading-title params="name: conceptSetName(), icon: 'shopping-cart', theme: 'dark'"></heading-title>
+<heading-title params="name: $root.currentConceptSet().name, icon: 'shopping-cart', theme: 'dark'"></heading-title>
```
chrisknoll: If this is inside the context of a component, and the `currentConceptSet` is a property of the component model, then use `$component.curentConceptSet()`.  However, there isn't a `currentConceptSet()` field of the model, instead, there's an assumption that the `$root` context is the `Application` viewmodel, which works, but it's not right.

Quick fix: if we add a `pureComputed` called `currentConceptSet` as defined as:

```
this.currentConceptSet = ko.pureComputed( () => return this.model.currentConceptSet());
```

Then you can use `$component.currentConceptSet().name` in the html template and it's clear where you can find the currentConceptSet's observable.

PR #1157, opened 2018-11-21
fix build due to inconsistency in hash sums

johnSamilin:
current `package-lock.json` is generated on windows machine and there are some issues with building on linux one:
`npm WARN tarball tarball data for file-saver@1.3.8 (sha512-spKHSBQIxxS81N/O21WmuXA2F6wppUCsutpzenOeZzOCCJ5gEfcbqJP983IrpLXzYmXnMUa6J03SubcNPdKrlg==) seems to be corrupted. Trying one more time.
npm ERR! code EINTEGRITY`


pavgra:
Hopefully, this is temporary change

chrisknoll:
I was noticing this too, but thought it was just user error... when I did an npm install it was regenearting the entire .lock file even tho nothing looked changed.  I think it might be some line-ending diff between platforms or something.

PR #1164, opened 2018-11-26
Adds Cohort Characterization stratification

wivern:
add stratification for OHDSI/WebAPI#649


chrisknoll:
Should we think about converting all of our sources to tab-indents?

js/pages/characterizations/components/characterizations/characterization-view-edit/characterization-design.js
```diff
@@ -164,10 +181,57 @@ define([
...
             this.isParameterCreateModalShown(true);
         }
+
+        handleConceptSetImport(criteriaIdx, item) {
+          console.log('import', item);
+          this.criteriaContext({...item, criteriaIdx});
+          this.showConceptSetBrowser(true);
+        }
+
+			onRespositoryConceptSetSelected(conceptSet, source) {
```
pavgra: Looks very similar to the method with same naming in `feature-analysis-view-edit`. Let's create a common class to exclude code duplication

js/pages/characterizations/components/characterizations/characterization-view-edit/characterization-results.html
```diff
@@ -82,6 +82,35 @@ <h3 data-bind="css: $component.classes('analysis-name'), text: ($data.domainId ?
                             </div>
                         <!-- /ko -->
                     </div>
+
+                    <!-- ko foreach: $data.stratas -->
```
pavgra: Why stratas go into a different table?

js/pages/characterizations/components/characterizations/characterization-view-edit/characterization-results.html
```diff
@@ -139,6 +168,52 @@ <h3 data-bind="css: $component.classes('analysis-name'), text: ($data.domainId ?
                             <!-- /ko -->
                         <!-- /ko -->
                     </div>
+                    <!-- ko if: reports.length !== 2 -->
```
pavgra: What is this for?

js/pages/characterizations/components/characterizations/characterization-view-edit/characterization-results.html
wivern: I believe we don't want to mix stratified results with ordinal analyses results, am I wrong or misunderstood your question?

js/pages/characterizations/components/characterizations/characterization-view-edit/characterization-results.html
wivern: Just to follow common behavior stratified results has the comparative mode in case when two reports in characterization

js/pages/characterizations/components/characterizations/characterization-view-edit/characterization-results.js
```diff
@@ -369,20 +393,19 @@ define([
             const convertedData = filteredData.map(analysis => {
                 let convertedAnalysis;
 
+                analysis.strataOnly = true;
```
pavgra: Would such option be better readable?

```
const stats = _.flatMap(analysis.reports, r => r.stat)
const stratifiedCnt = stats.filter(stat => stat.strataId > 0).length

analysis.strataOnly = stats.length === stratifiedCnt
analysis.stratified = stratifiedCnt > 0
```

PR #1165, opened 2018-11-26
Data sources fixes

fdefalco:
original files had spaces instead of the standard tab for whitespace so recommend adding &w=1 to the url to ignore the whitespace corrections.

### DIFF
--- a/js/pages/data-sources/components/reports/achillesHeel.html
+++ b/js/pages/data-sources/components/reports/achillesHeel.html
@@ -1,11 +1,11 @@
 <!-- ko if: !isLoading() -->
-<heading-title params="name: title, icon: 'area-chart'"></heading-title>
+<div class="dataSourceReportHeading" data-bind="text:title"> </div>
 <div class="panel panel-primary">
-  <div class="panel-heading">
-    Achilles Heel Results
-  </div>
-  <div class="panel-body">
-    <faceted-datatable params="columns: columns, reference: data"></faceted-datatable>
-  </div>
+	<div class="panel-heading">
+		Achilles Heel Results
+	</div>
+	<div class="panel-body">
+		<faceted-datatable params="columns: columns, reference: data"></faceted-datatable>
+	</div>
 </div>
 <!--/ko-->
--- a/js/pages/data-sources/components/reports/treemapDrilldown.js
+++ b/js/pages/data-sources/components/reports/treemapDrilldown.js
@@ -2,8 +2,8 @@ define([
 	'knockout',
 	'text!./treemapDrilldown.html',
 	'd3',
-    'atlascharts',
-    'utils/CommonUtils',
+	'atlascharts',
+	'utils/CommonUtils',
 	'utils/ChartUtils',
 	'const',
 	'pages/data-sources/classes/Report',
@@ -21,8 +21,8 @@ define([
 	ko,
 	view,
 	d3,
-    atlascharts,
-    commonUtils,
+	atlascharts,
+	commonUtils,
 	ChartUtils,
 	constants,
 	Report,
@@ -31,7 +31,7 @@ define([
 	class TreemapDrilldown extends Report {
 		constructor(params) {
 			super(params);
-			 
+
 			this.currentConcept = ko.observable({
 				name: '',
 			});
@@ -59,13 +59,13 @@ define([
 			this.upperLimitDistributionData = ko.observable();
 			this.recordsByUnitData = ko.observable();
 			this.valuesRelativeToNormData = ko.observable();
-			
+
 			this.commonBoxplotChartOptions = {
 				yMax: 0,
 				xLabel: 'Unit',
 				yLabel: 'Measurement Value',
 			};
-			
+
 			this.chartFormats = {
 				prevalenceByMonth: {
 					xScale: null,
@@ -128,7 +128,7 @@ define([
 					}
 				},
 			};
-			
+
 			this.currentReport = params.currentReport;
 			this.byFrequency = params.byFrequency;
 			this.byUnit = params.byUnit;
@@ -141,7 +141,7 @@ define([
 		}
 
 		parseAgeData(rawAgeData) {
-				this.ageData(this.parseBoxplotData(rawAgeData).data);
+			this.ageData(this.parseBoxplotData(rawAgeData).data);
 		}
 
 		parsePrevalenceByMonth(rawPrevalenceByMonth) {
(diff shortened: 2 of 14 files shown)


js/pages/data-sources/components/reports/treemapDrilldown.js
```diff
@@ -282,26 +305,27 @@ define([
...
-			this.isLoading(true);
 			this.getData()
-				.then((data) => this.parseData(data))
+				.then((data) => {
+					this.parseData(data);
+					this.context.model.loadingReportDrilldown(false);
+					this.scrollTo("#datasourceReportDrilldownTitle");
+				})
 				.catch((er) => {
+					this.context.model.loadingReportDrilldown(false);
```
johnSamilin: those two lines could be moved to `finally` block

js/pages/data-sources/data-sources.js
```diff
@@ -105,14 +104,17 @@ define([
 			];
 
 			this.model = params.model;
-			this.sources = ko.computed(() => {
-				return sharedState.sources().filter(function (s) {
-					return s.hasResults && s.hasCDM;
-				});
+			this.sources = sharedState.sources().filter(function (s) {
```
johnSamilin: why did you make it not computed? I guess it will make you reload the page if you went on it unauthenticated and then logged in

js/pages/data-sources/data-sources.js
```diff
@@ -121,9 +123,51 @@ define([
...
+
+			this.selectedReportSubscription = this.selectedReport.subscribe(r => {
+				this.updateLocation();
+			});
+
+			this.selectedSourceSubscription = this.currentSource.subscribe(r => {
+				this.updateLocation();
+			})
+
+			this.updateLocation = function () {
```
johnSamilin: `updateLocation`, `onRouterParamsChange` and `dispose` should be class methods

js/pages/data-sources/data-sources.js
```diff
@@ -121,9 +123,51 @@ define([
 			this.showSelectionArea = params.showSelectionArea == undefined ? true : params.showSelectionArea;
 			this.currentSource = ko.observable(this.sources[0]);
 			this.currentReport = ko.observable();
+			this.selectedReport = ko.observable();
```
johnSamilin: what is the difference between `currentReport` and `selectedReport`?

js/pages/data-sources/data-sources.less
```diff
@@ -1,4 +1,8 @@
 .data-sources {
+	& .drilldown-enabled tr,rect {
```
johnSamilin: these rules will result in the following css:
```
.data-sources .drilldown-enabled tr,
.data-sources rect
```
is that what you're trying to achieve?

js/pages/data-sources/data-sources.less
```diff
@@ -17,16 +21,18 @@
     }
   }
 
-  &.dataSourceReport {
-    width: 96%;
-    margin-left: auto;
-    margin-right: auto;
-    margin-top: 0px;
-  }
+	& .dataSourceReportHeading {
```
johnSamilin: I think since this block is being used on every report, it sould be a separate component with its own `.less` file, controller and a template

js/pages/data-sources/data-sources.js
fdefalco: It also triggered a second reload when the report drop down was changed and was causing issue with the change to make the report loading driven by routes.  It seems safe to say that since the sources are loaded once at startup that computing the list of sources that have results or cdm with each reference to the variable was unnecessary.

js/pages/data-sources/data-sources.less
fdefalco: No, was trying to achieve:

.data-sources .drilldown-enabled tr,
.data-sources .drilldown-enabled rect

Can you suggest the proper rules?

js/pages/data-sources/components/reports/treemapDrilldown.js
```diff
@@ -128,20 +128,31 @@ define([
...
 			this.byFrequency = params.byFrequency;
 			this.byUnit = params.byUnit;
 			this.byType = params.byType;
 			this.byValueAsConcept = params.byValueAsConcept;
 			this.byOperator = params.byOperator;
 			this.byQualifier = params.byQualifier;
-			params.currentConcept.subscribe(this.loadData.bind(this));
+			this.context = params.context;
+			this.currentConceptSubscription = params.currentConcept.subscribe(this.loadData.bind(this));
+			this.dispose = function () {
```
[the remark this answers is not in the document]
fdefalco: What is it that should be moved to a class method?

js/pages/data-sources/components/reports/treemapDrilldown.js
fdefalco: resolved in new commit commit.

js/pages/data-sources/data-sources.js
fdefalco: In changing the page to be route driven I created the selectedReport observable.  The selectedReport observable is bound to the drop down.  The currentReport observable is bound to the component that loads the report.  The routes already existed but did not work.  The route did not update when the currentReport observable changed nor did the report load if the user went directly to the URL containing the sourceKey and reportName.  The new behavior uses the selectedReport to manage the state of the drop down.  When the selectedReport observable is modified it triggers an update to the document.location which is then handled via routes.

js/pages/data-sources/data-sources.less
chrisknoll: I think this:

```
.data-sources {
	&.drilldown-enabled {
    	tr, rect {
      font-size: 1px // add styles here
		}
  	}
}
```

Validated here: http://winless.org/online-less-compiler

fdefalco: Fixed in https://github.com/OHDSI/Atlas/pull/1165/commits/c5d33800a2c54993a6ce045d8e5a375270fc1d58

js/pages/data-sources/data-sources.js
fdefalco: Resolved in https://github.com/OHDSI/Atlas/pull/1165/commits/95e3a3b6ec50ce77c74b79a4227261dbcb89617d

js/pages/data-sources/components/reports/achillesHeel.html
```diff
@@ -1,11 +1,11 @@
 <!-- ko if: !isLoading() -->
-<heading-title params="name: title, icon: 'area-chart'"></heading-title>
+<div class="dataSourceReportHeading" data-bind="text:title"> </div>
```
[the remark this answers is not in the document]
fdefalco: Given this and the later comment on having a shared component and conversation with @pavgra I've reverted to using the heading component but made the icon property optional.

https://github.com/OHDSI/Atlas/pull/1165/commits/d96ab3957d664838b3b70f87a65f095b9278e827

js/pages/data-sources/data-sources.less
fdefalco: reverted to shared heading component
https://github.com/OHDSI/Atlas/pull/1165/commits/d96ab3957d664838b3b70f87a65f095b9278e827

js/pages/data-sources/components/reports/treemapDrilldown.js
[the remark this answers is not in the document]
johnSamilin: `dispose`
I mean, instead of adding methods this way, it's better to use ES6 aproach:
`class Foo {
  dispose() { ... }
}`

PR #1166, opened 2018-11-27
Fixed input cohort name XSS...

NEONKID:
Issue WebAPI [382](https://github.com/OHDSI/WebAPI/issues/382)

This is an issue that appeared in ATLAS using the editable attribute in the div tag.
(Ref: https://html5sec.org/)

When sending a value to the WebAPI, we took action to remove the tag and used keyboard events to block the characters before and after the tag.

### DIFF
--- a/js/pages/cohort-definitions/cohort-definition-manager.html
+++ b/js/pages/cohort-definitions/cohort-definition-manager.html
@@ -105,7 +105,7 @@
 						<div class="heading">
 							Name:
 						</div>
-						<div class="divtext" data-bind="attr: { contenteditable:canEdit() }, htmlValue:$root.currentConceptSet().name"></div>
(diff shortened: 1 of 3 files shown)


NEONKID:
So when I install ATLAS, can I run it after the build with yarn or npm?

johnSamilin:
yes, you can require it in `cohort-definition-manager.js` and thus get rid of the regular expression

chrisknoll:
I think the real fix should go into the htmlValue binding, which is used to bind the value from the div content to the viewmodel:
https://github.com/OHDSI/Atlas/blob/master/js/extensions/bindings/htmlValueBinding.js

We use the xss module the same way we use it in the datatable binding:
https://github.com/OHDSI/Atlas/blob/master/js/extensions/bindings/datatableBinding.js#L79

See how it gets the xss config and then applies the xss filter on the input.  

This will capture all places where we use the htmlValue binding, without having to find each instance.

Otherwise, the defautl ko bindings should handle xss issues.  This is just a custom binding that doesn't handle it (like the datatable one).

Issue #1078, opened 2018-10-29
UI improvements for CC


pavgra:
1. Add facets to filter by Domain in Feature analyses list and Feature analyses popup
1. Allow picking multiple Feature Analyses from Feature analyses popup inside CC page (inc. select all filtered)
1. Add filtering by analysis domain into CC results page
1. Return back "Explore" button for concepts presented in CC results (like it was before in Generate w/ features)
![image](https://user-images.githubusercontent.com/10010071/47681662-76d94180-dba0-11e8-8af5-e74f6433c73a.png)

@chrisknoll , I remember we had some more (not speaking here about Generations tab) but was not able to find the comment which you or @pbr6cornell posted after our call

chrisknoll:
When either a feature or characterization project is open, the left nav does not indicate an 'open' project (green indicator) or if something in the nav is dirty (red indicator).

By clicking on the left nav, it should take you to the open project (if there is something open).

However, unlike all the other navs, you could have 2 different things open at the same time under cohort characterization. So, recommend you always have 2 tabs beneath the characterization nav, and beneath each tab, you have either the 'open' design, or the design list (either characterization or feature).  You may want to store the 'state' of the active tab so that when you click on the left nav, it returns to the last 'selected' tab that the user selected (the left nav's URL will be dynamic, based on the last 'active tab' so clicking on the left nav will direct you to the route of the selected sub-tab).

Hope that makes sense, if there was some other behavior you prefer, I'm happy to review it with you.

pavgra:
@wivern , address missed margins pls
![image](https://user-images.githubusercontent.com/10010071/47790349-4f43bf80-dced-11e8-9b5d-20768a702c26.png)

pavgra:
@wivern , recently there were updates to dirtyFlag caused by performance issues. From now, the flag is reacting to observables changes only. Thefeore, we need to update handling of design and make it similar to Pathways approach. Currently, it is impossible to create new / update design - button is always disabled.

pavgra:
@wivern , 

1. Why does this screen

- Show each line 2 times?
- Do not allow to differentiate between cohorts? - I would even propose to show only 1 cohort per popup or at least introduce selector and additional column / caption

![image](https://user-images.githubusercontent.com/10010071/48574744-f4f65180-e8dd-11e8-8cff-f27db3f2fa68.png)

2. Doesn't seem that anything is happening:

[...]

PR #1176, opened 2018-11-30
Added explore prevalence with stratas

wivern:
Added explore prevalence with stratas as part of #1078

### DIFF
--- a/js/pages/characterizations/components/characterizations/characterization-view-edit/characterization-results.js
+++ b/js/pages/characterizations/components/characterizations/characterization-view-edit/characterization-results.js
@@ -1,7 +1,7 @@
 define([
     'knockout',
     'pages/characterizations/services/CharacterizationService',
-	  'pages/characterizations/services/FeatureAnalysisService',
+    'pages/characterizations/services/FeatureAnalysisService',
     'text!./characterization-results.html',
     'appConfig',
     'services/AuthAPI',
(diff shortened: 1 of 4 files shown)


pavgra:
1. Open Explore:

image

1. Go to Ancestor

image

1. Go to Descendant (so equal to go back)

image

Breadcrumb is still in place

pavgra:
@chrisknoll , I do think that such check could be easily achieved by comparison of concept ID used to open the explore window and the current concept ID

chrisknoll:
Maybe i'm missing something about 'explore':  is explore used to limit the results to just those items that are in the ancestor/descendant chain of the selected concept?  In the original feature extraction functionality, you would start by looking across all features, and if you wanted to filter directly to a hierarchy chain, you click 'explore'.  When you are done with that narrow view, you click on the X to go back to all results.  

My question is: if you click on the original concept again to explore, does that really mean you want to cancel?

chrisknoll:
I'd appreciate it if we waited to merge these things until questions were completely addressed.

pavgra:
> is explore used to limit the results to just those items that are in the ancestor/descendant chain of the selected concept

yes

> In the original feature extraction functionality, you would start by looking across all features, and if you wanted to filter directly to a hierarchy chain, you click 'explore'. When you are done with that narrow view, you click on the X to go back to all results.

Don't think that anything changed. Our goal was just to return back the lost functionality

chrisknoll:
> Don't think that anything changed. Our goal was just to return back the lost functionality\

Good, so if we go by the original functionality:

Clicking explore on any feature limits results to those features to the hiearchy.
You can click explore again to go higher or lower into a hiearchy but still limit the features to those in the explored hiearchy.
Clicking back to explore the original concept _did not_ clear the explore mode.  If you wanted to return to 'all results' you would clear the selection in the top (the breadcrumb).

pavgra:
I believe the easiest way would be to pull master down and try it out, honestly. Here we have popup, so the initial screen is for ascendants / descendants of a concept picked from report-level prevalence table. Therefore, when you open popup, there is no breadcrumbs initially

Issue #1178, opened 2018-12-03
Logic of Concept Hierarchy viewer


pavgra:
Today there is a hardcoded list of vocabularies and concept classes which are displayed in Concept Hierarchy viewer:

<details><summary>CLICK to show vocab - concept class list</summary>
<p>

```
'ICD9CM.5-dig billing code': {
  childRelationships: [{
    name: 'Subsumes',
    range: [0, 1]
  }],
  parentRelationships: [{
    name: 'Is a',
    range: [0, 1]
  }]
},
'ICD9CM.4-dig nonbill code': {
  childRelationships: [{
    name: 'Subsumes',
    range: [0, 1]
  }],
  parentRelationships: [{
    name: 'Is a',
    range: [0, 1]
  }, {
    name: 'Non-standard to Standard map (OMOP)',
    range: [0, 1]
  }]
},
'ICD9CM.3-dig nonbill code': {
  childRelationships: [{
    name: 'Subsumes',
    range: [0, 1]
  }],
  parentRelationships: [{
    name: 'Non-standard to Standard map (OMOP)',
    range: [0, 999]
  }]
},
'RxNorm.Ingredient': {
  childRelationships: [{
    name: 'Ingredient of (RxNorm)',
    range: [0, 999]
  }],
  parentRelationships: [{
    name: 'Has inferred drug class (OMOP)',
    range: [0, 999]
  }]
},
'RxNorm.Brand Name': {
  childRelationships: [{
    name: 'Ingredient of (RxNorm)',
    range: [0, 999]
  }],
  parentRelationships: [{
    name: 'Tradename of (RxNorm)',
    range: [0, 999]
  }]
},
'RxNorm.Branded Drug': {
  childRelationships: [{
    name: 'Consists of (RxNorm)',
    range: [0, 999]
  }],
  parentRelationships: [{
    name: 'Has ingredient (RxNorm)',
    range: [0, 999]
  }, {
    name: 'RxNorm to ATC (RxNorm)',
    range: [0, 999]
  }, {
    name: 'RxNorm to ETC (FDB)',
    range: [0, 999]
  }]
},
'RxNorm.Clinical Drug Comp': {
  childRelationships: [],
  parentRelationships: [{
    name: 'Has precise ingredient (RxNorm)',
    range: [0, 999]
  }, {
    name: 'Has ingredient (RxNorm)',
    range: [0, 999]
  }]
},
'CPT4.CPT4': {
  childRelationships: [{
[...]

PR #1179, opened 2018-12-03
Show Concept Hierarchy for all vocabs, not only pre-defined

pavgra:
Covers item 2 from https://github.com/OHDSI/Atlas/issues/1178. Related to https://github.com/OHDSI/WebAPI/issues/649


gowthamrao:
This is really good new functionality!

PR #1184, opened 2018-12-06
resolved npm warning for deprecated license metadata

### LINKED ISSUE #245: Graphical View - tooltip appearing offscreen
In the graphical view, the tooltip for the timelines can appear offscreen. Good example of this is Cohort #3079 in the J&J environment. 

image

fdefalco:
updated d3-tip 
closes #245

### DIFF
--- a/js/settings.js
+++ b/js/settings.js
@@ -1,183 +1,184 @@
 const settingsObject = {
-  baseUrl: 'js',
-  config: {
-    text: {
-      useXhr: function (url, protocol, hostname, port) {
-        return true;
-      }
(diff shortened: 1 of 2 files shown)


fdefalco:
The upgrade of the tooltip library improves the occurrence of this issue however it is still possible to see the behavior.  In my testing this only occurs if your mouse over the element from a far right position.  Otherwise the behavior is improved to this:

image

I would suggest we incorporate this change and evaluate replacing the tooltip library with the 3.0 UI overhaul.

PR #1185, opened 2018-12-07
fixed measurement frequency distribution histogram

### LINKED ISSUE #851: Achilles - measurement frequency distribution histogram - not aligned with axis
### Expected behavior
histogram should be aligned with X-axis

### Actual behavior

image

To Reproduce
go to achilles
go to measurement
select a measurement
scroll down to frequency distribution visualization

johnSamilin:
fixes #851


js/pages/data-sources/components/reports/treemapDrilldown.js
```diff
@@ -204,6 +204,7 @@ define([
 					const yScaleMax = (Math.floor((Math.max.apply(null, freqData.yNumPersons) + 5) / 10) + 1) * 10;
 					this.chartFormats.frequencyDistribution.yMax = yScaleMax;
 					this.chartFormats.frequencyDistribution.xLabel = `Count ('x' or more ${report}s)`;
+					this.chartFormats.frequencyDistribution.ticks = frequencyHistogram.INTERVALS;
```
chrisknoll: Not sure about this setting:  this is setting the number of ticks to the number of intervals.  Probably not necessary: a histogram has a continuous x-axis, unlike a barchar which has distinct intervals.  You should check how this looks, but I'd guess that we only need 5 ticks at max.

johnSamilin: @chrisknoll what if we have say, 21 interval, why do we need only 5 ticks then?

chrisknoll: Because the x-axis is scalar.  we don't need the following ticks:
1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21

It would be enough to have:
1, 6, 12, 18, 21 or whatever D3 would create out of the ticks = 5.

Why do you think you need 21 ticks for 21 intervals?

chrisknoll: To give a concrete example: here is with your PR:
image

And here is with ticks set to 5:
image

pavgra: @johnSamilin, I agree with @chrisknoll that the overlapping and messed labels in your version are not OK, however we now have another problem - it is not possible to easily understand what X value is represented by a bar:

image

@johnSamilin , could you add bot X and Y values into the tooltip (with proper captions aligning to axes names)

PR #1187, opened 2018-12-10
fixed resize issue

### LINKED ISSUE #1105: resizing window causes unconstrained chart size
### Expected behavior
Charts should have a maximum size

### Actual behavior
Charts grow unrestricted

Resize the window to narrow width and then expand, repeat.

image

johnSamilin:
fixes #1105

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "rimraf": "^2.6.2"
   },
   "dependencies": {
-    "@ohdsi/atlascharts": "^1.5.0",
+    "@ohdsi/atlascharts": "^1.6.0",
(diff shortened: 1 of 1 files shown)


pavgra:
Depends on https://github.com/OHDSI/Visualizations/pull/38

anthonysena:
Per our discussion on the ATLAS call today, we're unable to reproduce #1105 on the data sources tab so this set of changes may be unnecessary. We should re-test again on the data sources module, cohort characterization, cohort reporting, cohort generation, etc (wherever the visualization charts are invoked).

PR #1197, opened 2018-12-26
Fixed multiple count requests on concept set page

### LINKED ISSUE #1196: Included Concepts - Count queries are running many time
### Expected behavior
there should be a single request for inclusion count
### Actual behavior
`getInclusionCount` is being called multiple times
Open concept set, close it and open another one

johnSamilin:
Fixes #1196

### DIFF
--- a/js/components/conceptsetInclusionCount/conceptsetInclusionCount.js
+++ b/js/components/conceptsetInclusionCount/conceptsetInclusionCount.js
@@ -20,9 +20,10 @@ define([
 			this.inclusionCount = ko.observable(0);
 
 			this.conceptSetExpression = params.conceptSetExpression || ko.observableArray();
+			this.conceptSetSubscriptionRateLimit = params.conceptSetSubscriptionRateLimit || 1000;
 
 			this.getInclusionCount = this.getInclusionCount.bind(this);
-			this.conceptSetExpressionSub = ko.pureComputed(() => ko.toJSON(this.conceptSetExpression)).subscribe(this.getInclusionCount);
+			this.conceptSetExpressionSub = ko.pureComputed(() => ko.toJSON(this.conceptSetExpression)).extend({rateLimit: this.conceptSetSubscriptionRateLimit}).subscribe(this.getInclusionCount);
 			this.getInclusionCount();
 		}
 
--- a/js/components/conceptsetInclusionCount/conceptsetInclusionCount.js
+++ b/js/components/conceptsetInclusionCount/conceptsetInclusionCount.js
@@ -12,7 +12,7 @@ define([
 	conceptSetApi,
 ){
 
-	class ConceptSetInclusionCount extends Component{
+	class ConceptSetInclusionCount extends Component {
 
 		constructor(params) {
 			super(params);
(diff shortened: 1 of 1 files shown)


js/components/conceptsetInclusionCount/conceptsetInclusionCount.js
```diff
@@ -22,7 +22,7 @@ define([
 			this.conceptSetExpression = params.conceptSetExpression || ko.observableArray();
 
 			this.getInclusionCount = this.getInclusionCount.bind(this);
-			ko.computed(() => ko.toJSON(this.conceptSetExpression)).subscribe(this.getInclusionCount);
+			this.conceptSetExpressionSub = ko.pureComputed(() => ko.toJSON(this.conceptSetExpression)).subscribe(this.getInclusionCount);
```
chrisknoll: What would you say about adding a rate-limit to this computed, so that the subscription is only fired after a little bit of quiet?  This way when someone is rapidly adding/removing concepts or they switch around the inclusion checkboxes, we only signal the sub after a 1 second delay?  It would be done like so:

```
// Shorthand: Specify just a timeout in milliseconds
someObservableOrComputed.extend({ rateLimit: 1000 });
```

From: https://knockoutjs.com/documentation/rateLimit-observable.html

johnSamilin: What benefits will it give us?

chrisknoll: The get inclusion count (if i understand it correctly) is a pretty heavy DB operation.  Maybe if there was caching in place, this wouldn't be much of a problem.  But, the normal activity of managing a concept set is to set up a set of high-level concepts, look at the included concepts, select the ones you want to remove, then click on 'exclude', then for some of them you want to exclude their descendants, so you exclude those...  

Here's a hypothetical sequence of events for the above example without rate limiting:
1. Search for term
2. Select a high level term.
3. Go to concept set editor {call: getInclusionCount()}
4. Go to included concepts, filter on some term (like pregnancy or 'history of')
5. select the first concept you want to exclude  {call: getInclusionCount()}
6. select a second concept to exclude  {call: getInclusionCount()}
7. select a third concept to exclude  {call: getInclusionCount()}
8. switch back to main concept set. click the first concept's exclude checkbox  {call: getInclusionCount()}
9. select next one to exclude  {call: getInclusionCount()}
10. select next one to exclude  {call: getInclusionCount()}
11. select first to include descendants  {call: getInclusionCount()}
12. select next one to include descendants  {call: getInclusionCount()}

Conversely, if we rate-limit the notification to the subscription you get this
1. Search for term
2. Select a high level term.
3. Go to concept set editor {call: getInclusionCount()}
4. Go to included concepts, filter on some term (like pregnancy or 'history of')
5. select the first concept you want to exclude  {call: getInclusionCount()}
6. select a second concept to exclude  {no call if selection is made within the rateLimit}
7. select a third concept to exclude  {no call if selection is made within the rateLimit}
8. switch back to main concept set.  {call: getInclusionCount() since the last selection will now notify subs}
9. select first concept to  exclude exclude  {rateLimit begins)
10. select next one to exclude  {no call if selection is made within the rateLimit}
11. select next one to exclude  {no call if selection is made within the rateLimit}
12. select first to include descendants  {no call if selection is made within the rateLimit}
13. select next one to include descendants  {call: getInclusionCount() since the last selection will now notify subs}

Every call to getInclusionCount is a relatively expensive query to the concept_ancestor table where we need to include/exclude descendants and it's not a trivial thing.  Any way we can limit the interaction to the database would be ideal.  I think it would be hard to cache the results of the concept set expression count because the permutations of the possible selections within a concept set leads it to be something that is not easily cached.

anthonysena:
@johnSamilin - these changes are great - gives us a big gain in terms of performance which I witnessed first-hand while testing some larger concept sets that contained > 10K included concepts. In addition, I also agree with @chrisknoll's feedback to include the rate limit on the computed observable as this also improves the performance when rapidly making changes to the included/excluded options in a concept set. To that end, I added a commit to this PR and will merge it in with the rate limit. You may want to consider this change for #1198.

Issue #1169, opened 2018-11-27
Prediction covariate setting time window issue


jreps:
### Expected behavior
When you try to edit the longterm, shortterm or medium term covariate setting value, if you remove the values it goes to 0 and then it won't let you type a negative number, you have to type a positive number and then add the -.

### Actual behavior
it would be good to be able to type -365 back into long term start days without first typing 365 and then adding a - in front

PR #1207, opened 2019-01-04
Use d3.scaleThreshold() for color classification.

chrisknoll:
Fixes #732.


chrisknoll:
Also added to this PR (to keep some of the PR noise down) was fix to #1169

Issue #1206, opened 2019-01-04
Cohort warning checks piling up


anthonysena:
### Expected behavior

The "Warnings" tab on a cohort definition is evaluated when a cohort definition is active and loaded into the application. Once the cohort is closed, the warnings are no longer evaluated.

### Actual behavior

Opening multiple cohorts in sequence appears to chain together checks for warnings about cohorts that are closed. 

### Steps to reproduce behavior

With the developer console open, open a cohort, then close it. Repeat this process multiple times and you will see the following network activity:

![image](https://user-images.githubusercontent.com/12902366/50701169-c746a280-101a-11e9-9547-5e07a90a3f1e.png)

### Diagnosis

The `js/pages/cohort-definitions/components/checks/conceptset-warning.js` should extend the base `Component` class which will allow it to properly dispose. At the moment, it appears that the component is appended to the cohort-definition-manager upon opening each cohort.

Issue #1205, opened 2019-01-04
Cohort Generation - Polling Stops


anthonysena:
### Expected behavior

While a cohort is generating, the status of the job should be checked periodically (based on the configured pollingInterval) to get the status of the job. 

### Actual behavior

When navigating between cohorts during a browser session, it appears that sometimes the app loses the ability to poll for information about a cohort generation job.

### Steps to reproduce behavior

This is a bit tricky to reproduce as it does depend on the timing of the pollInterval. To attempt to do this: open a cohort and start a generation. Wait for a bit for the polling to kick in (you should see the status change from Pending to Running). Then, close the cohort definition editor. In the console, I've noticed the following error:

`cohort-definition-manager.js (377): The value of id cannot be obtained from null`

In tracing through this, the use of setTImeout to poll for the `pollInterval` appears to be the problem here. What is happening is that the polling code is re-activated even after the cohort definition editor is closed thus throwing the error above. Then after that, subsequent views of other cohort definitions appear to lack the polling due to this error. Short term, I think a quick check of the presence of the `this.model.currentCohortDefnition()` would prevent this error. Longer term, we should look at how we're disposing of the page to make sure we're cleaning up this type of polling in general.

Issue #1202, opened 2018-12-31
Concept set will not load


anthonysena:
### Expected behavior

Clicking on a concept set allows the display to load properly.

### Actual behavior

An error is thrown since `isNameCorrect` is undefined.

### Steps to reproduce behavior

Using code at https://github.com/OHDSI/Atlas/commit/c640dffa6d5fb08773667d80ebf091f381987fa1, attempt to open a concept set.

### Debugging information

Tracing this, `js/pages/concept-sets/conceptset-manager.js` the computed property `canSave` relies on another computed variable `isNameCorrect` which is declared after `canSave` thus leading to the error. The solution is to move the property declaration `isNameCorrect` ahead of `canSave`.

Issue #1201, opened 2018-12-31
Related source codes not loading


anthonysena:
### Expected behavior

When reviewing a concept set, the related source code tab displays the related source codes for those concepts specified in the concept set.

### Actual behavior

An error is thrown since the sharedState variable is undefined.

### Steps to reproduce behavior

Using code at https://github.com/OHDSI/Atlas/commit/c640dffa6d5fb08773667d80ebf091f381987fa1, open a concept set and navigate to the source codes and in the developer console you will observe an error stating that sharedState is undefined.

### Debugging information

Tracing this, `js/const.js` has a circular dependency to `js/components/atlas-state.js` so the fix is to remove the require for atlas-state in const.js.

PR #1208, opened 2019-01-07
Various issues fixed

anthonysena:
This PR addresses a number of bugs as described in each commit message. Specifically:

#1200 - User bar permission warning
#1201 - Remove circular dependency for const.js
#1202 - Concept Set will not load
#1205 - Cohort generation polling
#1206 - Dispose of cohort warning subscription

Also, there is a commit for fixes related to negative controls which I did not log but am happy to do that if needed.

### DIFF
--- a/js/Model.js
+++ b/js/Model.js
@@ -55,7 +55,7 @@ define(
 				this.relatedConceptsOptions = constants.relatedConceptsOptions;
 				this.relatedSourcecodesOptions = constants.relatedSourcecodesOptions;
 				this.metatrix = constants.metatrix;
-				this.relatedSourcecodesColumns = constants.getRelatedSourcecodesColumns(this);
+				this.relatedSourcecodesColumns = constants.getRelatedSourcecodesColumns(sharedState, this);
 				this.enableRecordCounts = ko.observable(true);
 				this.loading = ko.observable(false);
 				this.loadingIncluded = ko.observable(false);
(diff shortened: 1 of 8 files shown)


chrisknoll:
> Also, there is a commit for fixes related to negative controls which I did not log but am happy to do that if needed.

Hi, @anthonysena, could you just summarize what should be checked for the fix for negative controls?

anthonysena:
Sure - looking at https://github.com/OHDSI/Atlas/pull/1208/commits/6e7501acf27057d06808d9612bfa2b42df494508, there were two changes I needed to make:

1. At some point, the JobDetailsService was replaced by the jobDetail class in `js/components/evidence/components/negative-controls.js` which caused the job polling to fail. This has been changed back and is functional.
2. `js/services/CDMResultsAPI.js` was refactored to remove some redundancy but the arguments passed into the `getConceptRecordCount` function were not being passed properly to the `getConceptRecordCountWithResultsUrl` function which created a bug when you want to toggle record counts between data sources in the candidate negative controls results grid.

Hope this helps!

chrisknoll:
@anthonysena : I am seeing the userbar warning once on startup, tho navigating to the cohort definitions does not seem to produce additional warnings.  Is the fix to completely eliminate the message, or is the message about permissions valid if it only appears once?

image

anthonysena:
Thanks @chrisknoll for the review! To answer your question: the change I made related to the permission warning was to only show it a single time whereas the code was writing the warning out to the console for each poll to the server thus repeating the same warning over and over.

PR #1211, opened 2019-01-09
Modified source diamon name Evidence to CEM

NEONKID:
Fixed #1210 
I think Evidence is now CEM.


NEONKID:
The same problem has been fixed in addition to the Priority setting section.

anthonysena:
Thanks @NEONKID for these fixes. @pavgra could you confirm that these are working for you since they involve security & configuration? Thanks!

pavgra:
@anthonysena , thank you for pinging me - it appeared that the method lacked security checks at all. I've logged the issue:
https://github.com/OHDSI/WebAPI/issues/743

PR #1223, opened 2019-01-15
Fixed load cc caption bug and pathways..

### LINKED ISSUE #1214: Error reading characterization caption
screenshot from 2019-01-10 17-48-05

When I select Characterization Caption and select another menu, I need to call up the previous Caption.

NEONKID:
Fixed #1214 

The function to be called is located below the called function.
In addition to CC (Characterizations), the same bugs in Pathways have been fixed at the same time.

### DIFF
--- a/js/pages/pathways/components/manager.js
+++ b/js/pages/pathways/components/manager.js
@@ -40,8 +40,12 @@ define([
 			this.dirtyFlag = sharedState.CohortPathways.dirtyFlag;
 			this.analysisId = ko.observable();
 			this.executionId = ko.observable();
-
 			this.loading = ko.observable(false);
+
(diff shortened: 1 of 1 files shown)


anthonysena:
@NEONKID can you merge the current master into your forked branch? #1224 appears to address the same issue for cohort characterization which may cause a conflict. The addition you made for pathways is appreciated and we'd like to get that merged in as well. Thanks!

NEONKID:
@anthonysena Okay. I have pushed my commit back to the current master branch. Please confirm. Thank you.

chrisknoll:
Ok, cool, this looks better: now we only see pathways/components/manager.js needing changes, while the prior version of the PR had others.  Looks better!

Issue #1229, opened 2019-01-16
Optimize concept set fails


anthonysena:
### Expected behavior

When viewing a concept set, clicking the "Optimize" button will attempt to find an optimized version of the concept set expression that removes any potential redundancies.

### Actual behavior

The current code on master throws the following error:

![image](https://user-images.githubusercontent.com/12902366/51256277-0659e980-1973-11e9-9774-eba5330e5977.png)

### Debugging

The "defaultSource" in the VocabularyService is not set properly somewhere thus the error is thrown. Likely we should change this so that the caller sends in the default source instead.

Issue #1228, opened 2019-01-16
Import concept set in cohort definition throws error


anthonysena:
### Expected behavior

When creating a cohort, importing a concept set for use in defining an event should not throw a JavaScript error.

### Actual behavior

We get the following error:

![image](https://user-images.githubusercontent.com/12902366/51252335-8b400580-1969-11e9-9cbc-4116aea1c49b.png)

### Steps to reproduce behavior

 Described above.

Issue #1133, opened 2018-11-16
5)	Clicking the "save new concept set from selection below" button in concept set compare tab creates odd concept set


jswerdel:
### Steps to reproduce behavior
a)	Open a concept set
b)	Click compare
c)	Add another concept set 
d)	Click save new concept set from selection below button

jswerdel:
![image](https://user-images.githubusercontent.com/19151846/48638998-39e2bc80-e9a0-11e8-82a4-532fa380735d.png)

Issue #728, opened 2018-05-29
Distance in hierarchy does not match value from concept_ancestor


chrisknoll:
### Expected behavior
Performing the query on concept_ancestor should match the results seen in Atlas.

### Actual behavior
min_levels_of_separation is 0 in the SQL query, but distance is 1 in Atlas.

### Steps to reproduce behavior
This query shows the min distance from concept ancestor from Hysterectomy to a CPT 4 procedure:

```
select * from concept_ancestor where ancestor_concept_id = 4127886 and descendant_concept_id = ```
2109161
```

However, in atlas, looking at the hierarchy, we get a distance of 1 at both the direct concept level (which I'd expect to be 0 distance) and the parents (which I'd expect 1, but the SQL query shows a distance of 0).

![distancepng](https://user-images.githubusercontent.com/6818777/40677013-69f8a808-634a-11e8-8a2f-752b3ad34488.png)

PR #1233, opened 2019-01-17
Concept set fixes

anthonysena:
This PR aims to address a number of concept set/vocabulary bugs:

#282 - Concept Set Comparison Caching Definition
#568 - Multiple resolveConceptSetExpressions result in race condition
#620 - Optimizing a null conceptset causes issue
#702 - Trashing an unsaved concept set
#728 - Distance in hierarchy does not match value from concept_ancestor
#1133 - Clicking the "save new concept set from selection below" button in concept set compare tab creates odd concept set
#1228 - Import concept set in cohort definition throws error
#1229 - Optimize concept set fails

The commits in this PR are organized by the issues above. Please let me know if you have any questions or concerns with these fixes.

### DIFF
--- a/js/Model.js
+++ b/js/Model.js
@@ -366,8 +366,7 @@ define(
 			// for the current selected concepts:
 			// update the export panel
 			// resolve the included concepts and update the include concept set identifier list
-			resolveConceptSetExpression() {
-				this.resolvingConceptSetExpression(true);
+			resolveConceptSetExpression(resolveAgainstServer = true) {
         this.includedConcepts.removeAll();
         this.includedSourcecodes.removeAll();
 				var conceptSetExpression = { "items": sharedState.selectedConcepts() };
@@ -379,7 +378,7 @@ define(
 				}
 				this.currentConceptIdentifierList(conceptIdentifierList.join(','));
 
-				return this.resolveConceptSetExpressionSimple(conceptSetExpression);
+				return resolveAgainstServer ? this.resolveConceptSetExpressionSimple(conceptSetExpression) : null;
 			}
 
 			resolveConceptSetExpressionSimple(expression, success) {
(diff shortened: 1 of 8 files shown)


chrisknoll:
### #282 

looks good, i checked the POST'd payload to have the updated concept set definition.

### #568 
still problematic. When I switch between concept sets, there is a `resolveConceptSetExpression` invoked twice:
image

I'm showing you the one where the request is the `items: []` payload, but you can see the subsequent `resolveConceptSetExpression` is invoked again.  it appears that if you pass `false` to resolve, it is supposed to bypass a call, but there seems to be some AJAX request:
``
return resolveAgainstServer ? this.resolveConceptSetExpressionSimple(conceptSetExpression) : null;
``

I didn't trace too deeply into resolveCocneptSetExpressionSimple() to know if that's where the AJAX call is coming from.

### #620 
I'm running off of latest master, and getting an error when optimizing an empty concept set ( a concept set with no items in it):

image

### #702 
This looks resolved, by nature of you don't get the 'trash' icon on an unsaved-new concept set.

### #728 
Looks good, distance is reporting as 0 in the selected concept.  It appears, however if there is a MedDRA PT for the concept it will have a distance of 0 but appear as a parent. I'm thinking this is just a nuance of the OMOP Vocabulary, that you can have a parent with a distance of 0.

### #1133 
This no longer results in the '0 concept set' result. 

### #1228 
This imports (overwrites) a cohort definition's concept set without error.

### #1229 
Optimizing no longer results in a javascript error.

chrisknoll:
Yes, I created an empty concept set, saved, and then optimized.  The optimize button is hidden until saved, so i'm unable to optimize an unsaved concept set.

anthonysena:
Thanks @chrisknoll for confirming - I believe I've addressed the issue with #620 with this commit: https://github.com/OHDSI/Atlas/pull/1233/commits/3b6ce673af89515af67eb99324bef558a7f7b7c1 

Also, for #568, just curious if you are switching between concept sets inside of a cohort definition? I'm unable to reproduce the behavior you showed above so I think I'm not testing it the same way you did.

anthonysena:
@chrisknoll - traced through the concept set resolve item and found that there was a redundant call made from the router - the appModel.loadConceptSet eventually resolves the concept set so I'm not sure why the other call was required. I've removed it as part of this commit: https://github.com/OHDSI/Atlas/pull/1233/commits/6b5228dae039e6864b8584a5cf7549e795add32f. Let me know if you know if you have any other questions/concerns

PR #1235, opened 2019-01-18
Fix inclusion impact report where censor period is being used

wivern:
Fixes OHDSI/WebAPI#364


chrisknoll:
This new column 'lost to censoring' is also visible in `FeasiblityAttritionReport.html`, so I think that needs to be added there as well.

PR #1236, opened 2019-01-18
Used DS with result & vocabulary daimons for Data Source tab

mpozhidaeva:
Fixes https://github.com/OHDSI/WebAPI/issues/720

### DIFF
--- a/js/utils/CommonUtils.js
+++ b/js/utils/CommonUtils.js
@@ -85,7 +85,11 @@ define([
 		return source.daimons.find(daimon => daimon.daimonType == 'Results') !== undefined;
 	}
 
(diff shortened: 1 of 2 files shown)


js/utils/CommonUtils.js
```diff
@@ -85,7 +85,11 @@ define([
 		return source.daimons.find(daimon => daimon.daimonType == 'Results') !== undefined;
 	}
 
-	function renderConceptSetItemSelector(s, p, d) {
+	function hasVocabulary(source) {
```
pavgra: Where is the new function used?

mpozhidaeva: oops, my bad - removed. thanks

PR #1245, opened 2019-01-22
fixed Confusing Included Concepts count

johnSamilin:
fixes https://github.com/OHDSI/WebAPI/issues/737

### DIFF
--- a/js/Model.js
+++ b/js/Model.js
@@ -392,12 +392,16 @@ define(
 			resolveConceptSetExpressionSimple(expression, success) {
 				const callback = typeof success === 'function'
 					? success
-					: ({ data: info }) => {
-							this.conceptSetInclusionIdentifiers(info);
-							this.currentIncludedConceptIdentifierList(info.join(','));
-							this.conceptSetInclusionCount(info.length);
-							this.resolvingConceptSetExpression(false);
-						};
+					: ({ data }) => {
+						let info = data;
+						if (!Array.isArray(info)) {
+							info = [];
+						}
+						this.conceptSetInclusionIdentifiers(info);
+						this.currentIncludedConceptIdentifierList(info.join(','));
+						this.conceptSetInclusionCount(info.length);
+						this.resolvingConceptSetExpression(false);
+					};
 				const resolvingPromise = httpService.doPost(sharedState.vocabularyUrl() + 'resolveConceptSetExpression', expression);
 				resolvingPromise.then(callback);
 				resolvingPromise.catch((err) => {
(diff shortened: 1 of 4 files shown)


js/components/conceptsetInclusionCount/conceptsetInclusionCount.js
```diff
@@ -30,7 +30,7 @@ define([
 		getInclusionCount() {
 			this.countLoading(true);
 			conceptSetApi.getInclusionCount(this.conceptSetExpression())
-				.then(({data}) => this.inclusionCount(data))
+				.then(({data}) => this.inclusionCount(Number.isInteger(data) ? data : 0))
```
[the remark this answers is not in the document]
pavgra: Let's show an explicit message as @mpozhidaeva suggests (e.g. "Cannot load Included concepts due to problems with Vocabulary daimon")

pavgra:
@johnSamilin , changed the vocabulary schema to incorrect one as described in the issue and cannot even reach the page:

image

johnSamilin:
@pavgra could you check again?

pavgra:
@johnSamilin , from the Concept Sets section point-of-view looks good. However, we also have concept sets embedded into Cohort Definitions. Would be great to handle such case there also

image

js/pages/concept-sets/components/tabs/included-conceptsets.html
```diff
@@ -16,7 +16,10 @@
 <!-- ko if: $root.loadingIncluded -->
   <loading></loading>
 <!-- /ko -->
-<!-- ko ifnot: $root.loadingIncluded -->
+<!-- ko if: $component.hasErrors -->
+  <empty-state params="message: 'Could not load included concept sets'"></empty-state>
```
pavgra: With the `alert` approach, is there a need in the change? In which case are we going to reach the view?

js/Model.js
```diff
@@ -399,20 +399,18 @@ define(
...
+							throw new Error();
+						}
+						this.conceptSetInclusionIdentifiers(info);
+						this.currentIncludedConceptIdentifierList(info.join(','));
+						this.conceptSetInclusionCount(info.length);
+					};
 				this.resolvingConceptSetExpression(true);
-				const resolvingPromise = httpService.doPost(sharedState.vocabularyUrl() + 'resolveConceptSetExpression', expression);
-				resolvingPromise.then(callback);
-				resolvingPromise.catch((err) => {
```
pavgra: What about direct calls of the method? e.g. from `Application.js`. Why don't you leave the `catch`?

PR #1267, opened 2019-01-24
Pathways - remove observable from status property

anthonysena:
@pavgra While testing #1252, I was running into an issue when loading up executions for pathways:

image

It is unrelated to related to changes #1252 but since I came across it, I wanted to see if you are getting anything similar? If so, let's either bring it into your branch or master.


anthonysena:
Nevermind - I see this is covered in #1251

PR #1275, opened 2019-01-29
Make visibility.core.js AMD compatable.

chrisknoll:
Removed visibility.core from package.json in order to change the implementation.

### DIFF
--- a/js/services/Poll.js
+++ b/js/services/Poll.js
@@ -1,4 +1,4 @@
-define(['knockout', 'visibilityjs'], (ko) => {
+define(['knockout', 'visibilityjs'], (ko, Visibility) => {
   const callbacks = new Map();
   
   const isPageForeground = ko.observable(Visibility.state() === "visible");
--- a/js/settings.js
+++ b/js/settings.js
@@ -132,7 +132,7 @@ const settingsObject = {
 		"colorbrewer": "../node_modules/colorbrewer/index",
 		"ohdsi-api": "../node_modules/@ohdsi/ui-toolbox/lib/umd/api/index",
 		"bootstrap-datetimepicker": "../node_modules/bootstrap-datetimepicker/src/js/bootstrap-datetimepicker",
-		"visibilityjs": "../node_modules/visibilityjs/lib/visibility.core"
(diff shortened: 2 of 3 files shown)


chrisknoll:
Btw, in order to clean out old package reference to visibility.js, you will need to run `npm install` again.

chrisknoll:
Sounds complicated.  Who is going to own the repo?  Who's going to publish our fork to NPM?  This is just a short term solution, I'll be submitting a PR to the upstream repo, and I figure it will take them a few weeks to notice it, maybe accept it, and then we can go back to using the NPM managed lib.

But, if you think it's worth the extra hassle, go ahead and we'll update our package.json to the new library.

chrisknoll:
I've opened a PR here https://github.com/ai/visibilityjs/pull/48 but initial feedback does not sound too promising.  Fingers crossed!

anthonysena:
@chrisknoll - thanks for putting this together and from my review & testing, it is working well in both bundled and non-bundled mode. 

I'm hopeful that the folks that manage the visibilityjs repo will include these changes thus making it less code for us to manage as a community. If not, perhaps we can fork the repo as @pavgra into an OHDSI repo for this specific module? Seems like overkill but then perhaps it can be managed in a similar fashion to the atlas charting library?

chrisknoll:
Update, the visibilityjs maintainers, in their infinite wisdom, have decided to reject the PR since it does not follow the model that they deem 'the best':

> This is why I prefer to not add AMD to force any AMD user to migrate to good builders with the latest practices.

So, we'll have to work from a fork (tho, after that experience, I'd rather we drop this library completely on principle).  I'm going to publish the library with AMD support under the `@ohdsi` namespace and will modify the PR to point to the correct version.

chrisknoll:
@anthonysena , i've pushed a forked version of the library to include AMD support. If you have already pulled this branch, you might need to delete your local branch before you pull again (otherwise strange merges may happen). Sorry, i was just keeping the PR commits clean.

chrisknoll:
I made the update in my fork here: https://github.com/chrisknoll/visibilityjs.  The package.json file that was published to npm refrences my fork (not the origional).

chrisknoll:
@pavgra , please approve or dismiss your review. Then we can merge it in. Thanks!

PR #1279, opened 2019-01-31
Added message about not found entities

mpozhidaeva:
Fixes https://github.com/OHDSI/WebAPI/issues/8


pavgra:
@mpozhidaeva , UiToolbox doesn't export STATUS:
https://github.com/OHDSI/UiToolbox/blob/master/src/index.js
image

Dear @johnSamilin , how do you approve PR w/o testing it?

pavgra:
@mpozhidaeva , current approach with a redirect to separate 404 page is not very UX friendly: hard to go back through browsing history - Atlas continues to push you to the 404 page. Therefore, I propose to avoid separate page and use an approach similar to the initial loader one (`<div data-bind="if: !currentViewAccessible()">`)

mpozhidaeva:
@pavgra - ok, thanks - I can try

Issue #1099, opened 2018-11-07
UI / UX issues at IR generation page


pavgra:
Following issues were reported at some clients sites:
1. Opening generation tab with existing stats takes too much time
1. When clicking generate button - the response lag time is very big, like 5 -10 seconds - before the button changes to running
1. Generation tab UI is messy:
- Keeps jumping during IR generation
- Delete buttons are not aligned
- Results are not displayed even though data is in place

PR #1292, opened 2019-02-02
UI/UX fixes for IR

pavgra:
fixes https://github.com/OHDSI/Atlas/issues/1099

- made IR generations page to be similarly styled as Cohorts one (to improve UI consistency across the app and bring better UX patterns from Cohorts page to IR)
- removed delay between Generation button click and loader being displayed

### DIFF
--- a/js/pages/incidence-rates/components/iranalysis/components/results.html
+++ b/js/pages/incidence-rates/components/iranalysis/components/results.html
@@ -47,7 +47,7 @@
 			<th data-bind="css: classes('tbl-th', 'duration', 'smallHeading dataCol')">
 				Duration
 			</th>
-			<th data-bind="css: classes('view-results')"></th>
+			<th data-bind="css: classes('tbl-th', 'view-results')"></th>
 		</thead>
 		<tbody data-bind="foreach: sources().filter(function(source) { return source.info() != null; })">
 			<tr>
--- a/js/pages/incidence-rates/components/iranalysis/components/results.html
+++ b/js/pages/incidence-rates/components/iranalysis/components/results.html
@@ -13,6 +13,10 @@
 		</select>
 	</div>
 	<div>
+		<select-sources-btn
+			data-bind="css: classes('generate-btn')"
+			params="sources: $component.generationSources, callback: $component.runGenerations"
+		></select-sources-btn>
 		<button type="button" class="btn btn-sm btn-success pull-right exportAnalysisCSV" data-bind="css: {'disabled': dirtyFlag().isDirty, 'btn-success': !dirtyFlag().isDirty()}"><i class="fa fa-download" aria-hidden="true"></i> Export Analysis to CSV</button>
 	</div>
 </div>
@@ -49,10 +53,7 @@
 			<tr>
 				<td data-bind="css: $component.classes('tbl-col')" nowrap>
 					<!-- ko if: info().executionInfo.status == "COMPLETE" -->
-						<button class="btn btn-sm btn-primary execute">
-							<i class="fa fa-play"></i>
-							Generate
-						</button>
+						<select-sources-btn params="sources: () => [source], callback: $component.runGenerations, wasGenerated: true, label: 'Rerun'"></select-sources-btn>
 					<!-- /ko -->
 				</td>
 				<td data-bind="css: $component.classes('tbl-col'), text: source.sourceName"></td>
--- a/js/pages/incidence-rates/components/iranalysis/components/results.html
+++ b/js/pages/incidence-rates/components/iranalysis/components/results.html
@@ -15,7 +15,7 @@
 	<div>
 		<select-sources-btn
 			data-bind="css: classes('generate-btn')"
-			params="sources: $component.generationSources, callback: $component.runGenerations"
+			params="sources: $component.generationSources, callback: $component.runGenerations, disabled: $component.isExecutionDisabled, disabledReason: $component.executionDisabledReason"
 		></select-sources-btn>
 		<button type="button" class="btn btn-sm btn-success pull-right exportAnalysisCSV" data-bind="css: {'disabled': dirtyFlag().isDirty, 'btn-success': !dirtyFlag().isDirty()}"><i class="fa fa-download" aria-hidden="true"></i> Export Analysis to CSV</button>
 	</div>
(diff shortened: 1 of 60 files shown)


chrisknoll:
Sorry we didn't discuss this first.  I haven't reviewed this PR request so I do not know what you mean by 'made consistent', but if we're going to change any of the generate behavior, here's my suggestion:

First, main problem with current implementation:  when you have many sources, the results screen that shows sources with results is intermingled with sources that do not have results. We have about 20 CDM sources set up in our env, and it's very cumbersome to find a source with results, and generating across multiple sources is also a bit cumbersome.

So, in form-follows-function principle, this is how I would define the generate function requirements:

1. Only sources with results should be displayed.
2. You should be able to batch-execute multiple sources by selecting the sources, then generate.

Item 1 is easy: just show sources that have results. That will eliminate  the problem of figuring out which sources have results, and reduce a lot of clutter.

Item 2 is also fairly simple.  Given the current IR layout:

image


You can see the Generate UI element at the top. Currently it is a drop-down that contains the set of sources (not expanded here) but, instead, the generate dropdown can instead be a buton that displays a modal of available sources.  Using a 'click-to-select' checkbox or row select, you can select the set of sources you want to generate on.  After selections are made, you click 'generate' to close the modal and kick off the jobs (or you can cancel if you change your mind). This has the added benefit of not accidentally 'generating' on a source when you didn't intend to (although I understand you can cancel the jobs, it's better not to make the mistake int he first place).

The IR UX experiene was the latest attempt at a streamlined cohort generation process. We never had an opportunity to migrate it to the other sections of the app, and since you have the time to update the generate UX, let's move forward by applying a UX that gives the above advantages instead of moving away from it and making the UX less optimal.

I'm happy to take this on, if you find that you do not have the cycles to do this. but I'm pretty firm against taking a step backwards in UX.

pavgra:
@chrisknoll , 

> I haven't reviewed this PR request so I do not know what you mean

Doesn't seem like a constructive approach when it is followed not by some questions but by a waterfall of some assumptions. Anyway, I returned back the existed behavior and going to migrate the same approach to cohorts (if it doesn't occur that the required amount of time is infeasible)

pavgra:
image

image

image

image

chrisknoll:
@pavgra ,
 I ran it locally, the changes look good, there is one strange behavior with the UI layout related to resizin the window.

At this size there's no horizontal scrollbar:
image

But at this resolution:
image

the horizontal bar appears.  Oddly, making the window smaller does cause the table to shrink down:
image

Is this something easily addressed?

chrisknoll:
Update:
image

Seems like the issue stems from the view report button overflowing the table cell but the overall table is not recognizing this and instead of limiting the column size, it just lets the column shrink

pavgra:
@chrisknoll , hopefully, columns are better now:

image

image

chrisknoll:
Scrollbar is gone, but we get this button wrapping when the window is still fairly wide. I'm gonig to try to adjust the style to apply either a no-wrap or fixed width on the column.  here's what i see:
image

I'll push up a change in a moment, uless you have a different suggestion.

chrisknoll:
Ok this si what I came up with finally:
image


I tried to work with relative sizes and using 'em' for the width specification, but the only way I could lock it in is to use 'px' units (which I think you would disapprove of?)  Anyways, I pushed and you can see for yourself.

pavgra:
@chrisknoll , if it works - good. I will push changes if smth better comes into my mind.

What do you think of such Select sources popup?
image

chrisknoll:
Looks very good. Way nicer.  I'm concerned about the select all, but I can worry about that later. What we've seen on our env is that sometimes people like to see cohorts and CC's run across all the souces, which then crushes our databases...so a select all makes it easier to do multi selects but in our env, it's highly unlikely anyone should want to cick taht.

chrisknoll:
When using the modal and making selections (in IR), get these errors when clicking on checkboxes (and I do not think this was happening before):

image

chrisknoll:
I believe the issue is how the `sources` computed is being used in the datatable.

```
this.sources = ko.computed(() => params.sources().map(s => ({ ...s, 
  selected: this.selectedSources().includes(s.key) })));
```
According to the `Renderers.renderCheckbox` function:

```
	function renderCheckbox(field) {
		return '<span data-bind="click: function(d) { d.' + field + '(!d.' + field + '()); } , css: { selected: ' + field + '}" class="fa fa-check"></span>';
	}
```

it is assuming that the 'selected' field in the collection is an observable, that when clicked will toggle the observable's value to true/false. 

I'm making changes now...as i'm looking deeper, there's some confusion about where there's a field 'selected' to maintain the selections, vs a list of 'selected' sources...so I hope to resolve all this shortly, and I will push up some changes.

chrisknoll:
Ok, I made some changes, and hope this follows your mental model.  At one point, I was confused a bit as to why the select-sources-btn was being put in each of the results rows, since in the results rows, is the result of a single source, and so no source selection would be necessary in this case (it would just call the execute method for the source).

But, I worked through it, and while I understood the idea of reusing a single 'generate' button for the single-source and multi-source case, i don't think I would have added this complexity into the component: I think where we just know we have a single source and we want to 're-run', I'd just have put a button to re-run that calls execute().  This would remove the need for a 'wasGenerated' parameter (which, the purpose of the component is to push a button, show a list of sources, and then take action for each selected source...it doesn't have to be 'generate').   And then also remove the logic within the 'source selector' where if you have just a single source, it should automatically call the generate method.  I think this would simplify the component into having a single purpose, but I acknowledge I may not see the bigger picture to your overall design.

If you could, can you please pull this branch down and run through your scenerios and make sure I didn't break some edge-case that I didn't anticipate?  If you are good, then we'll merge. I will apply my approval now, and you can decide how to proceed.

js/pages/incidence-rates/components/iranalysis/components/results.html
```diff
@@ -1,66 +1,105 @@
...
+						<button class="btn btn-sm btn-primary" data-bind="visible: info().executionInfo.isValid">
+							<i class="fa fa-eye"></i> Reports
+						</button>
+						<button class="btn btn-sm btn-danger removeResult">
+                            <i class="fa fa-trash"></i>
+                        </button>
+					</td>
 				<!-- /ko -->
-				<!-- ko ifnot: info().executionInfo.isValid -->
-				<td colspan="4"><div class="overflow-ellipsis" style="width:345px" data-bind="text: 'Error: ' + info().executionInfo.message, attr:{title: info().executionInfo.message}"></div></td>
```
chrisknoll: We removed the columns that store the execution error message, is this no longer being displayed in this revision?

pavgra:
@chrisknoll , I've re-pushed fix for console exceptions since my diff is smaller in times. I will integrate the job cancel execution capabilities now (as back-end was merged earlier today). I will also look into the single button usage question - I agree that the complexity is higher and may be I went too far

pavgra:
- Added ability to cancel IR:

image

- Added ability to open popup with the error message if execution failed:

image

requires https://github.com/OHDSI/WebAPI/pull/809 to work properly

pavgra:
@chrisknoll , thank you for the help. Do we think we are ready for the final merge now?

pavgra:
@chrisknoll , would you mind approving the related WebAPI PR?
https://github.com/OHDSI/WebAPI/pull/809

Issue #1278, opened 2019-01-30
Clean profile page before searching


mpozhidaeva:
### Expected behavior
Empty panel with just "Can't find -9"
### Actual behavior
Panel with previous results + "Can't find -9"
### Steps to reproduce behavior
Open the Profile tab -> Choose data source + type '12' to find the profile -> type '-19'
![image](https://user-images.githubusercontent.com/31308645/51977497-aeee6a00-2498-11e9-9626-feac34e9cef1.png)

PR #1293, opened 2019-02-04
Added conditions to hide profile blocks

aklochkova:
Fixes https://github.com/OHDSI/Atlas/issues/1278


pavgra:
Why did you need to rename the variable? Especially, when the new name semantically doesn't reflect real situation: you set `personIsFound` to `true` _before_ you loaded the info, so _before_ you actually now whether the person exists. So I would suggest to leave the variable name as it was and only add the required condition

PR #1299, opened 2019-02-05
Use lodash-amd library.

### LINKED ISSUE #1298: Import lodash-amd v4.71.11
### Expected behavior
Bundling the lodash library should only incur the cost of the modules that are referenced, not an entire library.

### Actual behavior
The lodash library imported via `"lodash": "^4.17.11",` in package.json leads to a very large bundle by importing the main lodash library (which is about 500k)

Disable cache and open your console and load the home page. You will get the lodash file downloaded:

image

### Proposed Solution

There is a package for lodash combiled with AMD modules.  In spite of what @ai believes, AMD support is still alive and kicking, and we can benefit from this library to only include dependencies that we actually require.

Changing package.json to have `"lodash-amd": "^4.17.11",` and updating the packages config to point to the new reference, we can use the specific lodash moduels via:

`define(['lodash/chunk'],(chunk) -> {  ... use chunk ... });`

chrisknoll:
Fixes #1298

### DIFF
--- a/js/assets/lodash-slim.js
+++ b/js/assets/lodash-slim.js
@@ -1,0 +1,77 @@
+define((require, exports) => {
+	const bind = require('lodash-amd/bind');
+	const chain = require('lodash-amd/chain');
+	const cloneDeep = require('lodash-amd/cloneDeep');
+	const curry = require('lodash-amd/curry');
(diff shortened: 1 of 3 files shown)


anthonysena:
Moving to 3.0 for consideration in that effort

PR #1323, opened 2019-02-11
Make source selection always reference the objects to select instead of returning keys.

chrisknoll:
When working through the changes for #1292, it was difficult to follow how the set of source choices was being transformed between the IR UI -> generate button -> select-source-popup -> IR execution callback.  In the process of several transforms, the starting point of a set of sources gets resolved to a set of keys.  While tracing through it, I was imaging some questions that might be reaised by someone who is not familar with the code:

How does the consumer know that the set of sources I pass in comes back as an array of strings that represent keys?
If i want to reuse some of this, or generalize it, does it require that an object being chosen have a 'key' property?

To simplify things, this PR changes it so that when you pass the set of sources as your 'set of options/choices', the call back will return the very same objects that you sent in, but only those that were selected by the user.

This has a nice benefit of if you want to make the button execute on a default source (without prompting the user to choose) you can just specify the 'selectedSources' as an array of a single source (you don't even need to pass in a list of sourceOptions) and clicking the button will execute directly on that source.  

In addition, this PR fixes the broken Select All and Deselect All behavior.

### DIFF
--- a/js/components/generation/select-sources-btn.html
+++ b/js/components/generation/select-sources-btn.html
@@ -8,7 +8,7 @@
 <!-- ko if: $component.shouldSuggestSelection() -->
     <select-sources-popup params="{
 		showModal: $component.isPopupShown,
-		sources: $component.sources,
+		sources: $component.sourceOptions,
 		selectedSources: $component.selectedSources,
 		submit: $component.generate
 	}"></select-sources-popup>
--- a/js/components/generation/select-sources-btn.html
+++ b/js/components/generation/select-sources-btn.html
@@ -8,7 +8,7 @@
 <!-- ko if: $component.shouldSuggestSelection() -->
     <select-sources-popup params="{
 		showModal: $component.isPopupShown,
-		sources: $component.sourceOptions,
+		sources: $component.sources,
 		selectedSources: $component.selectedSources,
 		submit: $component.generate
 	}"></select-sources-popup>
--- a/js/components/generation/select-sources-btn.js
+++ b/js/components/generation/select-sources-btn.js
@@ -23,16 +23,11 @@ define([
 			this.label = typeof params.label === 'undefined' ? 'Generate' : params.label;
 			this.wasGenerated = typeof params.wasGenerated === 'undefined' ? false : params.wasGenerated;
 
-			this.sourceOptions = ko.computed(() => (ko.utils.unwrapObservable(params.sources) || []).map(s => ({
-				name: s.sourceName,
-				disabled: typeof s.disabled !== 'undefined' ? s.disabled : false,
-				disabledReason: s.disabledReason,
-				source: s
-			})));
+			this.sources = params.sources || ko.observableArray();
 			this.selectedSources = params.selectedSources || ko.observableArray();
 			this.callback = params.callback;
 
-			this.shouldSuggestSelection = ko.computed(() => this.sourceOptions().length > 1);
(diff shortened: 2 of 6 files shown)


js/components/generation/select-sources-btn.js
```diff
@@ -23,17 +23,16 @@ define([
 			this.label = typeof params.label === 'undefined' ? 'Generate' : params.label;
 			this.wasGenerated = typeof params.wasGenerated === 'undefined' ? false : params.wasGenerated;
 
-			this.sources = ko.computed(() => params.sources().map(s => ({
+			this.sourceOptions = ko.computed(() => (ko.utils.unwrapObservable(params.sources) || []).map(s => ({
```
pavgra: If we pass sources to the popup anyway, do we really need this `map`? - it actually doesn't do anything useful then

js/components/generation/select-sources-btn.html
```diff
@@ -8,7 +8,7 @@
 <!-- ko if: $component.shouldSuggestSelection() -->
     <select-sources-popup params="{
 		showModal: $component.isPopupShown,
-		sources: $component.sources,
+		sources: $component.sourceOptions,
```
pavgra: I would either leave the prev name or rename both - so to keep things consistent

chrisknoll: I'll rename to sourceOptions.

js/components/generation/select-sources-btn.js
chrisknoll: From what I recall, the map is projecting a name and disabled field from the underlying source object, so the source is becoming something new (which is why I added the 'source' field to keep track of the origional, and called this collecton 'sourceOptions' instead of leaving it as 'source' because the param `source` and the mapped `source` were different things). But if the source object can be passed in 'as-is' and we modify the popup logic to read values directly from the source object, then we may not need this map.  However, this also is here to handle the case where the component did not receive an array of sources at all (you can make it work with just passing 'selectedSource').  

So if you want to try to remove this `map()`, where do you see the impact on subsequent code (the popup, the toggles, etc)?

pavgra: > where do you see the impact on subsequent code (the popup, the toggles, etc)?

As you truly mentioned - only in those places which rely on `name` (so we'll have to switch those to `sourceName`) and `disabled` (move the undefined check) + update your new code from using nested `sourceOption.source` to the `source`. But this will allow to remove a level of data nesting and processing

chrisknoll: Ok, let me work on it.

js/components/generation/select-sources-btn.html
chrisknoll: Ok, once I stripped out that map() call, it made sense to keep those names consistent (so the param is `source` and the field is source.  You'll notice that i still kept a 'sourceOptions' field, tho, because we are still doing a transformation between the provided `sources` into a structure that contains a field (selected) that indicates of the user has selected the option or not.  Hopefully, tho, you'll see how the context of the object changes from source-> sourceOption and then the selections are returned as the array of sources (that i the exact same object that was initially passed in via params).

This may not be 100% clean but I think it's easier to follow this way. And, I think we can have a generic 'entity selector' component where you just need to provide the array of objects and the column definitions on how to read the values from the object and display as a column, and then you can have differnt implementations of a selector (cohort, featureanalysis, soures, ... anything that might need the user to select from a list of things).

PR #1328, opened 2019-02-13
Fixed change detection on children update

### LINKED ISSUE #1321: Cohort definition editor - gender selection doesn't update dirty flag
### Expected behavior

When creating a cohort definition with a gender condition, I expect that when I add/remove an item to the list that it allows me to save after detecting a change.

### Actual behavior

Adding/removing items from the gender list does not appear to update the dirty flag that should indicate that a save is required.

Create a cohort, add an entry event for "any visit" and an attribute criteria of "gender". Save the cohort. Next, add a gender of "Female" and note that the save icon is not active to allow for a save.

vlbe:
Fixes #1321

### DIFF
--- a/js/assets/ohdsi.util.js
+++ b/js/assets/ohdsi.util.js
@@ -41,7 +41,7 @@ define(['jquery', 'knockout', 'lz-string', 'lodash', 'crossfilter'], function ($
 		version: '1.0.0'
 	};
 
-	// private functions 	
+	// private functions
 	function _pruneJSON(key, value) {
 		if (value === 0 || value) {
 			return value;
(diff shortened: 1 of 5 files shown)


js/assets/ohdsi.util.js
```diff
@@ -65,7 +65,7 @@ define(['jquery', 'knockout', 'lz-string', 'lodash', 'crossfilter'], function ($
 					if (obj.hasOwnProperty(key)) {
 						let path = `${currentPath}.${key}`.replace(/^\./, '');
 						if (typeof obj[key] !== 'undefined' && obj[key] !== null && typeof obj[key].subscribe === 'function') {
```
chrisknoll: I think this statement is detecting if the field is an observable, and for this, I'd use ko.isObservable() for this, described here: https://stackoverflow.com/questions/9625591/determine-if-an-object-property-is-ko-observable

chrisknoll:
It took me a bit to understand what the extender was doing, but looks like it is notifying a parent observable of a change when one of the child observable changes...and I think it might be worth-while to add a comment in the extender to describe it's function.

chrisknoll:
Or not...

PR #1330, opened 2019-02-14
Fixed missing loaders

### LINKED ISSUE #1329: Loaders are missing
Loaders are missing in the following cases

**Login**
image

**Source creation** (click Save on new Source page):
image

vlbe:
Fixes #1329

### DIFF
--- a/js/services/AuthAPI.js
+++ b/js/services/AuthAPI.js
@@ -429,13 +429,7 @@ define(function(require, exports) {
 
 	var setAuthParams = function (tokenHeader) {
         token(tokenHeader);
-        return new Promise((resolve, reject) => {
-            loadUserInfo().then(() => {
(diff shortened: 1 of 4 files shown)


js/services/AuthAPI.js
```diff
@@ -429,7 +429,13 @@ define(function(require, exports) {
 
 	var setAuthParams = function (tokenHeader) {
         token(tokenHeader);
-        loadUserInfo();
+        return new Promise((resolve, reject) => {
```
pavgra: Why do you need to wrap it in the separate promise? Why not just to `return loadUserInfo();`?

vlbe: Yes, it's no need to wrap. I'll fix it.

PR #1332, opened 2019-02-14
Refactored getDomains method due to security fixes (migrated from 2.6)

pavgra:


js/providers/Vocabulary.js
```diff
@@ -12,39 +12,53 @@ define(function (require, exports) {
 	loadedPromise.resolve();
 
 	var defaultSource;
-	var domainPromise = $.Deferred();
```
johnSamilin: seems like `ConceptPicker.js` still uses `VocabularyProvider` (`modules/WebAPIProvider/VocabularyProvider`) wich doesn't have these fixes

js/providers/Vocabulary.js
```diff
@@ -12,39 +12,63 @@ define(function (require, exports) {
...
-			}).length > 0
-		});
-		if (prioritySources.length > 0)
-			defaultSource = prioritySources[0];
-		else // find the first vocabulary or CDM daimon
-			defaultSource = sources.filter(function (source) {
+
+		return sourceAPI.getSources().then(function (sources) {
+			if (sources.length === 0) {
+				resolve(domains);
```
johnSamilin: where this `resolve` comes from?

PR #1337, opened 2019-02-15
Fixes #1170 - valid package name enforcement

### LINKED ISSUE #1170: prediction package name issue
After reviewing the prediction study you type a name for the study package and download it.  If you use special characters in the name for the study R will throw an error as packages cannot have special character names.  At the moment there is no check for the package name. It would be good to check the name and either show a warning instructing the user to change the name or automatically remove special characters.

anthonysena:
Enforces an alphanumeric when defining the R package name.

### DIFF
--- a/js/extensions/bindings/alphaNumericExtender.js
+++ b/js/extensions/bindings/alphaNumericExtender.js
@@ -1,0 +1,25 @@
+define(['knockout'], function (ko) {
+    ko.extenders.alphaNumeric = function(target) {
+        //create a writable computed observable to intercept writes to our observable
+        var result = ko.pureComputed({
+            read: target,  //always return the original observables value
+            write: function(newValue) {
+                const isAlphaNumeric = RegExp('^[a-zA-Z0-9.]+$');
+                let current = target();
+    
+                //only write if it changed
+                if (isAlphaNumeric.test(newValue)) {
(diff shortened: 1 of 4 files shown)


js/extensions/bindings/alphaNumericExtender.js
```diff
@@ -0,0 +1,25 @@
...
+            read: target,  //always return the original observables value
+            write: function(newValue) {
+                const isAlphaNumeric = RegExp('^[a-zA-Z0-9.]+$');
+                let current = target();
+    
+                //only write if it changed
+                if (isAlphaNumeric.test(newValue)) {
+					target(newValue);
+                } else {
+					target.notifySubscribers(current);
```
pavgra: @anthonysena , why do we need to notify subscribers if the value wasn't changed?

anthonysena: The `newValue` may contain non-alphaNumeric values and so it is tested on line 11. If that fails, we want to keep the `current` value. This is done by notifying the original subscriber (`target`) that the value has changed (back to `current`) which effectively resets the observable and removes the non-alphaNumeric character.

If I take out the `notifySubscribers` call, the observable continues to retain the non-alphaNumeric values which is not the desired functionality.

Issue #1338, opened 2019-02-15
Entities names should be unique


anthonysena:
Related to #1307, entities (concept sets, cohort definitions, estimation, etc) should enforce a unique name in the system.

pavgra:
Issues described in PR

PR #1339, opened 2019-02-15
Fixes #1307 - Unique concept set names

### LINKED ISSUE #1307: Concept set names and cohort names are not unique
### Duplicates are in production

Cohort ID 8700 and 8405
Concept ID 101 and 102 & 7780 and 5892

###

anthonysena:
fixes #1307

### DIFF
--- a/js/pages/concept-sets/conceptset-manager.js
+++ b/js/pages/concept-sets/conceptset-manager.js
@@ -203,9 +203,9 @@ define([
 			// Next check to see that a concept set with this name does not already exist
 			// in the database. Also pass the conceptSetId so we can make sure that the
 			// current concept set is excluded in this check.
-			conceptSetService.exists(conceptSet.name, conceptSet.id)
+			conceptSetService.exists(conceptSet.name(), conceptSet.id)
 				.then((results) => {
-					if (results.length > 0) {
+					if (results.data.length > 0) {
 						this.raiseConceptSetNameProblem('A concept set with this name already exists. Please choose a different name.', txtElem);
 						abortSave = true;
 					}


pavgra:
@anthonysena , this should be checked and enforced from back-end side at first. Because now nobody stops me from creating a concept set with a used name by POST-ing request directly (e.g. via Postman or other client app)

anthonysena:
Agreed and I opened #1338 in recognition of this gap. For this PR, my goal was to restore functionality that previously worked to ensure that concept set names are unique. Let me know if you have any concerns with this fix. Thanks!

pavgra:
@anthonysena , the endpoint `exists` fails for me with 400 code and following message:
```
could not initialize proxy - no Session (through reference chain: java.util.ArrayList[0]->org.ohdsi.webapi.conceptset.ConceptSet["modifiedBy"]->org.ohdsi.webapi.shiro.Entities.UserEntity_$$_jvsta0d_17["login"])
```

Seems like again smth with transactions again even though it looks weird and concept set creator's / updater's fields are not used in this case.

chrisknoll:
Is this happening because it's trying to serialize the ConceptSet entity down to the client, and while it's serializing, it's tryign to access the 'ModifiedBy' field, which is assocated to a UserEntity, so it attempts to lazy load it, and that leads to the 'No Session' error?  I think the object should be converted to a DTO prior to serializing down.

pavgra:
Wow, I even haven't looked at the return type since for me the `exists` method automatically associated with true / false. But here, yes, it even returns ` Collection<ConceptSet>`. So, as for me, the best approach would be to return true / false, otherwise to go for converter and DTOs

pavgra:
Here is PR for WebAPI - https://github.com/OHDSI/WebAPI/pull/852

Issue #709, opened 2018-05-24
View button on the results tab of the PLE not working


johnSamilin:
### Expected behavior
View button on the results tab of the PLE is disabled if no results present
### Actual behavior
View button on the results tab of the PLE not working

PR #1340, opened 2019-02-16
Added github oauth support(WebAPI #709)

ssuvorov-fls:
fixes https://github.com/OHDSI/WebAPI/issues/709

### DIFF
--- a/js/Application.js
+++ b/js/Application.js
@@ -242,13 +242,11 @@ define(
 					}
 				});
 			}
-      checkOAuthError() {
-        let hash = window.location.hash;
-        if (hash) {
-          if (hash.includes("oauth_error_email")) {
-            alert("Empty email received from oauth server. Check whether it has public access");
-          }
-        }
+			checkOAuthError() {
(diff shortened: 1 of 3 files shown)


js/Application.js
```diff
@@ -242,7 +242,14 @@ define(
 					}
 				});
 			}
-
+      checkOAuthError() {
```
pavgra: indentation

js/Application.js
```diff
@@ -242,7 +242,14 @@ define(
...
+        let hash = window.location.hash;
+        if (hash) {
+          if (hash.includes("oauth_error_email")) {
```
pavgra: `hash && hash.includes("oauth_error_email")` instead of nested `if`

js/Application.js
ssuvorov-fls: fixed

js/Application.js
ssuvorov-fls: fixed

Issue #1313, opened 2019-02-06
Cohort characterization report needs cohort location in tables and plots to be consistent


jamieweaver:
### Expected behavior
When viewing a cohort characterization result that compares two cohorts, each cohort should be consistently on the same side of the table for each analysis, and similarly on the same axis of each scatter plot for each analysis.

### Actual behavior
A cohort in a comparison is sometimes on the left of a table and sometimes on the right. Similarly, the same cohort is sometimes on the x-axis and sometimes on the y-axis. 

### Steps to reproduce behavior
View https://epi.jnj.com/atlas/#/cc/characterizations/160/results/120000
For example all prevalence covariates and drug era medium term

Issue #1194, opened 2018-12-17
Cohort Comparison between cohorts that have covariates in only one cohort, leads to mis-matched rows on cohort compare view.


chrisknoll:
### Expected behavior
When comparing cohorts, the value of a covariate should appear properly for both cohorts.

### Actual behavior
if the covariate is one (such as an age group) that only exists in one cohort but not the other, the 'count' column value is assigned wrong. When looking at the single-cohort view, the values of the covariates appear correctly.

### Steps to reproduce behavior
Create 2 cohorts, with mutually exclusive cohort covariates (make one cohort with age = 34 another with the age = 35.  Run a cohort characterization with age-group covariates.

Here's an example of output from our internal report:
![image](https://user-images.githubusercontent.com/6818777/50106208-288a0b80-01fd-11e9-8265-1daadae54ccc.png)

Note that the age group 20-24 shows a count of 36, but a pct of 2.9% which can not be correct since the prior row of a count of 1 is 3.7% and the next row count of 2 is 7.41%.  The age group 20-24 appears to only exist in 1 cohort, which is leading to the mismatched display (that value probably belongs to the right-hand cohort.

Looking at the measurement report:
![image](https://user-images.githubusercontent.com/6818777/50106267-625b1200-01fd-11e9-9dc1-a6819a9d4734.png)

Sorting by count shows a mismatch:  the count of 150 is 12%, however, sorting the same table by the PCT column to see the rows that have 100%:

![image](https://user-images.githubusercontent.com/6818777/50106320-8ae30c00-01fd-11e9-8e22-66876064cb6b.png)

You can see that the rows that show 100% in the left-cohort exists in the right cohort, and the 27 count is 100% of the cohort, So the prior table showing 150 count is 12% probably is the right-cohort stats shown under the left-cohort's columns.

chrisknoll:
Due to the complexity of the reporting and the additional logic around strata, I'm going to ask the original devs that worked on this functionality to implement a fix.

johnSamilin:
@chrisknoll could you provide an exported characterization so I could import it locally and reproduce an issue?

chrisknoll:
Attached:
[mismatchedCC.txt](https://github.com/OHDSI/Atlas/files/2812752/mismatchedCC.txt)

johnSamilin:
[...]

PR #1347, opened 2019-02-18
Fixes data display for comparative CC

pavgra:
fixes https://github.com/OHDSI/Atlas/issues/1194
fixes https://github.com/OHDSI/Atlas/issues/1313


anthonysena:
@chrisknoll could you review this one? If not, let me know and we'll see who else can test it out. Thanks!

chrisknoll:
As an aside, i was able to check this with a previously generated CC with strata. However, since the SqlRender update, it's failing per OHDSI/WebAPI#840. If we could prioritize that fix as well, then I think we'll have a good working CC function.

PR #1361, opened 2019-02-20
Fixes #1355 - Cohort Characterization List Throws Error

### LINKED ISSUE #1355: Cohort characterization list throws error
Related to changes from #1348, the cohort characterization list will not load without throwing an error. Traced it back to this bit of code: https://github.com/OHDSI/Atlas/blob/master/js/pages/characterizations/components/characterizations/characterizations-list.html#L6 which should pass the function pointer instead of resolving the function.

I've already made the change and will push up the fix shortly while reviewing OHDSI/WebAPI#840.

anthonysena:
Small change to properly pass the function and not the value.

fixes #1355


pavgra:
@anthonysena , had to push a couple of additional fixes

Issue #1237, opened 2019-01-19
Prediction model settings doesn't properly reset to defaults


anthonysena:
### Expected behavior

When adding a model to a prediction study that has a setting that takes an array of values (i.e Random Forest model with maxDepth of [4,10,17]), if I change the default, I should be able to rest this value using the "Reset To Default" button along side the text box in the editor:

![image](https://user-images.githubusercontent.com/12902366/51421066-de32dc00-1b66-11e9-9845-59af77c49648.png)

### Actual behavior

The Reset To Default button does not properly reset the value if the input is an underlying array of values.

### Work Around

If you use the "Reset To Default" button, the underlying array's value is reset but the UI is not updated to reflect it. Close the Settings editor and then re-open the editor and the proper values will be set.

### Diagnosis

The individual model editors utilize a name/value pair which allows for a lookup to a JSON specification that contains the individual model setting descriptions, defaults, etc. In the case of a model setting that takes an array, it looks like this:

```
this.maxDepth = {
	name: 'maxDepth',
	value: ko.observable(this.modelSettings.maxDepth() && this.modelSettings.maxDepth().length > 0 ? this.modelSettings.maxDepth().join() : ''),
};
```

When the value is reset to the default, the `this.modelSettings.maxDepth` observable is reset but the `this.maxDepth.value` observable is unaware of the change and this is why the UI is never updated. Instead of each individual editor creating a set of properties in use by the editor, we could extend the base class to contain an array of these name/value pairs. Then, when we are setting the default, we could ensure that the `value` observable is also reset when it is an array.

anthonysena:
Will be addressed properly through: https://github.com/OHDSI/Atlas/issues/1168#issuecomment-464180772

PR #1386, opened 2019-02-27
Prediction model settings interface refactor

### LINKED ISSUE #1168: Prediction study: Random forest input type error for varImp
### Expected behavior
The input for the varImp variable in the random forest model should be either: i) true or ii) false or iii) true,false  and then this should create a vector in the json settings: i) [true] or ii) [false] or iii) [true,false]
### Actual behavior
it creates a string in the json settings i) "true" or ii)'false" or iii) 'true,false'

This causes an error when running the atlas created plp package

anthonysena:
This pull request changes the style for all prediction model setting that take as an input a list of integers/floats. Previous, this was done by specifying a comma-delimited list of values:

image

I've updated this to provide a proper list and the ability to add/remove items from it:

image

In the process, the ability to select both true & false has been addressed to fix #1168 and the revised user interface fixes the bug with #1237.

As a note, I've created a generic `multi-input` component to handle the new user interface and extended the `numericExtender.js` to make this more robust for checking integers and floats.

Tagging @jreps, @PRijnbeek for awareness.


js/components/multi-input.js
```diff
@@ -0,0 +1,88 @@
...
+) {
+    const typeIdentifier = {
+      integer: "integer",
+      float: "float"
+    }
+    class MultiInput extends AutoBind(Component) {
+      constructor(params) {
+        super(params);
+        this.selectedValueType = params.selectedValueType || "integer";
+        this.selectedValueExtender = params.selectedValueExtender || {};
```
pavgra: Shouldn't the `selectedValueExtender` be picked based on `selectedValueType` to have a single source of truth? E.g. now it is possible to pass selectedValueType = 'integer' and no extender or extender for float

js/components/multi-input.js
```diff
@@ -0,0 +1,88 @@
...
+        this.valueLabel = params.valueLabel || "Value";
+
+        this.hasSelectedValues = ko.pureComputed(() => {
+          return (this.selectedValues && this.selectedValues().length > 0);
+        });
+        this.hasDefaults = ko.pureComputed(() => {
+          return (this.defaultValues && this.defaultValues.length > 0);
+        });
+        this.enableAdd = ko.pureComputed(() => {
+         return (this.itemToAdd() !== null && this.itemToAdd().toString().length > 0 && !isNaN(this.getParser()(this.itemToAdd())));
```
pavgra: Why would you need `isNaN` if the extender would be enforced based on `
`selectedValueType`?

js/components/multi-input.less
```diff
@@ -0,0 +1,27 @@
+.multi-input {
+    margin-bottom: 20px;
```
pavgra: I would keep consistent in using `px` vs `rem` (`rem`-s are preferrable as for me)

js/pages/prediction/utils.js
```diff
@@ -25,6 +25,15 @@ define(
         }
     }
 
+    function getDefaultModelSettingName(defaultModelSettings, settingName) {
+        const settingList = defaultModelSettings.filter(item => item.setting === settingName);
```
pavgra: Why not
```
const setting = defaultModelSettings.find(item => item.setting === settingName);
return setting ? settingList[0].name : "SETTING NAME NOT FOUND!!";
```
?

Shouldn't it throw exception if the setting is not found?

js/components/multi-input.js
anthonysena: I see your point @pavgra - let me explain a bit more why I did it this way and perhaps there is another way to approach it. The `selectedValueExtender` is applied to the `itemToAdd` which has the effect of enforcing the extender rules for the value you are entering in the text box. For integers/floats, the `numeric` extender does a pretty good job of enforcing a standard. I also wanted to have a way to also enforce some type of validation before you add the item to the list. So, I added the `selectedValueType` to add in the proper parser to and to use that before adding the item to the list.

If I understand your suggestion, the extender and type should be embedded into the component so that it is simply a matter of specifying the `selectedValueType` only. If so, I agree and I'll make that change.

pavgra: > so that it is simply a matter of specifying the `selectedValueType` only. If so, I agree and I'll make that change.

yes, to avoid cases when you provide selectedValueType = 'integer' but the component allows string to be entered

js/components/multi-input.js
anthonysena: So the numeric extender has to allow for some non-numeric inputs - for example, if you are typing "-1", the extender must allow you to enter "-" first before it attempts to reformat anything. Without the isNaN, it would be possible to enter "-" for a list of integers and so this check prevents that from happening.

js/pages/prediction/utils.js
anthonysena: Agreed that it should throw an exception - I'll make this change as well.

js/components/multi-input.js
pavgra: Makes sense. But then I would also tighten this to `selectedValueType` (as the `selectedValueExtender` which we discussed above) and abstract it into a separate method, smth like `checkValue`. So, as a suggestion, why not to introduce an interface (base class) which contains three fields: `type`, `extender`, `checkValue`, and extend that for Integer, Float, whatever. And then work via that interface.

anthonysena: Sure - let me add that into this branch.

anthonysena:
@pavgra - I've made the changes per your feedback. Let me know if you have any other questions/concerns. Thanks!

pavgra:
Multi selected is shrinked:
image
image

pavgra:
image

anthonysena:
I'm unable to reproduce the problem on my side using the latest changes on this branch:

### Decision Tree

image

### Random Forest

image

PR #1388, opened 2019-02-27
Fixes Error Export All Concept Sets to CSV

wivern:


js/services/file.js
```diff
@@ -20,6 +20,7 @@ define(
               const blob = new Blob([xhr.response], { type: "octet/stream" });
               saveAs(blob, filename);
             }
+            reject({ status: xhr.status, statusText: xhr.statusText });
```
pavgra: Why not put inside of else for clarity?

PR #1359, opened 2019-02-20
Provide a button or method to export the tables of cohort characteriz…

johnSamilin:
fixes #1356

PR #1396, opened 2019-02-28
CC refactoring and CSV export

### LINKED ISSUE #1356: Provide a button or method to export the tables of cohort characterization
### Expected behavior

When reviewing cohort characterization results, I would expect that I can export results displayed in a tabular format to a csv.

### Actual behavior

When reviewing cohort characterization results, the user interface lacks the ability to export to a table.

Related to #1144 

#### Prevalence Related Columns
- Analysis ID + Name
- Strata ID + Name
- Cohort ID + Name
- Covariate ID + Name
- Count
- Prevalence or %

#### Distribution Related Columns
- Analysis ID + Name
- Strata ID + Name
- Cohort ID + Name
- Covariate ID + Name
- Min/Max/P10/P25/Median/P75/P90/StdDev/Mean
- Count

#### Comparison Related Columns
- Analysis ID + Name
- Strata ID + Name
- Target Cohort ID + Name
- Comparator Cohort ID + Name
- Covariate ID + Name
- Target Count
- Target Prevalence or %
- Comparator Count
- Comparator Prevalence or %
- Std. Diff Of Mean

pavgra:
Instead of https://github.com/OHDSI/Atlas/pull/1359:
- Single table template for all stat types (prev. / distribution / perv. comparison / dist. comparison)
- Single target data structure provided to the table template
- Hierarchy of converters to remove redundancy and provide better code structuring
- Added export buttons

fixes #1356


js/pages/characterizations/services/conversion/BaseStatConverter.js
```diff
@@ -0,0 +1,134 @@
...
+
+        extractStrata(stat) {
+            throw "Override extractStrata with actual implementation";
+        }
+
+        setNestedValue(result, field, strataId, cohortId, value) {
+        	if (result[field][strataId] === undefined) {
+				result[field][strataId] = {};
+			}
+			result[field][strataId][cohortId] = value;
```
johnSamilin: possible `cannot set field of undefined`

chrisknoll:
For distribution export, let's arrange the column order:
Avg -> StdDev : let's put these 2 columns side by side because one realtes to the other.

When you move Avg out then the distribiton columns will be ordered:
Min->P10->P25->Median (also known as P50)->P75->P90->Max.

If you want to put the Avg and StdDev colms on the left side of those, or following the percentile columns that's fine, but i think it would be preferable to have:
{Analysis Columns} -> {Count} -> {Avg, StdDev} -> {Percentile Cols}

In the tables, while you can filter, it doesn't show the total number of rows in the table such that when filtering it would say 'Showing X of Y total rows'.  But we don't even see the total rows.

Demograpnics analysis:  technically you can't explore those because those aren't associated to concepts.  Same would go to custom Criteria Features...only thing you can really explore is those comming from Feature Extraction....we might be able to explore by criteria features if there was a way to assocaite a specific concept to a criteria feature, but that has a ton of pitfalls so I wouldn't even recommend it.

pavgra:
> In the tables, while you can filter, it doesn't show the total number of rows in the table such that when filtering it would say 'Showing X of Y total rows'. But we don't even see the total rows.

@chrisknoll , hm, are you talking about smth different?
image

chrisknoll:
Maybe it's when you are in strata view?
image

image

image

image

There's nothing there for me to circle.

chrisknoll:
I wonder if this is some sort of datatables version?

pavgra:
Now I see it. Weird

pavgra:
@chrisknoll , if I haven't missed anything, your comments were addressed

PR #1398, opened 2019-02-28
[WIP] Changed polling service to use setTimeout function

vlbe:
related to https://github.com/OHDSI/WebAPI/issues/879

### DIFF
--- a/js/components/user-bar.js
+++ b/js/components/user-bar.js
@@ -23,7 +23,7 @@ define([
 			jobDetailsService,
 			momentApi,
 			lodash,
-			PollService,	
+			PollService,
 		) {
 	class UserBar extends Component {
 		constructor(params) {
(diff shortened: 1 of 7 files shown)


js/components/user-bar.js
```diff
@@ -143,12 +143,15 @@ define([
...
+
 							}
 						});
 					})
 					.catch(() => {
-						console.warn('The server error occurred while getting all notifications');                        
+						console.warn('The server error occurred while getting all notifications');
+					})
+					.finally(() => {
+						PollService.start(this.pollId);
```
pavgra: Why do you add `PollService.start(this.pollId);` in each place manually? Why cannot this logic (`setTimeout -> callback -> setTimeout -> ...`) be encapsulated into PollService?

js/services/Poll.js
```diff
@@ -14,23 +14,30 @@ define(['knockout', 'visibilityjs'], (ko, Visibility) => {
 
   class PollService {
     add(callback, interval = 1000, ...args) {
-      const intervalId = setInterval(() => {
         if (isPageForeground()) {
```
pavgra: So how the polling will be continued after you first switched from the tab somewhere and then reached it back?

PR #1405, opened 2019-03-01
Fixes columns order in CC results page

wivern:


js/pages/characterizations/services/conversion/BaseStatConverter.js
```diff
@@ -50,7 +50,7 @@ define([
             analysis.reports.forEach((r, i) => r.stats.forEach(mapCovariate(r)));
 
             cohorts.forEach((c, i) => {
-              for (let strataId of strataNames.keys()) {
+              for (let strataId of Array.from(strataNames.keys()).sort()) {
```
pavgra: Shouldn't
- the values be either sorted inside map
or
- the sorting be repeated inside the template
?

js/pages/characterizations/components/characterizations/characterization-view-edit/characterization-results.html
```diff
@@ -64,22 +64,18 @@ <h3 data-bind="css: $component.classes('analysis-name'), text: $component.analys
                                                 <th data-bind="css: $component.classes('th-cov-count'), attr: { rowspan: $parents[1].stratified ? 2 : 1 }, text: $data"></th>
                                             <!-- /ko -->
                                         <!-- /ko -->
-                                        <!-- ko foreach: Array.from($parent.strataNames) -->
-                                            <!-- ko ifnot: $data[0] === 0 -->
+                                        <!-- ko foreach: Array.from($parent.strataNames).filter(a => a[0] !== 0).sort((a,b) => a[0] - b[0]) -->
```
pavgra: let's move the logic to controller

pavgra:
Aren't you sorting by name in one place and by ID - in another?

PR #1415, opened 2019-03-05
Fixes IR allows running even if previous execution is not completed yet

wivern:
Fixes OHDSI/WebAPI#898


js/pages/incidence-rates/ir-manager.js
```diff
@@ -173,35 +177,45 @@ define([
...
-			.catch(() => {
+			} catch(e) {
 				this.close();
-			});
+			} finally {
+				this.loadingInfo(false);
+			}
+		}
+
+		async getSummaryList(id, source, silently = true) {
```
pavgra: Maybe, `loadSmth` instead of `getSmth`? the function returns void

pavgra:
Polling for existing results should be silent (w/o `n/a`):
ir-polling

Issue #1418, opened 2019-03-05
Bundling application removes .done() from Promise


chrisknoll:
### Expected behavior
Invoking .done() on a promise should work after budnling.

### Actual behavior
promise.done is not a function

### Steps to reproduce behavior

Bundled Mode:
Set breakpoint on line 1011 on cohort-definition-manager.js on this line:
```
        vocabularyApi.getConceptSetExpression(newConceptSet.id).done(function (result) {
```

When you invoke the UI to import a concept set from the repository, the breakpoing will trigger.  I created a local var with the result of the getConceptSetExpression, and the promise returned does not have `.done()`
![image](https://user-images.githubusercontent.com/6818777/53828560-06766e80-3f4c-11e9-8045-8c3f3d5a71e4.png)

However, when running non-bundled, (note: breakpoint in non-bundled is on line 918:

![image](https://user-images.githubusercontent.com/6818777/53828625-2e65d200-3f4c-11e9-9603-f61d19be369a.png)

Something during bundling (possibly babel transform?) is causing the promise API to not include a `done()` method. I found on promise.org that done is not standardized:
https://www.promisejs.org/

Note that promise.done (used in the examples in this section) has not been standardised. It is supported by most major promise libraries though, and is useful both as a teaching aid and in production code. I recommend using it along with the following polyfill (minified / unminified):
```<script src="https://www.promisejs.org/polyfills/promise-done-7.0.4.min.js"></script>```

So, i do not know if we need to add this polyfill as part of bundling.

PR #1419, opened 2019-03-05
Added Promise.done polyfill

pavgra:
fixes https://github.com/OHDSI/Atlas/issues/1418

In non-bundled mode `less.js` adds polyfill for non-standardized `Promise.done`:
image

During bundling this thing seems to be cut off by requirejs optimizer. Even though the `Promise.done` is not a standard thing and as for me should be avoided going further, right before the release of 2.7.0 I'm not treating migration of `done` -> `then` as a good idea. Therefore, I've brought that exact piece of code from `less.js` into `polyfill.js` which is imported by bundler.

### DIFF
--- a/build/polyfill.js
+++ b/build/polyfill.js
@@ -1,1 +1,12 @@
 define(['../node_modules/@babel/polyfill/dist/polyfill'], () => {});
+
+if (typeof Promise.prototype.done !== 'function') {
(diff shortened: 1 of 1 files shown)


anthonysena:
Tested this in bundled mode and it worked well. Thanks @pavgra!

PR #1421, opened 2019-03-05
Export to CSV in explore view.

chrisknoll:


chrisknoll:
Ok, I applied the same .less styles that are used elsewhere.

Unfortunately, this seems to be a bot of copy-paste between different .less files. Maybe there's a way to have some of these button layouts shared?

pavgra:
If we'd always use bundler, we could create functions / shared classes in less and import those. But now I think your approach is OK

chrisknoll:
> If we'd always use bundler, we could create functions / shared classes in less and import those. 

I'm not sure a bundler is required for that:

Consider if we had a shared module with several BEMHelper instances in it:

```
define(['BEMHelper'],function(BEMHelper) { 

return {
  chartClasses: new BEMHelper(),
  datatableClasses: new BEMHelper()
}
});
```

Etc, such that instead of having to copy and paste things like button styles or graph axis styles, we can just import the 'sharedClasses' module and reference the instance of the bemhelper (chartClasses vs datatableClasses) to expose the specific styles we need.

pavgra:
http://lesscss.org/features/ -> Section "Import Statements"
\+
http://lesscss.org/features/#mixins-feature

PR #1435, opened 2019-03-07
fixed permissions checks

### LINKED ISSUE #1420: Unauthorized error message is not displayed
There is no error message regarding the lack of permissions on following Atlas pages for unauthorized user:
- Search
- Concept Sets
- Estimation
- Prediction
- Jobs

\+ in some pages AJAX requests are attempted to be done even though there are no perms

johnSamilin:
fixes #1420

### DIFF
--- a/js/pages/concept-sets/concept-manager.html
+++ b/js/pages/concept-sets/concept-manager.html
@@ -219,3 +219,5 @@
 		</div>
 		<div class="clear"></div>
 	</div>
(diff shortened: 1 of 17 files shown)


pavgra:
What about preventing AJAX calls if a user reaches pages but does't have enough perms?

pavgra:
What about Vocabularies tab?

PR #1436, opened 2019-03-07
Adds concept code to cc results

### LINKED ISSUE #1427: Characterization
### Expected behavior

Covariates should be uniquely represented in table, no redundant records

### Actual behavior

Redundant covariates observed.  Likely some issue with SQL join.  Confirm that joins are on covariate_id and not covariate_name,

Example results showing redundant records from drug domain:

image

Review results from characterization with >=2 cohorts and using Features using Drug Group Era.

wivern:
Closes #1427

### DIFF
--- a/js/pages/characterizations/services/conversion/PrevalenceStatConverter.js
+++ b/js/pages/characterizations/services/conversion/PrevalenceStatConverter.js
@@ -35,7 +35,11 @@ define([
                     title: 'Concept ID',
                     data: 'conceptId',
                     render: (d, t, r) => {
-                        return `<a href="/#/concept/${r.conceptId}" data-bind="tooltip: '${r.conceptName}'">${r.conceptId}</a>`
+                        if (r.conceptId === null || r.conceptId === undefined) {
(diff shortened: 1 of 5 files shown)


js/pages/characterizations/services/conversion/PrevalenceStatConverter.js
```diff
@@ -29,7 +29,16 @@ define([
...
 
 		getDefaultColumns(analysis) {
-		    return [ this.getСovNameColumn() ];
+		    return [
+		        this.getСovNameColumn(),
+                {
+                    title: 'Concept ID',
+                    data: 'conceptId',
+                    render: (d, t, r) => {
+                        return `<a href="/#/concept/${r.conceptId}" data-bind="tooltip: '${r.conceptName}'">${r.conceptId}</a>`
```
anthonysena: While testing this, I ran into an issue caused by a concept name with an apostrophe in it - perhaps use `CommonUtils.renderBoundLink`?

PR #1437, opened 2019-03-07
Adds concept code to cc results

### LINKED ISSUE #1427: Characterization

wivern:
Closes #1427


pavgra:
The fix will be merged back from RC to master when RC is finished

PR #1439, opened 2019-03-07
ATLP-642 Error page after IR deletion

wivern:
Fixes Error page after IR deletion

### DIFF
--- a/js/pages/incidence-rates/ir-manager.js
+++ b/js/pages/incidence-rates/ir-manager.js
@@ -171,16 +171,14 @@ define([
 			this.init();
 		}
 
-		async pollForInfo({silently = false, analysisId = null} = {}) {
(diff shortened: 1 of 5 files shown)


js/pages/incidence-rates/ir-manager.js
```diff
@@ -246,7 +243,16 @@ define([
 				this.pollForInfo();
 				this.pollTimeout = PollService.add(() => this.pollForInfo({ silently: true }), 10000);
 			});
-		};
+		}
+
+		onRouterParamsChanged(params = {}) {
+			const { analysisId } = params;
+			if (!analysisId) {
```
pavgra: Why `parseInt` isn't used here? If I'm not mistaken, `analysisId` in such case equals to `"0"` (string)

pavgra:
When I pick an IR from list:
image

PR #1440, opened 2019-03-07
Fixes Error page after IR deletion

wivern:


js/pages/incidence-rates/ir-manager.js
```diff
@@ -316,6 +316,7 @@ define([
 			IRAnalysisService.deleteAnalysis(this.selectedAnalysisId()).then(() => {
 				this.selectedAnalysis(null);
 				this.dirtyFlag(new ohdsiUtil.dirtyFlag(this.selectedAnalysis()));
+				PollService.stop(this.pollTimeout);
```
pavgra: Why this is not done by just `dispose`?

pavgra:
What about `onRouterParamsChanged`?

js/pages/incidence-rates/ir-manager.js
```diff
@@ -177,30 +171,36 @@ define([
 			this.init();
 		}
 
-		async pollForInfo({silently = false} = {}) {
+		async pollForInfo({silently = false, analysisId = null} = {}) {
```
pavgra: Why do you need to pass the `analysisId` and then compare it to `selectedAnalysisId`? And then check for empty id (`if (id) {`)?

I assume that it might be related to Polling, but then you need just to `PollService.stop(this.pollTimeout);` before doing `this.pollTimeout = PollService.add(() => this.pollForInfo`, don't you?

Issue #1461, opened 2019-03-08
Profiles search doesn't display results in bundled mode


pavgra:
In non-bundled mode `this` in `profileChart.js` -> `categoryScatterPlot` -> `rectangle` is either `instanceof SVGElement` or `undefined`.

In bundled mode the `this` equals to `Window` instead of `undefined`:
![image](https://user-images.githubusercontent.com/10010071/54041007-a3c3e380-4194-11e9-925e-82c6cc7d8908.png)

And later fails inside of d3 when it attempts to call  `querySelectorAll` on `Window`:
![image](https://user-images.githubusercontent.com/10010071/54041037-b5a58680-4194-11e9-8dbb-26fdf8972b89.png)

PR #1462, opened 2019-03-08
Fixed Profiles strict mode

pavgra:
fixes https://github.com/OHDSI/Atlas/issues/1461


anthonysena:
@pavgra  I'm getting the following errors in profiles (does not appear to be related to this PR) but wanted to see if you are also getting this:

image

Its related to the tooltip positioning when hovering over events in the patient profile timeline.

PR #1511, opened 2019-03-14
Fixed IR Analysis reset to new instance

### LINKED ISSUE #1509: Profiles nav resets IR analysis to new instance.

1. Open any IR
1. Close the IR
1. Observe that IR icon in the menu is green
1. Go to 'Profiles' tab

### Expected behavior
IR icon is white, not green, in step 3

### Actual behavior
IR icon is green on the left pane in step 3. And if you go to IR tab the list of all IR's is expected but IR from step 1 is opened

vlbe:
Fixes #1509

### DIFF
--- a/js/pages/incidence-rates/ir-manager.js
+++ b/js/pages/incidence-rates/ir-manager.js
@@ -1,11 +1,11 @@
 define([
-	'knockout', 
+	'knockout',
 	'text!./ir-manager.html',
 	'services/IRAnalysis',
 	'services/SourceAPI',
 	'services/CohortDefinition',
-	'./components/iranalysis/IRAnalysisDefinition', 
-	'./components/iranalysis/IRAnalysisExpression', 
+	'./components/iranalysis/IRAnalysisDefinition',
+	'./components/iranalysis/IRAnalysisExpression',
 	'assets/ohdsi.util',
 	'appConfig',
 	'atlas-state',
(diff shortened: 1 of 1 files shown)


chrisknoll:
@vlbe,  I see the code changes, but a gap in my understanding is why this wouldn't have impacted all the other navs.  But some do, and some don't and I worry that there's a deeper issue we should address:

After opening up an IR and closing it, the following Navs result in a new instance:
- Search
- Profiles

However, the following **do not** result in a new instance:
- Home
- Concept Sets
- Cohort Definitions
- Characterizations
- Pathways
- Estimation
- Prediction
- Jobs
- Configuration
- Feedback

So, at risk of drawing this out more, is there something that needs to be investigated as to why some of the navs resulted in the behavior, while others did not?  Your code change only impacts handling onRouteChanged inside IR analysis, but since Search and Profiles seem to reset the IR analysis, what is different about Search and Profiles route navigation?  Is there some deeper issue about these navs?

vlbe:
@chrisknoll 
We have overridden methods **onRouterParamsChanged** and **dispose** in ir-manager. In Page component in costructor we subscribe to routerParams and if params change we call  **onRouterParamsChanged**. So the  problem was that we don't unsubscribe when page dispose because in ir-manager dispose() doesn't call super.dispose().

PR #1512, opened 2019-03-14
Fixes Incidence Rate Generation errors in UI 

### LINKED ISSUE #1507: Incidence Rate Generation errors in UI
### Expected behavior

When generating an IR analysis using the Generate pop-up to a new source that has yet to have an execution, we expect the interface to:

1. Not receive an error in the developer console
2. To poll for the job status properly

### Actual behavior

1. We received an error in the console when generating on a source that had 0 executions. The response of the execution service returned a 204 error which lead to properties on the returned job object containing no properties.

2. We've also observed that other sources with results also change to "running" even though it was not the one we selected for generation: 

image

Once the job kicked off, we got the display above even though we only ran this on JMDC. After the first poll on the new job, the sources then reset to their proper running state.

We used an IR that was already generated on a single source. We then generated on a 2nd source and observed the problems mentioned above.

wivern:
Fixes #1507

### DIFF
--- a/js/components/generation/select-sources-popup.js
+++ b/js/components/generation/select-sources-popup.js
@@ -24,7 +24,7 @@ define([
 
 			this.showModal = params.showModal;
 			this.selectedSources = params.selectedSources;
-			this.sourceOptions = ko.computed(() => params.sources().map(s => ({ source: s, selected: ko.computed(() => this.selectedSources().includes(s)) })));
+			this.sourceOptions = ko.computed(() => params.sources().map(s => ({ source: s, selected: ko.computed(() => this.selectedSources().find(ss => ss.sourceKey === s.sourceKey)) })));
 			this.submit = params.submit;
 
 			this.tableDom = "Bfrtp";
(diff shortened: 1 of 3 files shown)


js/components/generation/select-sources-popup.js
```diff
@@ -24,7 +24,7 @@ define([
 
 			this.showModal = params.showModal;
 			this.selectedSources = params.selectedSources;
-			this.sourceOptions = ko.computed(() => params.sources().map(s => ({ source: s, selected: ko.computed(() => this.selectedSources().includes(s)) })));
+			this.sourceOptions = ko.computed(() => params.sources().map(s => ({ source: s, selected: ko.computed(() => this.selectedSources().find(ss => ss.sourceKey === s.sourceKey)) })));
```
pavgra: Why this was switched to sourceKey?

wivern: Well, because source object is changing during polling

PR #1515, opened 2019-03-15
Fixes Google IAP session refresh. Add Google IAP logout

pavgra:
fixes https://github.com/OHDSI/WebAPI/issues/968

### DIFF
--- a/js/components/welcome.html
+++ b/js/components/welcome.html
@@ -42,6 +42,7 @@
 			<div data-bind="if: isLoggedIn() && !tokenExpired()">
                 <div data-bind="text: $component.status"></div>
 				<span data-bind="css: classes({ element: 'signout-btn', extra: ['btn btn-sm btn-default'] }), click: $component.signout, visible: $component.token()"><i class="fa fa-sign-out"></i> Sign Out</span>
+				<span data-bind="css: classes({ element: 'signout-btn', extra: ['btn btn-sm btn-default'] }), click: $component.signoutIap, visible: $component.isGoogleIapAuth()"><i class="fa fa-sign-out"></i> Sign Out</span>
 			</div>
 		</div>
 	</div>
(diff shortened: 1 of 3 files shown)


js/components/welcome.html
```diff
@@ -42,6 +42,7 @@
 			<div data-bind="if: isLoggedIn() && !tokenExpired()">
                 <div data-bind="text: $component.status"></div>
 				<span data-bind="css: classes({ element: 'signout-btn', extra: ['btn btn-sm btn-default'] }), click: $component.signout, visible: $component.token()"><i class="fa fa-sign-out"></i> Sign Out</span>
+				<span data-bind="css: classes({ element: 'signout-btn', extra: ['btn btn-sm btn-default'] }), click: $component.signoutIap, visible: $component.isGoogleIapAuth()"><i class="fa fa-sign-out"></i> Sign Out</span>
```
chrisknoll: Could we abstract this `isGoogleIAPAuth()` check?  Just at the look of it:  the signout button is hidden in the case of google IAP.  Can we have a class of authentication providers that 'supportsLogout'? Could we define a set of capabilities related to each auth provider, and make the UI behave based on auth capabilities vs. specific auth providers.

Thinking a bit more on this:  once someone logs in to Atlas, shouldn't the mechanism of authentication be irrelevant after they authenticate via one of the many providers?  When they sign out in atlas, shouldn't' that just mean we invalidate their authentication session in atlas?

pavgra: > When they sign out in atlas, shouldn't' that just mean we invalidate their authentication session in atlas?

They should. And this is exactly what is done here: a user is redirected to IAP page which invalidates the user's session. This is Identity Proxy which stores the session in case of Google IAP provider versus the Atlas app in all other cases: `GoogleIapJwtAuthFilter` vs `AtlasJwtAuthFilter`

> Could we define a set of capabilities related to each auth provider, and make the UI behave based on auth capabilities vs. specific auth providers.

We could. I just follow the rule of making minimally required changes when we talk about fixes right before the release. Therefore, I think such abstraction could take place in 2.8

PR #1517, opened 2019-03-15
added default sorting for characterization reports results

### LINKED ISSUE #1127: Presort characterization table results to display highest to lowest %
### Request
When the characterization tables display they are sorted alphabetically by concept name. It would be more intuitive to have the default be sorting by highest to lowest % for each analysis generated.

johnSamilin:
fixes #1127


johnSamilin:
>What about comparison mode - where is sort by stdDiff?
This is the last column, it's being used by default

js/pages/characterizations/services/conversion/BaseStatConverter.js
```diff
@@ -76,10 +76,15 @@ define([
                 defaultColNames: this.getDefaultColumns().map(col => col.title),
                 perStrataColNames: this.getReportColumns(0,0).map(col => col.title),
                 columns: columns,
+                defaultSort: this.getDefaultSort(columns.length, cohorts.length),
                 data: Array.from(data.values()),
             };
         }
 
+        getDefaultSort(columnsCount, cohortsCount) {
+            return [[ columnsCount - 1, "desc" ]];
```
pavgra: Let's add comment that this is either for Std Diff (in comparison mode) or Pct (in single cohort mode)

PR #1524, opened 2019-03-15
Fixed typerror on concept set tab of new ir page

### LINKED ISSUE #1519: Uncaught TypeError: VocabularyAPI.getConceptsById(...).then(...).fail is not a function
### Expected behavior
No error
### Actual behavior
Uncaught TypeError: VocabularyAPI.getConceptsById(...).then(...).fail is not a function
1) Open Incident Rates
2) Create new
3) Open Concept Set tab
4) Click new Concept Set
5) Look in console

vlbe:
Fixes #1519

### DIFF
--- a/js/components/conceptsetbuilder/components/IncludedConcepts.js
+++ b/js/components/conceptsetbuilder/components/IncludedConcepts.js
@@ -165,9 +165,9 @@ define([
 		self.refresh = function() {
 			self.isLoading(true);
 			self.selectedConcepts([]);
-			VocabularyAPI.resolveConceptSetExpression(self.conceptSet()).done(function (identifiers) {
-				VocabularyAPI.getConceptsById(identifiers).then(function (concepts) {
-					concepts.data.forEach(function (concept) {
+			VocabularyAPI.resolveConceptSetExpression(self.conceptSet()).then(function (identifiers) {
+				VocabularyAPI.getConceptsById(identifiers).then(({ data: concepts }) => {
+					concepts.forEach(function (concept) {
 						concept.RECORD_COUNT = 'timeout';
 						concept.DESCENDANT_RECORD_COUNT = 'timeout';
 					});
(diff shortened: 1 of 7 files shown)


js/components/conceptsetbuilder/components/IncludedConcepts.js
```diff
@@ -159,24 +159,24 @@ define([
...
-		
+
 		// behaviors
-		
+
 		self.refresh = function() {
 			self.isLoading(true);
 			self.selectedConcepts([]);
-			VocabularyAPI.resolveConceptSetExpression(self.conceptSet()).then(function (identifiers) {
+			VocabularyAPI.resolveConceptSetExpression(self.conceptSet()).done(function (identifiers) {
```
chrisknoll: My understanding on `promise.done` is that it's not an official syntax.  @pavgra , can you confirm that? I thought we saw an issue with the bundler that done was stripped out, but we polly-filled it in. Should we depend on the polyfill or avoid using `.done()`?

chrisknoll:
Although there are no errors when performing the steps to reproduce, there is an issue on the 'Included Concepts' Tab:

Steps to reproduce:
1. New IR
2. Go to concept sets
3. New Concept Set
4. Click Add
5. Search for 'psoriasis'
6. Click one of the standard concepts (I chose 4063431 Plaque psoriasis)
7. Observe included/mapped tabs look like they are refreshing.
8. Observe the requests in console is returnign data
9. Failure: the included concepts tab doesn't include any concepts, even the one that was added in the concept set itself.

Here's an image:
image

js/components/conceptsetbuilder/components/IncludedConcepts.js
vlbe: Actually resolveConceptSetExpression using jquery.ajax under the hood. We talk about refactoring with @pavgra and decided not to make any refactoring in this rc branch.

PR #1541, opened 2019-03-16
Fix concept link display

### LINKED ISSUE #1527: [ATLAS 2.7 TEST] Search Classifications are no longer highlighted purple
### Expected behavior
Classification concepts are highlighted in purple

### Actual behavior
Classifications concepts are highlighted in blue in the search

1. Go to Search --> 21000471
2. See the code is listed in BLUE

image

But it is classification:
image

This behavior looks fine on other screens.

anthonysena:
Updates the commonUtils to share a common way to determine the `standard_concept` designation of a concept and to return a css class defined in atlas.css. 

Fixes #1527

### DIFF
--- a/js/utils/CommonUtils.js
+++ b/js/utils/CommonUtils.js
@@ -60,7 +60,7 @@ define([
 
 	function getLinkClass(data) {
 		var switchContext;
-		if (data.STANDARD_CONCEPT == undefined) {
+		if (data.STANDARD_CONCEPT === undefined) {
 			switchContext = data.concept.STANDARD_CONCEPT;
 		} else {
 			switchContext = data.STANDARD_CONCEPT;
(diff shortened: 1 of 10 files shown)


js/utils/CommonUtils.js
```diff
@@ -58,23 +58,27 @@ define([
 		return false;
 	}
 
-	function contextSensitiveLinkColor(row, data) {
+	function getLinkClass(data) {
```
pavgra: Which link? I would give it a more specific name, e.g. `getConceptLinkClass`

pavgra:
Why not to apply the utility method to `ConceptManager` and `IncludedConceptsets` (and maybe some other places)?

anthonysena:
@pavgra - I made some updates based on your feedback. The one place I did not change were in the references in const.js - when I attempted to update these, I found that there were other areas that broke since the `sharedState` would not resolve. I'd imagine there is some dependency chain that is broken by including commonUtils in const.js so I've avoided that for now.

Issue #1518, opened 2019-03-15
Can't see Included Concepts in the table of Concept Sets


anemniugin:
### Expected behavior
Can see  included concepts

### Actual behavior
Empty table

### Steps to reproduce behavior
Login
Go to Concept Sets
Create new Concept Set
Go to Search
Search "aspirin"
Add one or more Aspirin to cart
Go to concept set by link at the top of the table
Open Included Concept tab(with digit)

if you set checkbox e.g. Descendants, and set it off, you can see correct rows in Included Concepts tab,

anthonysena:
Looking into this a bit more, I'm finding that the trouble related to this issue stems from the way that we attempt to load up included concepts. Here is the code:

https://github.com/OHDSI/Atlas/blob/v2.7.0-rc/js/Model.js#L779

```
loadIncluded(identifiers) {
	this.loadingIncluded(true);
	const data = identifiers || this.conceptSetInclusionIdentifiers();
```

It is assumed that the `identifiers` for the concept set (i.e. the list of concepts that are resolved from the concept set expression) are passed into this function to look up the identifiers. When `identifiers` are not passed into the function, it will attempt to use `this.conceptSetInclusionIdentifiers()` which is set when calling the resolveConceptSetExpressionSimple function:

https://github.com/OHDSI/Atlas/blob/v2.7.0-rc/js/Model.js#L399

In the scenario described in this issue, the problem lies in the fact that the concept set has yet to be saved and as a result, both the `identifiers` and `this.conceptSetInclusionIdentifiers()` are not specified therefore returning 0 included concepts. 

The approaches that both @pavgra and I put into #1546 simply adds more calls to `resolveConceptSet` to work around the root cause of this issue. Instead of applying this work around which then causes needless calls to the same REST endpoints, we should examine the way in which we are resolving a concept set to find the included concepts & source codes. This should be uniformly applied regardless of if the concept set is new or saved.

PR #1546, opened 2019-03-16
Fixes Concept Set reload when new concepts picked from vocabulary

pavgra:
fixes https://github.com/OHDSI/Atlas/issues/1518

### DIFF
--- a/js/Model.js
+++ b/js/Model.js
@@ -348,6 +348,8 @@ define(
 					}
 				});
 				this.currentConceptSetMode.subscribe(this.onCurrentConceptSetModeChanged);
+
+				sharedState.selectedConcepts.subscribe(() => this.resolveConceptSetExpression());
 				/*
 					probably unreachable code
 					
--- a/js/Model.js
+++ b/js/Model.js
@@ -349,9 +349,6 @@ define(
 				});
 				this.currentConceptSetMode.subscribe(this.onCurrentConceptSetModeChanged);
 
-				sharedState.selectedConcepts.subscribe(() => {
-					this.resolveConceptSetExpression().then(() => this.onCurrentConceptSetModeChanged(this.currentConceptSetMode()));
-				});
 				/*
 					probably unreachable code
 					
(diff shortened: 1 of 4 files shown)


anthonysena:
This fixes the behavior reported in #1518 however it breaks in a different scenario. Do all of the steps in #1518 - then, select one of the concept's descendants. Switch to the Included Concepts tab and select one of the concepts to include it in the concept set. The display then disappears. The concept is added to the concept set and then you need to switch to the Concept Set Expression and back to the Included Concept tabs for it to re-load. Before this change, the concept is properly added to the concept set and the display remained intact.

pavgra:
@anthonysena , requests are duplicated now (when you first time switch to the Included Concepts tab)
image

anthonysena:
I'm going to push this to 2.7.1 and explain why in #1518 to keep the bug & troubleshooting together.

chrisknoll:
Be aware I am also looking at issues related to datatables, may want to reference my changes in #1582.

anthonysena:
Sounds good @chrisknoll - the other noteworthy item here is related to the changes underway in #1590. I think once those are in master, we can circle back on this PR - I recall from working with @pavgra on this item that part of the challenge (for me at least) is that there were functions firing in Model.js and in the downstream components and it was fairly difficult to coordinate them all since they were in different places. Let us know if you think it is sensible to wait for you to finish #1582 and then also incorporate those changes here.

anthonysena:
@pavgra - coming back to this item and looked deeper into the behavior you noted earlier (that there were multiple calls to resolve included concepts). What appears to happen is that there is a redundant call the first time we try to find the included concepts - what I noticed is that one call will pass an empty concept set and then finally sends up the current concept set. However, subsequent calls are not repeated even when the concept set definition is changed.

When I contrast this behavior to https://github.com/OHDSI/Atlas/pull/1546/commits/0e6c6b62505532f15e7a80f5084ad2049f64d86b#diff-37aeda83c046a4ff6731a45a79b2cfabR352, the subscription is done at the global level. The result is that any changes to an active concept set then make a call to resolve it which adds a lot of extra, unnecessary traffic to the server.

As discussed between us and @chrisknoll, the ideal solution is to decompose and decouple the concept set services/components so that these actions are controlled by the containing page vs having these handlers at the global application level. I won't attempt these changes in this PR we should aim to do this for 2.8.

From my side, this PR is good to go and addresses the issue in #1518. Let me know if you have any further concerns.

PR #1562, opened 2019-03-18
Handled subscriptions to the cohort definition

aklochkova:
Fixes https://github.com/OHDSI/WebAPI/issues/973

### DIFF
--- a/js/pages/cohort-definitions/cohort-definition-manager.js
+++ b/js/pages/cohort-definitions/cohort-definition-manager.js
@@ -620,7 +620,7 @@ define(['jquery', 'knockout', 'text!./cohort-definition-manager.html',
 				// reset view after save
 					cohortDefinitionService.deleteCohortDefinition(this.model.currentCohortDefinition().id()).then( (result) => {
 						this.model.currentCohortDefinition(null);
-						document.location = "#/cohortdefinitions"
+					document.location = "#/cohortdefinitions"
 				});
 			}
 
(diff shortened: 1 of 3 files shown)


js/pages/cohort-definitions/cohort-definition-manager.js
```diff
@@ -619,6 +619,10 @@ define(['jquery', 'knockout', 'text!./cohort-definition-manager.html',
 
 				// reset view after save
 					cohortDefinitionService.deleteCohortDefinition(this.model.currentCohortDefinition().id()).then( (result) => {
+						let subscriptions = this.model.cohortDefinitionSubscriptions;
```
chrisknoll: Question: should these disposals be put into a 'dispose()' method of the view model?  It will be called when the component is removed from the view, which should happen as a result of the 'deleteCohort' action.

js/pages/cohort-definitions/components/checks/conceptset-warnings.js
```diff
@@ -11,7 +11,10 @@ define(['knockout', 'text!./conceptset-warnings.html',
     function conceptSetWarnings(params){
       var self = this;
       self.model = params.model;
-      self.cohortDefinitionId = self.model.currentCohortDefinition().id || ko.observable(-1);
+      self.cohortDefinitionId = ko.observable(self.model.currentCohortDefinition().id()) || ko.observable(-1);
+      self.model.cohortDefinitionSubscriptions.push(
+          self.model.currentCohortDefinition.subscribe((definition) => self.cohortDefinitionId(definition ? definition.id() : -1))
```
chrisknoll: the way this looks:  subscribe to model.currentCohortDefinition, on change update self.cohortDefinitionId either the new definitionId or -1.

Couldn't this be handled by:
`
self.cohortDefinitionId = ko.pureComputed(() => self.model.currentCohortDefinition() && self.model.currentCohortDefinition().id() || -1)
`

`

js/providers/Model.js
```diff
@@ -111,6 +111,7 @@ define(
 				this.includedSourcecodes = ko.observableArray();
 				this.cohortDefinitions = ko.observableArray();
 				this.currentCohortDefinition = ko.observable();
+				this.cohortDefinitionSubscriptions = [];
```
chrisknoll: Let's move this field out of Model.js (we're trying to get rid of Model.js completely since it has a heap of different functions that should be refactored out to either services or the components these behaviors belong to).

In this case, this subscription list is being populated by the `conceptset-warning.js` , so let's move this array over to conceptset-warnings.js, and push/dispose these subscriptions when that component disposes.

js/pages/cohort-definitions/components/checks/conceptset-warnings.js
```diff
@@ -92,7 +92,9 @@ define(['knockout', 'text!./conceptset-warnings.html',
         self.runDiagnostics(self.cohortDefinitionId(), expressionJSON);
       };
 
-      self.model.currentCohortDefinition.subscribe(() => self.getWarnings());
+      self.model.cohortDefinitionSubscriptions.push(
```
chrisknoll: Noted below: push subscriptions into a local `subscriptions[]` property.

js/pages/cohort-definitions/cohort-definition-manager.js
```diff
@@ -1169,6 +1169,10 @@ define(['jquery', 'knockout', 'text!./cohort-definition-manager.html',
 
 		// dispose subscriptions / cleanup computed observables (non-pureComputeds)
 			dispose () {
+				let subscriptions = this.model.cohortDefinitionSubscriptions;
```
chrisknoll: move this dispose section over to conceptset-warnings.  The idea is to dispose in the same context that you created.

Issue #1529, opened 2019-03-15
[ATLAS 2.7 TEST] Error Message on Import


ericaVoss:
### Expected behavior
I have bad user behavior on import screen I have a readable error telling me

### Actual behavior
Error reads "[OBJECT OBJECT]"

### Steps to reproduce behavior

1. Go to search --> import 
2. Add this to the search:  
3. It throws "[OBJECT OBJECT]" error

![image](https://user-images.githubusercontent.com/6305439/54459762-0d686280-473e-11e9-8294-869308513a8d.png)

PR #1568, opened 2019-03-19
added error message for case without identifiers

AntonStepanof:


AntonStepanof:
fixes https://github.com/OHDSI/Atlas/issues/1529

Issue #1269, opened 2019-01-25
Role definition in atlas - copy existing roles to new 


gowthamrao:
Would be nice if we can 

- clone/copy existing role profiles to new 
- import/export roles as JSON object across Atlas instance

PR #1574, opened 2019-03-20
added export functionality

johnSamilin:
fixes https://github.com/OHDSI/Atlas/issues/1269

### DIFF
--- a/js/Model.js
+++ b/js/Model.js
@@ -848,16 +848,12 @@ define(
 						console.warn('Roles are not updated');
 						return Promise.resolve();
 					}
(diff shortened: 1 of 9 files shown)


pavgra:
@johnSamilin , now the new name appears but the name input is disabled (cannot edit it)

pavgra:
1. Export existing role
2. Paste its contents into import field
3. Start changing name

image

pavgra:
1. Import role
1. Click on Configuration -> Manage Permissions
1. The new role is not in list (however, it is when you reload page)

PR #1579, opened 2019-03-21
added some messages

### LINKED ISSUE #1130: Cohort Pathways language on design page
### Request
1. Can the length of time for evaluation of the event cohorts be exposed to the user on the treatment pathway design page so that they do not have to return to the target cohort to evaluate the observation time selected? 
2. Under Event Cohorts section, the language introduces a new term which I think should read "Each Event Cohort defines the step in a pathway that may occur for a person in the **Target (not Treatment)** Cohort.
3. Also if the analysis settings options can be more clearly articulated that would be helpful; suggestions below
1. Combination window - "Window of time when two event cohorts need to overlap to be considered a combination." 
2. Minimum cell count - "Minimum number of subjects in the target cohort for a given event in order to be counted in the pathway". Sorry I'm not sure myself what this really means...
3. Maximum path length - "Maximum number of steps in a given pathway to be included in the sunburst plot."

johnSamilin:
fixes #1130

### DIFF
--- a/js/pages/pathways/components/tabs/pathway-design.html
+++ b/js/pages/pathways/components/tabs/pathway-design.html
@@ -12,7 +12,7 @@
 	</div>
 	<div data-bind="css: classes('linked-entities')">
 		<linked-cohort-list params="title: 'Event Cohorts',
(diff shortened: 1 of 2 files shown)


chrisknoll:
Requesting additional information from @jhardin29.

PR #1584, opened 2019-03-22
Fixed ir generation jobs loading

### LINKED ISSUE #1583: IR Generation jobs are not shown
### Expected behavior
IR Generation jobs are shown
### Actual behavior
IR Generation jobs are not shown
1) Open IR
2) Click Generation tab
3) Click generate button
4) Job is shown in table
4) Go to another page (ex. Profiles)
5) Return back to IR
6) Open Generation tab
7) Job is not shown in table

vlbe:
Fixes #1583

### DIFF
--- a/js/pages/incidence-rates/ir-manager.js
+++ b/js/pages/incidence-rates/ir-manager.js
@@ -49,6 +49,7 @@ define([
 	class IRAnalysisManager extends AutoBind(Page) {
 		constructor(params) {
 			super(params);
+			console.log('constru')
 			// polling support
 			this.pollTimeout = null;
 			this.model = params.model;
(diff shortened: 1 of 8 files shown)


js/pages/incidence-rates/ir-manager.js
```diff
@@ -275,6 +275,9 @@ define([
 			const { analysisId } = params;
 			if (analysisId && parseInt(analysisId) !== (this.selectedAnalysis() && this.selectedAnalysis().id())) {
 				this.onAnalysisSelected();
+			} else if (this.selectedAnalysis() && this.selectedAnalysis().id()) {
+				this.pollForInfo();
+				this.pollTimeout = PollService.add(() => this.pollForInfo({ silently: true }), 10000);
```
pavgra: These two lines:
```
this.pollForInfo();
this.pollTimeout = PollService.add(() => this.pollForInfo({ silently: true }), 10000);
```
already exist in `onAnalysisSelected`.

Wouldn't it be better to extract those into a separate function? Or even introduce a new flag in `PollService.add` which would kick off the provided function immediately (so encapsulate the behaviour)

vlbe: @pavgra, agree and i also noticed that in PollService at first we call fn and then wait for another call, so it's no need to call for ex. this.pollForInfo() before adding it to PollService.

PR #1588, opened 2019-03-22
removed paging from export tab

### LINKED ISSUE #1547: [ATLAS 2.7.1] Export Text does not copy full text in 1 click
### Expected Behavior
In Cohort Definition Print Friendly, To copy the print-friendly text of the cohort, the text should able to be copied in full without missing information, such as subsequent pages of concepts in the paged concept set datatble.

### Actual Behavior

If you select-all on the print friendly cohort definition,   the concept sets are only partially copied because the paging of the table is obscuring everything except the first 10 or 25 concepts.  To get all the concepts, the user would have to go to each table, and choose 'show all rows' which is cumbersome.  

In addition, copying all will include the buttons for navigation which isn't what you would put into a 'print friendly form'.

### Steps to Reproduce
1. Open cohort definition with multiple concept sets, where some of the concept sets have > 50 concepts in them.
2. Go to print friendly tab
3. Observe that you are unable to highlight the entire print-friendly text.  Anything shown on page 2 is not visible by default.

### Suggestion
Do not page concept sets in the print-friendly view, and use simple HTML table structures to display data so that it is easily copied-pasted into something like a word document.

johnSamilin:
fixes #1547

### DIFF
--- a/js/components/cohortdefinitionviewer/components/ConceptSetViewer.js
+++ b/js/components/cohortdefinitionviewer/components/ConceptSetViewer.js
@@ -13,6 +13,7 @@ define([
 		constructor(params) {
 			super(params);
 			this.data = params.conceptSet.expression.items;
+			this.paging = params.doPaging === undefined ? true : params.doPaging;
 			this.columns = [
 				{
 					title: 'Concept Id',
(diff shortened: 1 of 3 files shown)


pavgra:
As discussed, this will be implemented as middle-tier service

PR #1589, opened 2019-03-22
tooltips are not closed when changing report type(#1586)

### LINKED ISSUE #1586: tooltips are not closed when changing report type
For data sources

### Expected behavior
Tooltip is closed when the type of report is changed

### Actual behavior
Tooltip is still shown

1. Select report type with treemap
2. Try to select another report type in drop down menu
3. After click on new report type the tooltip appears if the mouse is located on the treemap
4. After the report type changes the tooltip is still visible

ssuvorov-fls:
fixes #1586

### DIFF
--- a/js/components/Chart.js
+++ b/js/components/Chart.js
@@ -31,7 +31,7 @@ define([
     }
 		
 		dispose() {
-			this.renderer && this.renderer.dispose();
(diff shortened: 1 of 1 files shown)


pavgra:
Confirmed that was fixed by https://github.com/OHDSI/Visualizations/pull/42

PR #1590, opened 2019-03-22
Removed component-specific stuff from Model.js

vlbe:
First step to remove model.js

### DIFF
--- a/js/pages/configuration/roles/role-details.js
+++ b/js/pages/configuration/roles/role-details.js
@@ -1,333 +1,329 @@
 define([
-    'knockout',
-    'text!./role-details.html',
-    'components/Component',
-    'utils/AutoBind',
-    'utils/CommonUtils',
-    'services/role',
-    'lodash',
-    'assets/ohdsi.util',
-    'services/User',
-    'services/AuthAPI',
-    'databindings',
-    'components/ac-access-denied',
-    'less!./role-details.less',
-    'components/heading'
-], function (
-    ko,
-    view,
-    Component,
-    AutoBind,
-    commonUtils,
-    roleService,
-    _,
-    ohdsiUtils,
-    userService,
-    authApi,
-) {
-    const defaultRoleName = "New Role";
-    class RoleDetails extends AutoBind(Component) {
-        constructor(params) {
-            super(params);        
+  'knockout',
+  'text!./role-details.html',
+  'components/Component',
+  'utils/AutoBind',
+  'utils/CommonUtils',
+  'services/role',
+  'lodash',
+  'assets/ohdsi.util',
+  'services/User',
+  'services/AuthAPI',
+  'atlas-state',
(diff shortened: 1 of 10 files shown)


chrisknoll:
Great work. This is a very sensitive/fragile class so we'll have to make sure that we have very deep testing here because the methods used in Model.js are not obviously linked to any specific feature in Atlas...but once we get this squared away, it will be much cleaner.

js/pages/jobs/job-manager.js
```diff
@@ -1,66 +1,72 @@
...
+      this.jobs = ko.observableArray([]);
+      this.columns = ko.observableArray([
+        { title: 'ExecutionId', data: 'executionId' },
+        { title: 'Job Name', data: 'jobParameters.jobName' },
+        { title: 'Status', data: 'status' },
+        { title: 'Start Date', data: 'startDate', type: 'datetime-formatted' },
+        { title: 'End Date', data: 'endDate', type: 'datetime-formatted' },
+      ]);
+      if (config.userAuthenticationEnabled) {
+        this.columns.splice(3, 0, {
```
pavgra: Let's remove magic values (either create constants or calculate the numbers)

js/pages/configuration/roles/role-details.js
```diff
@@ -1,333 +1,329 @@
...
+          ? authApi.isPermittedCreateRole()
+          : authApi.isPermittedReadRole(this.roleId());
+      });
+      this.canEditRole = ko.pureComputed(() => {
+        return this.isAuthenticated() && this.isNewRole()
+          ? authApi.isPermittedCreateRole()
+          : authApi.isPermittedEditRole(this.roleId());
+      });
+      this.canEditRoleUsers = ko.pureComputed(() => {
+        return (
```
pavgra: Let's not mix refactoring and code formatting changes - hard to review

js/pages/configuration/routes.js
```diff
@@ -27,7 +29,7 @@ define(
         '/role/:id': new AuthorizedRoute((id) => {
           appModel.activePage(this.title);
           require(['./roles/role-details'], function () {
-            appModel.currentRoleId(id);
+            atlasState.ConfigurationRole.selectedId(id !== 0 ? id : null)
```
pavgra: Are you sure that the zero is numeric, not string here? I would add `parseInt`

js/pages/configuration/sources/source-manager.js
```diff
@@ -300,62 +274,78 @@ define([
...
-            return roleService.getList();
-        })
-        .then((roles) => {
-          this.model.roles(roles);
-          this.loading(false);
-          this.goToConfigure();
-        })
-        .catch(() => this.loading(false));
+
+      try {
```
pavgra: Let's not mix refactoring and code syntax changes - hard to review

Issue #1594, opened 2019-03-22
Configuration screen not refreshed after granted permission for new Datasource


spbima:
### Expected behavior
In Configuration screen have displayed vocabularies info correctly

### Actual behavior
Configuration screen had problems with showing vocabularies info. 
For new Datasource after granted permissions list has not refreshed and has displayed "unavailable" for vocab version (had to hard refresh in browser)

### Steps to reproduce behavior
1. Create new Datasource
2. Grant permissions
3. Configuration screen displayed "unavailable" for vocab info

PR #1597, opened 2019-03-26
Issue 1594 refresh configuration screen

AntonStepanof:
fixes https://github.com/OHDSI/Atlas/issues/1594

### DIFF
--- a/js/pages/configuration/configuration.js
+++ b/js/pages/configuration/configuration.js
@@ -65,7 +65,6 @@ define([
     }
 
     async onPageCreated() {
-      await authApi.loadUserInfo();
       sourceApi.initSourcesConfig();
       super.onPageCreated();
     }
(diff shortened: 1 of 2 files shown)


js/pages/configuration/configuration.js
```diff
@@ -61,7 +61,8 @@ define([
         }
       });
       
-		  this.canImport = ko.pureComputed(() => this.isAuthenticated() && authApi.isPermittedImportUsers());
+      this.canImport = ko.pureComputed(() => this.isAuthenticated() && authApi.isPermittedImportUsers());
+      sourceApi.initSourcesConfig();
```
pavgra: Data loading should take place in `onPageCreated` / `onRouterParamsChanged`

AntonStepanof: updated.

AntonStepanof: Also I added small change in SourceApi. I removed default value at source.version. This change prevent from blink during initialization.

Issue #1591, opened 2019-03-22
New Dataset screen is not returned to Dataset list


spbima:
This issue happens on a clean instance. When there are no data sources at all and you are configuring the first one.
Exactly all is working for you after configuration is done.

### Expected behavior
New Dataset screen is returned to Dataset list

### Actual behavior
New Dataset screen is not returned to Dataset list after you click Save (had to hard refresh in browser)

### Steps to reproduce behavior
1. You need clean instance
2. Press new Datasource
3. Input data and press Save

AntonStepanof:
It will be fixed at https://github.com/OHDSI/Atlas/issues/1593

pavgra:
#1593 was closed so pls cherry-pick what you need and create a PR

PR #1608, opened 2019-03-27
added check for vocabularyUrl

AntonStepanof:
fixes https://github.com/OHDSI/Atlas/issues/1591


wivern:
It seems this is covered by #1605

AntonStepanof:
ok, I will close that

Issue #1592, opened 2019-03-22
Roles screen is not updated with new role for dataset


spbima:
This issue happens on a clean instance. When there are no data sources at all and you are configuring the first one.
Exactly all is working for you after configuration is done.

### Expected behavior
Roles screen is updated with a new role after the dataset created

### Actual behavior
Roles screen is not updated with a new role for the dataset. It looks list is taken from cache (had to hard refresh in a browser)

### Steps to reproduce behavior

1. You need clean instance
2. Create Datasource
3. Go to Role list
4. Role for new Datasource had not displayed

PR #1614, opened 2019-03-28
Issue 1591 fix problem during return to datasources(pr from closed ticket 1593)

AntonStepanof:
fixes https://github.com/OHDSI/Atlas/issues/1591
fixes https://github.com/OHDSI/Atlas/issues/1592


wivern:
@AntonStepanof  I guess this should be to the v2.7.1 branch

Issue #1620, opened 2019-03-29
Doesn't display expiration time for a session for IAP authorization


spbima:
### Expected behavior
Expiration time for a session for IAP authorization displayed or expiration time removed (doesn't show _exp: null_)

### Actual behavior
Doesn't display expiration time for a session for IAP authorization

![iap-29032019](https://user-images.githubusercontent.com/804934/55224368-6ac1e080-5221-11e9-801a-b6c1e1e11d65.jpg)

### Steps to reproduce behavior
Login to Atlas with IAP authorization

PR #1622, opened 2019-03-29
Issue 1620 doesnt display null exp time

AntonStepanof:
fixes https://github.com/OHDSI/Atlas/issues/1620


chrisknoll:

PR #1623, opened 2019-03-29
Added loader after selecting concept set

### LINKED ISSUE #1615: Block  re-selection posibility after choise
### Expected behavior
Can see "pre-loader" or anything else to block possibility choose something from the table

### Actual behavior
Can click and choose some elements after 1st click
If I click several times in concept set Every click add a row in Concept Set tab.

1. Create Concept set with elements
2. Create Cohort Definition
3. Add initial Event
4. Add condition era( Import concept set)
5. Click several times on the row with concept set before windows closing

vlbe:
Fixes #1615


pavgra:
I believe, there is a similar case in `cohort-definition-manager.js` -> `onConceptSetRepositoryImport (newConceptSet)`. I would address them both.

PR #1632, opened 2019-04-01
Issue 1547 update export text view cohort definition

AntonStepanof:
fixes https://github.com/OHDSI/Atlas/issues/1547

### DIFF
--- a/js/utils/Clipboard.js
+++ b/js/utils/Clipboard.js
@@ -23,7 +23,43 @@ define(['clipboard'], (clipboard) => {
 				console.log(e);
 			});
 		}
-	};
+
+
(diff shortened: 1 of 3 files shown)


js/utils/Clipboard.js
```diff
@@ -23,7 +23,43 @@ define(['clipboard'], (clipboard) => {
 				console.log(e);
 			});
 		}
-	};
+
+
+        copyToClipboard(clipboardButtonId, clipboardButtonMessageId, appendedText) {
```
pavgra: Why did you leave previous `copyToClipboard` implementation? How could it be used? This is not Java

chrisknoll:
@pavgra, please flush the toilet.

PR #1643, opened 2019-04-03
Fixes creation of new Datasource with keyfile

wivern:


chrisknoll:

Issue #1526, opened 2019-03-15
[ATLAS 2.7 TEST] Hitting the all shopping cart icon changes show entries


ericaVoss:
This is low priority . . . 

### Expected behavior
When you hit the shopping cart to select all concepts, it changes the show entries

![image](https://user-images.githubusercontent.com/6305439/54458286-e27c0f80-4739-11e9-83ea-921cb6c9045b.png)

### Actual behavior
the shopping cart selection does not refocus the screen.

### Steps to reproduce behavior

1. select concept 434592
2. go to the "related concepts" tab
3. Show 100 entries
4. hit the shopping cart to select all
5. the entries will go back to 15

![bitmoji](https://render.bitstrips.com/v2/cpanel/c8260099-f50c-4153-a292-6abae910deb2-29ae54b2-6f50-4175-8e50-852ea13b047d-v1.png?transparent=1&palette=1&width=246)

Issue #1581, opened 2019-03-21
Facet-Datatable performance/ bugs in concept sets


chrisknoll:
### Expected behavior
1.  When using facet filter (on left) to limit rows to a subset, after clicking 'select all' at the top, the filtered rows are properly selected, and existing filters are not reset.

2. Un-selecting a filter in the facet should return to the un-filtered rows quickly.

### Actual behavior
1.  When using facet filter (on left) to limit rows to a subset, after clicking 'select all' at the top, the filtered rows are properly selected, however, the filter is reset, and all rows are returned.

2. With a large table, reloading all rows is very slow.

### Steps to reproduce behavior
Using the following concept set expression:
[Issue_1581.txt](https://github.com/OHDSI/Atlas/files/2992951/Issue_1581.txt)

This has ~200 concepts (which is reasonable) but, resolves to ~45k rows.  We should profile what loading the table with data is bottle-necking at and see if we can adjust our approach.

Notes: it looks like XSS filtering is being applied to every row, and in this type of data, XSS is not a concern, so we could not apply XSS filtering to this table.

PR #1652, opened 2019-04-05
Select all concepts fix

anthonysena:
Per #1526, when using the "select all" on a faceted data table, the concepts are properly selected but the table's state is reset wiping out any user-defined facet filters, options, etc. To accomplish this, I've removed the call to `ko.contextFor(this).$component.reference.valueHasMutated()` since this call causes the entire data table to be re-drawn which is inefficient. 

Instead, I've reused the `table` variable that is declared earlier in the function to obtain the concepts that exist in the data table with all of the filters/options applied. The loop below:

```
for (var i = 0; i < table.rows()[0].length; i++) {
	table.cell(i,0).data('<i class="fa fa-shopping-cart"></i>');
}
```

What this is doing is obtaining the length of the filtered data: `table.rows()[0].length` and then applies the "selected shopping cart" to each row in the table: `table.cell(i,0).data('<i class="fa fa-shopping-cart"></i>');`

### DIFF
--- a/js/Application.js
+++ b/js/Application.js
@@ -122,8 +122,9 @@ define(
 							}
 						}
 						sharedState.selectedConcepts(selectedConcepts);
-						ko.contextFor(this)
-							.$component.reference.valueHasMutated();
+						for (var i = 0; i < table.rows()[0].length; i++) {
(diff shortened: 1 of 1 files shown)


anthonysena:
This PR also addresses portion of #1581.

PR #1653, opened 2019-04-05
Vocabulary search - handle special characters

### LINKED ISSUE #250: Vocabulary Search doesn't honor special characters
Case: search for ICD10: E78.5 does not return results due to the '.' character.

anthonysena:
Fixes #250 through proper encoding and a work-around for the "/". The way search currently work is by redirecting to a new URL:

`atlas/#/search/<search term>`

If your `<search term>` contains a "/" it messes up the route handling since it can't determine if the "/" is part of the URL or instead is part of the search term. I've added a work-around that eliminates the "/" from the search itself and replaces it with a space. This is a hack for sure but is a quick fix and takes advantage of the current search which will insert a wildcard for spaces.

### DIFF
--- a/js/pages/vocabulary/components/search.js
+++ b/js/pages/vocabulary/components/search.js
@@ -190,9 +190,13 @@ define([
 				this.getVocabularies();
 			}
 		}
+		
+		encodeSearchString(searchTerm) {
+			return encodeURIComponent(searchTerm.replace("/", " "));
+		}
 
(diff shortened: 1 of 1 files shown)


pavgra:
@anthonysena , pressing enter and clicking the button gives a bit different outcome - in the first case I see `ML DG` in search input, in second - `ML/DG`

PR #1669, opened 2019-04-10
Inclusion criteria for Location region

pavgra:
To support recently approved `location_region_id` column
https://github.com/OHDSI/CommonDataModel/issues/252


chrisknoll:
Is this a CDM v5.3 or 6.0 struct?

If it's 6.0, I believe we're going to want to split circe into a 1.x line and a 2.x line for the different cdm support. Or, we can come up with a way to handle either/or.

js/components/cohortbuilder/CohortExpression.js
```diff
@@ -29,6 +29,9 @@ define(function (require, exports) {
 		}));
 		self.CollapseSettings = {CollapseType: ko.observable(data.CollapseSettings && data.CollapseSettings.CollapseType || "ERA"), EraPad: ko.observable(data.CollapseSettings && data.CollapseSettings.EraPad || 0 ) }
 		self.CensorWindow = ko.observable(new Period(data.CensorWindow));
+
+		self.minCdmVersion = data.minCdmVersion || null;
```
chrisknoll: I would like this to work like any other semVer expression described here:
https://www.telerik.com/blogs/the-mystical-magical-semver-ranges-used-by-npm-bower

ie:
>=5.3 < 6.0 
or
5.3-6.0  

We'd have to provide a UI component that will help users specify this value.

This is described in this issue:
https://github.com/OHDSI/circe-be/issues/27

So before rushing ahead to provide your own solution to handling CDM versions, please let's discuss this on a WG call.

Issue #1542, opened 2019-03-16
[ATLAS 2.7 TEST] Cohort Pathways Search Shows Nothing


clairblacketer:
### Expected behavior
When searching on the cohort pathways page for either target or event cohorts the list would filter based on the search term

### Actual behavior
If a search term is entered it automatically shows no matching records

### Steps to reproduce behavior
Enter a search term in the search box for either target or event cohorts in a cohort pathway

PR #1676, opened 2019-04-15
update search for editable name field

AntonStepanof:
fixes https://github.com/OHDSI/Atlas/issues/1542

### DIFF
--- a/js/extensions/bindings/datatableBinding.js
+++ b/js/extensions/bindings/datatableBinding.js
@@ -213,7 +213,13 @@ define([
 			jQuery.fn.dataTableExt.search.pop();
 			jQuery.fn.dataTableExt.search.push(
 				function (settings, data, dataIndex) {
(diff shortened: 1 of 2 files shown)


js/extensions/bindings/datatableBinding.js
```diff
@@ -209,8 +209,24 @@ define([
 			// drawing may access observables, which updating we do not want to trigger a redraw to the table
 			// see: https://knockoutjs.com/documentation/computed-dependency-tracking.html#IgnoringDependencies
 			ko.ignoreDependencies(table.draw);
+
+			jQuery.fn.dataTableExt.search.pop();
```
pavgra: The table shouldn't know anything about its contents. Seems like very specific and tightly-coupled solution. I would look into `HTML5 data-* attributes` or mechanism which stays behind it:
https://datatables.net/examples/advanced_init/html5-data-attributes.html

PR #1678, opened 2019-04-15
Created mechanizm for exporting all reports from cohort characterization

### LINKED ISSUE #1513: cohort characterization export - provide way to export all domains vs. having to export individually - enhancement request
### Enhancement request
Instead of clicking each analysis performed could an export all feature be included? This would avoid having to repeatedly click for each feature. 

Thank you!

ssuvorov-fls:
fixes #1513


js/pages/characterizations/components/characterizations/characterization-view-edit/characterization-results.js
```diff
@@ -80,8 +84,13 @@ define([
             this.data = ko.observable([]);
             this.domains = ko.observableArray();
             this.filterList = ko.observableArray([]);
+            this.selectedItems = ko.computed(() => {
+                filterUtils.getSelectedFilterValues(this.filterList());
```
wivern: This arrow function always returns nothing. Is this a mistake or such behavior by design?

ssuvorov-fls: fixed

ssuvorov-fls:
wait icon was added

Issue #1249, opened 2019-01-23
Component disposal promoted from Page class


anthonysena:
From a review with @chrisknoll, I'd like to propose promoting the `dispose()` method and `subscription` property from the `js/page/Page.js` class to the `js/components/Component.js` class, All components that extend `Component` should have a dispose() method that is declared with a base implementation that performs the cleanup. Since `Page` extends `Component`, it could then use the inherited `subscription` property as it does now.

Tagging @wivern @pavgra for their thoughts.

chrisknoll:
I like this since it gives us a consistent 'subscription disposal' construct that any component can leverage, without having to come up with multiple implementations on how to track the component's subscription.

pavgra:
@anthonysena , due to missed deadlines I'd propose to move the issue into 2.8 milestone

PR #1696, opened 2019-04-23
Issue 1249 promote disposal from page

AntonStepanof:
fixes https://github.com/OHDSI/Atlas/issues/1249


AntonStepanof:
> Why is this mechanism not utilized everywhere? E.g. I'm looking into `data-sources.js` and observing own `dispose` method while you could just replace
> 
> ```
> this.selectedReportSubscription = this.selectedReport.subscribe(r => {
>     this.updateLocation();
> });
> ...
> ```
> with
> 
> ```
> this.subscriptions.push(this.selectedReport.subscribe(r => this.updateLocation()));
> ...
> ```
> and drop the DataSource's `dispose` method

Updated dispose method for another cases

PR #1699, opened 2019-04-23
Reverted extra changes back

aklochkova:
Fixes https://github.com/OHDSI/Atlas/issues/1338

this PR is only for fixing 'master' branch


aklochkova:
@pavgra hi! because that code changes were my own initiative and they aren't relative to this task. and there is no code style convention, so i decided to revert all of them back.

PR #1700, opened 2019-04-24
Fixed Cohort Definition page and CC results

aklochkova:
Fixes https://github.com/OHDSI/Atlas/issues/1338

### DIFF
--- a/js/Model.js
+++ b/js/Model.js
@@ -508,6 +508,7 @@ define(
 					this.currentCohortDefinitionInfo([]);
 				} else {
 					const { data: cohortDefinition } = await httpService.doGet(config.api.url + 'cohortdefinition/' + cohortDefinitionId);
+					cohortDefinition.expression = JSON.parse(cohortDefinition.expressionStr);
 					this.currentCohortDefinition(new CohortDefinition(cohortDefinition));
 
 					const { data: generationInfo } = await httpService.doGet(config.api.url + 'cohortdefinition/' + cohortDefinitionId + '/info');					
--- a/js/Model.js
+++ b/js/Model.js
@@ -508,7 +508,6 @@ define(
 					this.currentCohortDefinitionInfo([]);
 				} else {
 					const { data: cohortDefinition } = await httpService.doGet(config.api.url + 'cohortdefinition/' + cohortDefinitionId);
-					cohortDefinition.expression = JSON.parse(cohortDefinition.expressionStr);
 					this.currentCohortDefinition(new CohortDefinition(cohortDefinition));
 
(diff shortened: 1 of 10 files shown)


js/Model.js
```diff
@@ -508,6 +508,7 @@ define(
 					this.currentCohortDefinitionInfo([]);
 				} else {
 					const { data: cohortDefinition } = await httpService.doGet(config.api.url + 'cohortdefinition/' + cohortDefinitionId);
+					cohortDefinition.expression = JSON.parse(cohortDefinition.expressionStr);
```
pavgra: 1. Would be reasonable to encapsulate it in a service method
2. Now you have two methods for getting the `expression` and no description (comments) is provided what is the difference between those

aklochkova: I removed this line because now cohort expression comes to UI already as Object

aklochkova:
Cohort Definition screen - tested
image

Cohort Characterization results - tested
image

pavgra:
Please give a meaningful name to the PR

aklochkova:
now saving of Cohort Def is working and Criteria params are displayed

image
image

pavgra:
1. Create a new cohort
2. Add Visit criteria
image
3. Save
4. Observe changed state
image

aklochkova:
@pavgra now it isn't displayed after save

js/components/cohortbuilder/components/VisitOccurrenceTemplate.html
```diff
@@ -101,7 +101,7 @@
 					</div>
 				</td>
 			</tr>
-			<tr data-bind="if: PlaceOfServiceLocation() !== undefined, visible: PlaceOfServiceLocation() !== undefined">
+			<tr data-bind="if: PlaceOfServiceLocation() !== null, visible: PlaceOfServiceLocation() !== null">
```
chrisknoll: could you try just:
```
if: PlaceOfServiceLocation, visible: PlaceOfServiceLocation
```

The bindings will automatically unwrap for you, and it will check the 'falsy' value of the binding. that way you don't have any confusion between undefiend or null, since those are both falsy.

pavgra:
@aklochkova , now it isn't displayed _at all_

js/components/cohortbuilder/components/VisitOccurrenceTemplate.html
aklochkova: @chrisknoll i just tried, it doesn't work

pavgra: @chrisknoll , the issue here is in data supplied from WebAPI, UI is not the place to fix it. Previously, WebAPI didn't return nulls for empty filters, but later @aklochkova unintentionally changed that to do so. Therefore, the distinction between undefined and null was lost

chrisknoll:

PR #1707, opened 2019-04-30
Implements RunAs functionality

wivern:


js/components/welcome.js
```diff
@@ -139,6 +141,23 @@ define([
             });
         };
 
+        self.runAs = function() {
+          self.isInProgress(true);
+          $.ajax({
```
pavgra: 1. Move to service
2. Use `httpService` instead of jQuery

wivern: 1. runAs moved to AuthAPI
1. unfortunately `httpService` does not support `x-www-form-urlencoded` but only `multipart/form-data`. Multipart might be a good solution but `httpRequest.getParts` returns 0 parts when called inside Shiro filter.

PR #1717, opened 2019-05-14
Fixed creation of AD sync job

### LINKED ISSUE #1716: Not possible to create AD sync job
### Expected behavior
It should be possible to create AD sync job

### Actual behavior
It's not possible to save entity as Execute dropdown is empty

image

vlbe:
Fixes #1716


package.json
```diff
@@ -48,7 +48,6 @@
     "@ohdsi/visibilityjs": "^2.0.2",
     "ajv": "^6.10.0",
     "bootstrap": "^3.3.7",
-    "bootstrap-datetimepicker": "^0.0.7",
```
chrisknoll: What was the reason to switch from npm-hosted bootstrap-datetimepicker to a local copy?

chrisknoll:
We currently have a date-picker implementation for other components in Atlas: https://github.com/OHDSI/Atlas/blob/master/js/extensions/bindings/datepickerBinding.js, is this adding a different implementation?  if so, is it wise to have different components that perform the same task?

vlbe:
@chrisknoll No, implementation remains the same. This datetimepicker component uses this library:
https://eonasdan.github.io/bootstrap-datetimepicker/
And we have local copy of this library. Then in https://github.com/OHDSI/Atlas/pull/1095 we switched to npmjs hosted library but the problem is that it is not the same lib that we used(just the same name). Actually i managed to find npmjs hosted lib that we need - https://www.npmjs.com/package/eonasdan-bootstrap-datetimepicker
I'll update PR. Thanks.

chrisknoll:
@vlbe, I'm sorry I don't think I'm qualified to give this a complete review: I don't have any AD sync jobs to test this with or an environment that would let me access those screens.  I just wanted to point out the NPM library issue, and also question if we had a redundant datetimepicker component (I believe we do, in this case.  The other places in the code uses a jquery-ui implementation of a date picker, and this one appears to be a bootstrap one; maybe as a separate effort we can unify this, but no need to change out the implementation for this PR since it's not related).

PR #1760, opened 2019-05-22
Delete concept set - fixes #1750 - master

### LINKED ISSUE #1750: Concept set holding stale references
### Expected behavior

Deleting a concept set removes it from the system and the interface.

### Actual behavior

Deleting a concept set removes it from the system but does not clear it from the interface display.

1. Open a saved concept set. Note the included concepts.
2. Delete the concept set.
3. Use the vocabulary to search for a term that was not in the previously deleted concept set
4. Import the concept into the concept set
5. Navigate to the concept set display

You will then see the previous concept set's data along with the newly imported concept.

anthonysena:
Uses the model's clearConceptSet() method after deletion to remove stale references to fix #1750.

fixes #1750


chrisknoll:
Delete branch?

PR #1765, opened 2019-05-22
Fixed security issues with rendering in datatables

vlbe:


pavgra:
https://datatables.net/manual/data/renderers

> The text helper will ensure that any potentially dangerous HTML in the source data will not be executed by escaping the HTML entities. This can be useful if the data being loaded may come from a potentially untrusted data source and can help mitigate XSS attacks

Issue #1733, opened 2019-05-20
Feedback screen is not shown after logout


ssuvorov-fls:
### Expected behavior
Feedback page should be seen after logout

### Actual behavior
Feedback page is not seen. It can be seen after page reload or navigating to another page and returning back to feedback page

### Steps to reproduce behavior
1. Log in
2. Navigate to feedback page
2. Log out

anemniugin:
The same issue in home page

PR #1767, opened 2019-05-23
added check for feedback screen for logout

AntonStepanof:
fixes https://github.com/OHDSI/Atlas/issues/1733

### DIFF
--- a/js/pages/Router.js
+++ b/js/pages/Router.js
@@ -78,7 +78,7 @@ define(
         }, {});
         // anyway, we should track the moment when the user exits and check permissions once again
 				authApi.isAuthenticated.subscribe((isAuthenticated) => {
-					if (!isAuthenticated) {
+					if (!isAuthenticated && window.location.hash !== '#/feedback') {
             this.setCurrentView('white-page');
 						this.schedulePageUpdateOnLogin(this.activeRouteHandler);
 					}
--- a/js/pages/Router.js
+++ b/js/pages/Router.js
@@ -14,7 +14,7 @@ define(
 	) => {
     return class AtlasRouter {
       constructor() {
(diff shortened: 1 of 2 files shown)


js/pages/Router.js
```diff
@@ -78,7 +78,7 @@ define(
         }, {});
         // anyway, we should track the moment when the user exits and check permissions once again
 				authApi.isAuthenticated.subscribe((isAuthenticated) => {
-					if (!isAuthenticated) {
+					if (!isAuthenticated && window.location.hash !== '#/feedback') {
```
vlbe: I think it is better to use this.getModel().activePage() instead of windows.location.hash
+ it also needs to be done for Home page.

AntonStepanof: updated

js/pages/Router.js
```diff
@@ -78,7 +78,7 @@ define(
         }, {});
         // anyway, we should track the moment when the user exits and check permissions once again
 				authApi.isAuthenticated.subscribe((isAuthenticated) => {
-					if (!isAuthenticated) {
+					if (!isAuthenticated && this.getModel().activePage() !== 'Feedback' && this.getModel().activePage() !== 'Home') {
```
pavgra: const NON_SECURE_PAGES = ['Home', 'Feedback'];
NON_SECURE_PAGES.includes(this.getModel().activePage())

AntonStepanof:
> You are duplicating data: you should determine which Routes do not require authentication and don't store the list (which would require maintenance). Probably smth like the following:
> 
> ```
> const nonSecureRoutePaths = Object.keys(routes).filter(path => !(routes[path] instanceof AuthorizedRoute));
> ```

updated

Issue #1740, opened 2019-05-21
IR execution is allowed on disabled datasources


konstjar:
### Expected behavior
Generate button is disabled

### Actual behavior
Generate button is active

### Steps to reproduce behavior
Remove all users from Source user role, open IR
![image](https://user-images.githubusercontent.com/4144173/58083622-04729200-7bc2-11e9-8798-dd34b70515ff.png)

PR #1774, opened 2019-05-23
Disabled 'Rerun' button and viewing reports

aklochkova:
Fixes https://github.com/OHDSI/Atlas/issues/1740


pavgra:
Is there an appropriate check at middle-tier?

PR #1792, opened 2019-05-29
webapi-1110: generate version tags in UI for CC execution

acumarav:
Closes OHDSI/WebAPI#1110

### DIFF
--- a/js/pages/characterizations/services/CharacterizationService.js
+++ b/js/pages/characterizations/services/CharacterizationService.js
@@ -1,9 +1,11 @@
 define([
     'services/http',
     'appConfig',
+    'utils/ExecutionUtils'
 ], function (
     httpService,
     config,
+    executionUtils,
 ) {
     function loadCharacterizationList() {
         return httpService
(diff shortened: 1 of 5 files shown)


acumarav:
I have opened PR in Atlas:
image
https://github.com/OHDSI/Atlas/pull/1792

I changed failed executions to plain text as click on a link cause 500 error and open empty popup

js/pages/characterizations/services/CharacterizationService.js
```diff
@@ -44,7 +44,16 @@ define([
     function loadCharacterizationExecutionList(id) {
         return httpService
             .doGet(config.webAPIRoot + 'cohort-characterization/' + id + '/generation')
-            .then(res => res.data);
+            .then(res => generateVersionTags(res.data));
+    }
+
+
+    function generateVersionTags(sourceInfoList) {
```
wivern: Param name is confusing, smth like `ccGenerations` would be better.
Actually it isn't sources list.

js/pages/characterizations/services/CharacterizationService.js
```diff
@@ -44,7 +44,16 @@ define([
...
+        const sortedIds = sourceInfoList.map(info => info.id).sort();
```
wivern: Probably sort descending by date is more suitable, since look what I got below:
88bfd983-86f8-4061-9f39-b41991535579

js/pages/characterizations/services/CharacterizationService.js
acumarav: done

pavgra:
Could you also fix the issue with plain text instead of a link in case of Failed analyses?

js/pages/characterizations/services/CharacterizationService.js
acumarav: Fixed - version generated according to distinct hashCode index

PR #1794, opened 2019-05-29
Added sort column to cohort definition list to improve performance. 

### LINKED ISSUE #1571: Slow Redraw/Sort speed with large data table
### Expected behavior
Sorting column of large datatable should be quick.

### Actual behavior
On a cohort definition list with ~7000 entries, takes about 2 seconds to sort by 'last updated date'.

1. Open cohort definition list with large number of elements. (> 5,000)
2. Sort by Last Updated
3. Observe delay  in sort.

### Additional information

I created a quick 5 second Chrome profile that captures a sort event:
image

It appears most of the energy is spent in the `datetime-formattted-desc` column. If you go to http://www.ohdsi.org/web/atlas/#/cohortdefinitions, that list has about 3000 entries in it, and the resize, sorting is instantaneous. Taking a profile doesn't show any obvious spikes, but it doesn't appear to do any heavy sorting.

Looking deeper in `datetime-formatted-desc`, a lot of time (relatively speaking) is spent inside moment.js
- configFromStringAndFormat
- unescapeFormat

### Implementation Suggestion

Instead of trying to parse out the dates and sort them (which could be an O(N^2) cost) we should have an integer representation of the date field (such as ms from epoc) that is more easily sorted, and we configure DataTables to use this special sort column for those date columns.

chrisknoll:
2 Issues addressed:
1. Sorting a large table on date resulted in excessive date parsing and comparison.  The solution adds a new 'timestamp' column that is pre-calculated as an integer so sorting can be performed without parsing.

2. Added css classes to fix the created/updated and author columns.  This uses a fixed column width.  The reason this felt necessary was due to the following behavior:

Without fixed columns:
before

With fixed Columns:
after

Fixes #1571.

### DIFF
--- a/js/components/cohort-definition-browser.html
+++ b/js/components/cohort-definition-browser.html
@@ -1,4 +1,4 @@
-<faceted-datatable data-bind="visible:loading()==false" params="orderColumn: 3, reference:reference, columns: columns, options:options, rowClick:rowClick">
+<faceted-datatable data-bind="visible:loading()==false" params="orderColumn: 3, autoWidth: false, reference:reference, columns: columns, options:options, rowClick:rowClick">
 </faceted-datatable>
 
 <loading data-bind="visible:loading()" params="status: 'retrieving cohort definitions'"></loading>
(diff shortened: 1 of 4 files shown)


chrisknoll:
Side note: this is intended for 2.8, but would be useful to have for 2.7.2.  There's been a bit of groaning and teeth-gnashing when pulling up the cohort definition list in our internal env, and this would eliminate the pause when loading the table.

PR #1829, opened 2019-06-04
Extracted Cohort Definition's reports to separate components and utilized Tab component

pavgra:
Decouple reports from cohort manager and use existing Tabs component to reduce code duplication


js/pages/cohort-definitions/components/reporting/inclusion-reports/inclusion-reports.js
```diff
@@ -22,16 +24,18 @@ define([
...
+						cohortId: ko.computed(() => params.cohort().id()),
+						reportType: constants.INCLUSION_REPORT.BY_PERSON,
 					},
 				},
 				{
 					title: 'By Events',
 					componentName: 'feasibility-report-viewer-with-header',
 					componentParams: {
-						sourceName: params.sourceName,
-						report: ko.computed(() => params.report() && params.report().byEvent)
```
chrisknoll: This is a strange shift:  the component for viewing the report takes a report data structure to the component. It doesn't care about the source (aside from using it as a report title, which could have been put intot he report object) and it doesn't care about the cohort Id (also, could be some meta about the cohort int he report object such as the name).

The architecture of these 'render' components has always been to be detached from the mechanism that loads the data.  Ie: you wouldn't put a service call to some 'scatterplot loader function' inside of the scatterplot renderer:  you just pass in the data that you want to render.  These report-type components should work the same, with the 'owning' component being the one to decide how to pipe the data into the component.  Separation of concerns.

chrisknoll:
I requested changes before, but dismissed after I re-reviewed, but then there was subsequent changes, so I'll just ask here:

How is the generation results view signaling the report viewer to load a different report from a different source?  The old way separated the concerns with loading the report data from rendering the report data. I think you are still maintaining the report viewer, but it's not exactly clear to me what mechanism is signaling the `loadReport()`. I _think_ you're depending on toggele creating/destroying the component and then the constructor gets fired when the component is re-initialized?

pavgra:
@chrisknoll , nice catch! I've pushed fix

js/pages/cohort-definitions/components/reporting/cohort-reports/feasibility-report-viewer-with-header.js
```diff
@@ -23,6 +23,8 @@ define([
 			this.report = ko.observable();
 
 			this.loadReport();
+			this.source.subscribe(s => s && this.loadReport());
```
chrisknoll: If source and cohort both change, this will cause the load to trigger twice.  

What bug are we fixing by changing this behavior?  Refactoring into smaller components is one thing, but I'm missing the point of this particular design change.

pavgra: What do you mean by `particular design change`?

pavgra: > If source and cohort both change

Currently, there is no such use case. I even do not know how to switch cohort w/o the component disposal

chrisknoll: > What do you mean by particular design change?

Discussed on the call, but essentially it's the difference between the 'imperative' form of a user action directly leads to loading data and assigning a report value, vs an 'inversion of control' form where user action leads to updating a source or cohort definition update which then through a network of subscriptions lads to the 'smart component' to reload the data and render.

You are correct that the element of the `feasibility-report-viewer` taking the report structure and rendering the results was not changed, and the primary change of focus in this PR is to 'wrap' the component render into a 'controller' component which will handle the responsibility of responding to model changes (such as the selected cohort or the selected source observable).

chrisknoll: > Currently, there is no such use case. I even do not know how to switch cohort w/o the component disposal

I believe if you change the cohortID in the URL so that the route changes, it will load a new cohort definition.  I'm not sure if the routing handles drilling down through to the inclusion impact report, but if it did, then you could change the source and cohort via a URL change that will not result in a component dispose.  Even if it doesn't work that way today, it's probably a future bugfix that the routes are handled this way, so we should expect that it will work that way in the future.

To that end:  I'd create a single observable object that contains the cohort and the selected source, and have a single subscriber to that composite datastructure.  I thin this can be defined as a computedObservable that will return the object with the cohort + source, and if either of those underlying objects change, the subscription will be notified once (I believe that's how a computed observable would be notified with 2 dependent observables).

pavgra: >  if either of those underlying objects change, the subscription will be notified once 

How would those change at the same moment of time? If they don't, how would you transform two events separated in time w/o `debounce` which doesn't seem to be a good solution?

And the overall question: do you want to address this topic now (not having a concrete case)?

pavgra: @chrisknoll ,

>   thin this can be defined as a computedObservable that will return the object with the cohort + source, and if either of those underlying objects change, the subscription will be notified once

I assume such behavior is achievable if we enable `deferUpdates` in Knockout -
https://knockoutjs.com/documentation/deferred-updates.html . But I remember that there were some difficulties with doing this in Atlas: if I'm not mistaken, you've already tried this.

Otherwise:
> Standard – Notifications happen immediately and synchronously. Dependencies are often notified of intermediate values.

chrisknoll: `How would those change at the same moment of time?

If we had full routing support then:
/cohortdefinition/1/inclusionReport/Source1->/cohortdefinition/2/inclusionReport/Source3

Would lead to both a cohort definition ID change and source key change at the same time, and without proper handling, the inclusion report will load for cohort 2, source1 and then cohort 2, source 3.  

>  If they don't, how would you transform two events separated in time 
We can use `deferred` computed observables to minimize the subscription ntoifications until after the current 'request thread' has been handled.

Here's the fiddle that shows the problem: https://jsfiddle.net/cknoll/3ojmde41/1/

And this one uses deferred: https://jsfiddle.net/cknoll/3ojmde41/2/

While there is no concrete case here, i think we should future-proof this, and at the saem time, maybe pick up a technique that we can apply elsewhere in the codebase to prevent multiple round-trips to the server.  My proposed chage is:

1. Add a new computed observable `reportParamas` to the `FeasibiltyReportViewerWithHeader` component:

```
this.reportParams = ko.pureComputed( () => { 
  if (this.cohortId() && this.source()) {
    return { cohortId: this.cohortId(), source: this.source() };
  else return null; 
}).extend({deferred: true});
```

Replace your subscriptions with this:

```
this.subscriptions.push(this.reportParams.subscribe(p => s && this.loadReport()));
``

And in `loadReport` refer to the values in reportParams instead of the individual observables.

The only concern here is to only call `loadReports` when all of the paramater values have 'settled', and I believe the above is the appropriate pattern to apply.

Comment (pavgra) [reply js/pages/cohort-definitions/components/reporting/cohort-reports/feasibility-report-viewer-with-header.js]: @chrisknoll , nice! I've applied this approach

File: js/pages/cohort-definitions/components/reporting/cohort-reports/cohort-reports.js [hunk 3]
```diff
@@ -0,0 +1,46 @@
...
+
+			this.params = params;
+
+			this.tabs = [
+				{
+					title: 'By Person',
+					componentName: 'feasibility-report-viewer-with-header',
+					componentParams: {
+						source: ko.computed(() => params.source()),
+						cohortId: ko.computed(() => params.cohort().id()),
```
chrisknoll: These need disposal or use pureComputed.

pavgra:
@chrisknoll , any other comments?

chrisknoll:
Just let me run it locally and I'll sign off.  Thanks.

pavgra:
@chrisknoll , thank you for thorough review and highlight of issues

Issue #1801, opened 2019-05-30
Incidence Rates: embedded concept set import fails


wivern:
### Expected behavior
Imported concepts shown in the IR embedded conceptset

### Actual behavior
![ir-conceptset](https://user-images.githubusercontent.com/1109753/58628881-8a3dbe00-82e3-11e9-8477-ff2ce210caaf.png)

### Steps to reproduce behavior
![09f92c82-c70c-429b-8177-f11b463e11f9](https://user-images.githubusercontent.com/1109753/58628885-90cc3580-82e3-11e9-9432-ab03e40d5933.png)

chrisknoll:
Should be fixed in #2132

PR #1831, opened 2019-06-04
update input type of concepts import to json

AntonStepanof:
fixes https://github.com/OHDSI/Atlas/issues/1801


js/components/conceptpicker/ConceptPicker.js
```diff
@@ -83,8 +83,8 @@ define(['jquery','knockout', 'text!./ConceptPickerTemplate.html', './InputTypes/
 			self.importValues("");			
 			return; // Nothing to do, import values is blank.
 		}
-		
-		var importConceptIds = this.importValues().split(',').map(function (e) { return +e; });
+		var parsedItems = JSON.parse(this.importValues());
+		var importConceptIds = parsedItems.items.map(item => new ConceptSetItem(item).concept.CONCEPT_ID);
```
pavgra: How this reflects "isExcluded", "includeDescendants" and "includeMapped" from the given JSON?

chrisknoll: The original implementation of this import function was to import by concept_id. you can see that the prior code split a string on `,`, then de-duplicated the list (on old line 89), and it would use the vocabulary provider to resolve the conceptID to  the concept object.

If you want to completely change the import implementation to accept a JSON string of an expression instead of a conceptId, that's fine (tho not backwards compatable), but you will need to look further into the code to recognize that if you are importing a json expression of a concept set, then you're not going to use the vocabulary provider to resolve the concepts (they will already appear in the JSON).

Maybe you want to split this function into 2 modes: import by concept_id and import json expression?

vlbe: @chrisknoll Split import functionality to two modes:
1) Import by Ids
2) Import JSON

chrisknoll:
The import JSON works, but importing by conceptID does not.

Steps to reproduce:
1. Add a new concept set.
2. Click the 'Import' on the bottom
3. Type in 4063431 as the concept Id.  This will resolve to 'Plaque psoriasis' from SNOMED.
4. Observe following exception in console:

image

chrisknoll:
Also, I checked on master, and both functions seem to work except for that (on master) the input fields (for the concept set name) are not focusable (ie I can not click in and edit the name) and I also can not highlight any of the text in the table.  All the 'clickable' elements work (the buttons and the checkboxes) so, there's some very strange behavior here...maybe some sort of CSS selector is making it disabled.

chrisknoll:
Another issue (not sure why I didn't get this error from the beginning) but when I refresh the page with an IR analysis project in the URL, I am getting a 404 on one of the modules:

image


I checked, and the error is correct in that the ConceptSetItem was moved over to 'conceptpicker' (not sure why, a concept picker deals with concepts, a concept set deals with ConceptSetItems, so why was ConceptSetItem moved over to the concept picker?).

I would move that class back to under the conceptsetbuilder module folder.

chrisknoll:
Ok, on the note of the 'read only area' it was a conflict between the 'hasFocus' binding and the deferred update. In a separate PR, we've removed the has focus binding and now we don't see that issue any more.

chrisknoll:
I believe all concept set functionality (including import) has been addressed in PR #2132.  So I will close this one, but please re-open if you think this PR is still necessary (I do not think so since the IR concept set editor has been updated to be the same as all the other UI).

PR #1833, opened 2019-06-04
Cohort definitions were added to export/import(#1821)

### LINKED ISSUE #1821: Import/export of Incidence rates , uses COHORT_ID reference rather than cohort embedding
### Expected behavior

Importing a incidence rate analysis from one environment which was exported from another environment should work by ensuring the cohorts and specifications are properly ported over.

### Actual behavior

Importing an incidence rate analysis has COHORT_ID references, which are not assured to be the same across environments, so analysis is misspecified and new cohorts must be selected. If the new environment doesn't have cohorts specified, then unstable behavior ensues upon generation.

Import an incidence rate analysis from one environment to another.

ssuvorov-fls:
fixes #1821


pavgra:
Rebuilt according to review in https://github.com/OHDSI/Atlas/pull/1842

PR #1843, opened 2019-06-05
Cohort reports via custom components

pavgra:
Tagging @anthonysena , @chrisknoll


chrisknoll:
Hi, @pavgra,
 Just letting you know, i'm still digesting this.  I'm looking at the web components information and getting myself familiar with those concepts and what you've introduced here.  I'm a little confused as to why there's a 'by person' component and a 'by event' component: those are the same components just with different data passed in to render.  So I'm still working that one out as well.

pavgra:
@chrisknoll , thanks for that! Yes, opinions and views from a side would be extremely useful in this case when a totally new mechanism is suggested. Hopefully, during WG call I described clearly enough the flow of my thoughts and decisions related to this development.

> why there's a 'by person' component and a 'by event' component: those are the same components

The reason is:
- The approach introduces dynamic discovery of reports so each report should be presented by a separate component;
- I established a common set of parameters to be passed to any cohort report: cohortId & dataSourceId (`cohort-reports.js` -> `const componentParams =  {...`). To distinguish between 'By person' and 'By event' I would need to pass another flag which doesn't fit well since it would be very specific for these two reports

chrisknoll:
> I established a common set of parameters to be passed to any cohort report: cohortId & dataSourceId (cohort-reports.js -> const componentParams = {...). To distinguish between 'By person' and 'By event' I would need to pass another flag which doesn't fit well since it would be very specific for these two reports

Hi, I should have said: the "by person" and "by event" are 2 'views' for the same report "Inclusion Report".  Back pre 2.6 when we had the 'generate with features', the reports section showed 2 tabs:  Inclusion report and Cohort Features:
CohortReportTabs

I don't know if it was this PR that split that report into 2 separate reports, or if that was part of the base PR, but, those reports should be brought back together, since you should see the Inclusion Report as one report containing a By Person and By Event view.

pavgra:
@chrisknoll , not sure whether I stated this explicitly but neither this nor previous PR (related to tabs refactoring) attempted to change any applied logic. The two distinct tabs have already been presented in 2.7.1. Here is a screenshot from the current http://www.ohdsi.org/web/atlas instance:
image

chrisknoll:
Right and I believe you are considering the two tabs 2 separate reports?

pavgra:
I did consider that, right

chrisknoll:
So what I'm saying is:  the 2 tabs belong to a single report 'Inclusion Report', not 2 separate reprots 'By Person' and 'By Event'.  I'll look at the current codebase and make the necessary modifications so that those are enclosed into a single report 'Inclusion Report'.  I believe some of the complexity in the way this PR is written is to allow the 'By Person' and 'By Event' tabs to appear as different report components (with different names) and this isn't necessary or desired.  If the 'Inclusion Report' is registered as a cohort report then you should just need to load the 'Inclusion Report' component.

I'm still preparing some technical feedback on how things are operating here, but at a high level: I think this PR is modifying the wrong part of atlas to support a dynamic, registry based loading of cohort reports.  The Inclusion report isn't a cohort report (it reports things about people who didn't get into the cohort).  The reports you'd find under characterization or the Heracles Reports is cohort reports, so some of my feeback is goign to be that we don't need the 'inclusion report' to be dynamic in the way you describe.  Your example of a new report is the 'geospatial report', and that isn't something that would be put alongside the Inclusion Report, rather that would be something that would be under a dedicated section for cohort reports (like we have with the Heracles reports, but @pbr6cornell has requested that those reports get moved under Cohort Characterization because those are characterization reeports from the people in the cohort.

pavgra:
> So what I'm saying is: the 2 tabs belong to a single report 'Inclusion Report', not 2 separate reprots 'By Person' and 'By Event'.

Ok, I propose to open another issue and tackle it there since current PR hasn't done any changes related to this topic.

> The Inclusion report isn't a cohort report

I am very open to other proposed names: just tell me how to call that

> Your example of a new report is the 'geospatial report', and that isn't something that would be put alongside the Inclusion Report

I would keep the discussion focused on current PR and not dive into what I'm willing to put into custom reports as long as that is not contributed into Atlas codebase. The PR is providing a plugin mechanism, let's talk that through.

chrisknoll:
> Ok, I propose to open another issue and tackle it there since current PR hasn't done any changes related to this topic.

That's fine, I will work on this in a separate branch, since I think i haven't conveyed the purpose of the Inclusion Report properly, and so it makes it difficult for people without that background to know where those reports fit in to the application.

> I am very open to other proposed names: just tell me how to call that

Just to be clear, it's not about names, it's about function.  Picking a different name and putting those types of reports together would be missing the point.

> I would keep the discussion focused on current PR and not dive into what I'm willing to put into custom reports as long as that is not contributed into Atlas codebase. The PR is providing a plugin mechanism, let's talk that through.

I'll update my draft review comments with annotation about whether it's about the plugin mechanism or how the reports are being grouped together in the component registry.  Both those elements are part of this PR so it's difficult to separate them out.

pavgra:
Seems like I'm missing something here. We have 2 items currently: Inclusion by Events, Inclusion by Person.  You are saying that you are going to organize the items properly in a separate PR since that is the behavior which already existed. What is, then, meant under:

> how the reports are being grouped together in the component registry. Both those elements are part of this PR

?

What `reports are grouped` if there are only 2 of them now and you are going to reorganize them separately?

chrisknoll:
Not to be repetitive, but let me restate what I said earlier:

> Hi, I should have said: the "by person" and "by event" are 2 'views' for the same report "Inclusion Report".  Back pre 2.6 when we had the 'generate with features', the reports section showed 2 tabs:  Inclusion report and Cohort Features:


CohortReportTabs

Please look closely at the above screenshot.  Prior to 2.6, there were 2 reports under the cohort generation reports:  Inclusion Report and Cohort Features.  This was put here because, at that time,  you specify if you want to generate cohort with features or not, so it felt like it made sense to make the report appear under the 'generation' tab.

Since cohort characterization, we removed _one_of the reports (Cohort Features) to the new Cohort Characterization report and left the remaining single report under cohort generation because it was still something that was directly tied to cohort generation.  But, we didn't want to just leave a report view with a single tab (what's the point of a tab with a single tab?) so we removed the 'Inclusion Report' tab and just displayed the Inclusion Report by default without the tab header.  

I'm coming to the conclusion of 2 things here:  1. you should think about how the UX should appear in the case where a single report is configured via plugin:  do you show a tab view with a single tab, or will you just have the entire display be the single report.  And 2: I think applying the plugin pattern to this area of the application just doesn't make sense and is distracting from the focus of evaluating the plugin approach you've invented.  I'd like to focus on that element of the PR, if you would rather do so, or we can continue to debate the report structure that I designed.

pavgra:
All is clear, except for

> I think applying the plugin pattern to this area of the application just doesn't make sense

Why cannot you have different reports depending on a generated cohort? Or how the plugin mechanism applied to the area harms or confuses anyone?

pavgra:
Closed in favor of https://github.com/OHDSI/Atlas/pull/1900

Issue #1840, opened 2019-06-04
Concept set included source codes not updated until save


anthonysena:
### Expected behavior

Navigating between the included concepts and included source codes reflects changes in the concept set expression.

### Actual behavior

When a concept set has unsaved changes, the included source codes does not reflect the current state of the concept set.

### Steps to reproduce behavior

Create a new concept set with concept id == 1503297 (metformin). Navigate to the included concepts and verify 1 concept is shown. Do the same for source codes and verify the # of concepts displayed (for me this is 74). Next, go back to the expression, select descendants. The count on the included concepts will change (for me this is 5930) but navigating to the tab continues to show only 1 concept. The included source codes is the same story - the original list is shown.

anthonysena:
As a note: I've also noticed this same problem for concept sets that are part of a cohort definition in investigating #1850.

PR #1852, opened 2019-06-06
Fixed concept set included source codes not updated until save

### LINKED ISSUE #1839: Concept set export not updating until after saving
### Expected behavior

Concept set export reflects the current: concept set expression, concept identifier list and included concept identifier list.

### Actual behavior

Concept set export reflects the last set of attributes that were saved.

Create a new concept set with concept id == 1503297 (metformin). Select its descendant concepts. From the included concepts, select a 2 concepts to include. Navigate back to the expression, remove the descendant check to have a concept set with 3 concepts. Navigate to Export and you will see only metformin

### Suggested approach

Prevent export until the concept set is saved.

vlbe:
Fixes #1839 #1840

### DIFF
--- a/js/Model.js
+++ b/js/Model.js
@@ -412,6 +412,7 @@ define(
 				this.resolvingConceptSetExpression(true);
 				const resolvingPromise = httpService.doPost(sharedState.vocabularyUrl() + 'resolveConceptSetExpression', expression)
 					.then(callback)
+					.then(this.resolvingConceptSetExpression(false))
 					.catch(() => this.handleVocabularyDataSourceFailure());
 
 				return resolvingPromise;
@@ -662,7 +663,7 @@ define(
 				}
 			}
 
-			async loadRepositoryConceptSet(conceptSetId, viewToShow, mode, forceLoad = false) {
+			async loadRepositoryConceptSet(conceptSetId, viewToShow, mode) {
 				// $('body').removeClass('modal-open');
 				this.componentParams({});
 				if (conceptSetId == 0 && !this.currentConceptSet()) {
@@ -676,7 +677,6 @@ define(
 				if (
 					this.currentConceptSet()
 					&& this.currentConceptSet().id == conceptSetId
-					&& !forceLoad
 				) {
 					this.currentConceptSetMode(mode);
 					this.handleViewChange(viewToShow, { conceptSetId, mode });
--- a/js/Model.js
+++ b/js/Model.js
@@ -662,7 +662,7 @@ define(
 				}
 			}
 
-			async loadRepositoryConceptSet(conceptSetId, viewToShow, mode, forceUpdate = false) {
+			async loadRepositoryConceptSet(conceptSetId, viewToShow, mode, forceLoad = false) {
 				// $('body').removeClass('modal-open');
 				this.componentParams({});
 				if (conceptSetId == 0 && !this.currentConceptSet()) {
@@ -676,7 +676,7 @@ define(
 				if (
 					this.currentConceptSet()
 					&& this.currentConceptSet().id == conceptSetId
-					&& !forceUpdate
+					&& !forceLoad
 				) {
 					this.currentConceptSetMode(mode);
 					this.handleViewChange(viewToShow, { conceptSetId, mode });
--- a/js/Model.js
+++ b/js/Model.js
@@ -378,6 +378,16 @@ define(
 				this.currentView(view);
 			}
 
+			setConceptSetExpressionExportItems() {
+				var highlightedJson = this.syntaxHighlight(sharedState.conceptSetExpression());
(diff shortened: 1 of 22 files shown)


anthonysena:
@vlbe - did you also happen to test these changes on a concept set that is part of a cohort definition? I didn't not that earlier in issue #1840 but it is also present in that area of the application. My hope is the changes in this PR address it but wanted to see if you had confirmed this already.

anthonysena:
@vlbe - I've confirmed that this problem still remains for concept sets in the cohort definition. Sorry I did not mention that originally in the issue. Let me attempt to carry your fix into that area and would ask for a review.

js/Model.js
```diff
@@ -412,6 +412,7 @@ define(
 				this.resolvingConceptSetExpression(true);
 				const resolvingPromise = httpService.doPost(sharedState.vocabularyUrl() + 'resolveConceptSetExpression', expression)
 					.then(callback)
+					.then(this.resolvingConceptSetExpression(false))
```
pavgra: @anthonysena, this is a function call which will be resolved before `httpService.doPost` and `callback` complete

chrisknoll: So, are you saying the proper syntax would be something like:

```
.then(() => this.resolvingConceptSetExpression(false))
```

Or do some sort of `bind` where the function will be invoked with specific parameters?  Is there a standard approach we should take for this sort of coding pattern?

pavgra: Right, either option should work. Honestly, don't have a preference here

anthonysena: Thanks @pavgra and @chrisknoll for the feedback here. My focus initially was simply setting the `resolvingConceptSetExpression` to false at the time of the vocabulary service call. I've revised this to use the suggested approach:

.then(() => this.resolvingConceptSetExpression(false))

And all is well. PR is updated. Thanks!

vlbe:
@anthonysena
There is an issue that JSON for export is not updated when you save Concept Set. There are two ways as i've found  to solve that
1) after saving - .loadRepositoryConceptSet (that's why at original pr i've added "forceLoad" flag, because loading is skipped 
2) after saving call for resolveConceptSetExpression. But it adds ANCESTORS to concepts in JSON.

anthonysena:
Thanks @vlbe for that feedback. I should have noted that I removed the `forcedLoad` flag since I did not see if used anywhere in the code base when invoking the `loadRepositoryConceptSet` function. From what I could see, it was `forcedLoad = false` and inside the function it was used as `!forcedLoad` which seemed to make it redundant since I didn't see any callers using this newly added parameter. Perhaps I missed it?

Let me take a look at the behavior saving and after resolve to see where we might prevent the ANCESTORS from creeping into the JSON.

js/pages/concept-sets/components/tabs/conceptset-expression.js
```diff
@@ -48,14 +46,30 @@ define([
       );
     }
 
+    toggleCheckbox(d, field) {
+      if (this.canEdit) {
+        const concept = this.selectedConcepts().find(i => !!i.concept && !!d.concept && i.concept.CONCEPT_ID === d.concept.CONCEPT_ID);
```
pavgra: Seems like this repeats code in `js/components/conceptset/conceptset-editor.js`: wouldn't it be better to move the code into a shared function?

anthonysena: Moved toggleCheckbox and renderCheckbox to CommonUtils: https://github.com/OHDSI/Atlas/pull/1852/commits/c9cd4b96dbe7da048229f1d782c9efb6c2173e3e

js/utils/CommonUtils.js
```diff
@@ -186,6 +192,16 @@ define([
 	const f = (a, b) => [].concat(...a.map(d => b.map(e => [].concat(d, e))));
 	const cartesian = (a, b, ...c) => (b ? cartesian(f(a, b), ...c) : a);
 
+	const toggleCheckbox = function(hasPermissions, selectedConcepts, d, field, successFunction) {
```
vlbe: Just a question, maybe let's rename this to smth like toggleConceptSetCheckbox? If it's commonUtils so it's supposed that this functions would fit to everything.

anthonysena: Thanks for that feedback @vlbe - I agree that making those functions more descriptive would be best since they reside in a common utility class. I've made that change in https://github.com/OHDSI/Atlas/pull/1852/commits/f0214db8b2e6e9d2d8d97d3b1f71b159afab65cd

anthonysena:
@vlbe @pavgra - please let me know if you have any additional feedback. If not I'll merge this in. Thanks!

vlbe:
@anthonysena Looks good!

PR #1870, opened 2019-06-07
Fixes Estimation & Prediction lists default sorting column

### LINKED ISSUE #1867: Estimation & Prediction lists default sorting does not follow other lists
### Expected behavior
As all other lists are sorted by Modified At field, PLE/PLP lists should have the same sorting

### Actual behavior
Currently PLE/PLP lists are sorted by Created At field

Open Estimation and Prediction lists and see sorted field

wivern:
Fixes #1867


pavgra:
As discussed, let's check all other sections and apply the approach there

anthonysena:
> I believe we need a COALESCE(updated, created) logic. And if I'm not mistaken, something like this was done by @chrisknoll recently in Cohort Definitions

That's right @pavgra - here is a reference to that PR: https://github.com/OHDSI/Atlas/pull/1794/files

chrisknoll:
Hi,
 No, that PR for atlas didn't solve the COALESCE(updated, created), that PR just injects a 'timestamp' value into the list so that sorting can be done by the timestamp instead of a formatted date string.

Solving the COALESCE problem should be handled on the back end directly from the service: return the created date when modified date is not available.

js/components/circe/components/CohortConceptSetBrowser.js
```diff
@@ -81,6 +77,9 @@ define(['knockout', 'text!./CohortConceptSetBrowserTemplate.html', 'services/Voc
 
 			VocabularyProvider.getConceptSetList(url)
 				.done(function (results) {
+					results.forEach(cs => {
```
pavgra: These 3 lines are repeated at least 7 times. Wouldn't it be better to move them into Utils?

PR #1873, opened 2019-06-07
adds confirmation dialog when deleting role

### LINKED ISSUE #1868: Delete role button does not show confirmation dialog
### Expected behavior
When Delete button is clicked for Role, confirmation dialog should be shown

### Actual behavior
No confirmation dialog

1. Create new role and save
2. Click delete button

wivern:
Fixes #1868

### DIFF
--- a/js/pages/configuration/roles/role-details.js
+++ b/js/pages/configuration/roles/role-details.js
@@ -324,16 +324,19 @@ define([
         }
 
         async delete() {
-            if (confirm("Are you sure to delete role?")) {
-							this.loading(true);
-							await roleService.delete(this.roleId());
-							var roles = this.model.roles().filter((role) => {
-								return role.id != this.roleId();
-							});
-							this.model.roles(roles);
-							this.close();
-							this.loading(false);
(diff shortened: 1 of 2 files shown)


js/pages/configuration/roles/role-details.js
```diff
@@ -324,14 +324,16 @@ define([
...
         async delete() {
-            this.loading(true);
-            await roleService.delete(this.roleId());
-            var roles = this.model.roles().filter((role) => {
-                return role.id != this.roleId();
-            });
-            this.model.roles(roles);
-            this.close();
-            this.loading(false);
+            if (confirm("Are you sure to delete role?")) {
```
pavgra: Move to util method, same as here
https://github.com/OHDSI/ArachneUI/blob/d84c956e95a8ca8cbb4cb436b9927aed329299f1/src/modules/Admin/components/AdminList/Table/container.js#L66

pavgra: `Are you sure you want to delete the role?`

PR #1884, opened 2019-06-08
Disabled export for unsaved entities

### LINKED ISSUE #1864: Cohort characterization, IR and pathways export
### Expected behavior

In a characterization, the export JSON reflects the contents of the "design" tab.

### Actual behavior

The export JSON does not reflect the contents of the "design" tab when there are unsaved changes.

Create a new characterization with a single cohort and 1 or more features. Save it. Then add another cohort definition to the design. Navigate to Utilities -> Export and observe that the newly added cohort definition is not part of the definition

### Idea

Prevent export until the design is saved to the server since that mechanism is used to grab the cohort definition expression for export.

vlbe:
Fixes #1864

### DIFF
--- a/js/components/utilities/export.js
+++ b/js/components/utilities/export.js
@@ -6,6 +6,7 @@ define([
     'components/Component',
     'utils/AutoBind',
     'utils/CommonUtils',
+    './const',
     'less!./export.less',
 ], function (
     ko,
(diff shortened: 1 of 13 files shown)


js/components/utilities/export.js
```diff
@@ -25,15 +25,19 @@ define([
             this.entityId = params.entityId;
             this.isPermittedExport = params.isPermittedExport || (() => false);
             this.exportService = params.exportService;
-
+            this.dirtyFlag = params.dirtyFlag;
+            this.errorMessage = params.errorMessage || 'Export is not permitted';
```
anthonysena: I think we'd like to have separate messages in the 2 possible scenarios for this component:

1. You do not have permissions to export 
2. The export is prevented until the calling entity is saved

I'm supportive of passing in an error message as you've done here for item 2 above. In the event that you do not have permissions, the component should explicitly state this.

vlbe: @anthonysena 
So, as far as i understood we can do the following
1) If entity is not saved we can show either errorMessage(if it is passed to params) or default message "The export is prevented until the calling entity is saved - "You do not have permissions to export"
Right?
2) if export is not permitted  -

anthonysena: That's correct @vlbe. You could make both messages parameterized for the component but suspect we'll only override the `errorMessage` explicitly.

PR #1889, opened 2019-06-10
Fix candidate negative control display

### LINKED ISSUE #1888: Negative control link display
### Expected behavior

Links to concepts identified in the negative control process are properly formatted.

### Actual behavior

image

anthonysena:
Updates the negative control table options to use the `render` property where applicable. Updated to centralize formatters to a single function call. Fixes #1888

### DIFF
--- a/js/components/evidence/utils.js
+++ b/js/components/evidence/utils.js
@@ -12,8 +12,18 @@ define(['knockout'], function(ko){
             row.ancestorSplicerCount > 0
         )
     }
+
+    function formatNumberWithCommas(val) {
+      return val.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ",");
+    }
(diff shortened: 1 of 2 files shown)


js/components/evidence/utils.js
```diff
@@ -12,8 +12,18 @@ define(['knockout'], function(ko){
             row.ancestorSplicerCount > 0
         )
     }
+
+    function formatNumberWithCommas(val) {
+      return val.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ",");
```
pavgra: Don’t we have `numeral` library for this?

anthonysena: Nice catch - I forgot about that library. I'll update to use that instead.

PR #1890, opened 2019-06-10
Fix polling delay in IR - fixes #1883

### LINKED ISSUE #1883: IR Display Lag
### Expected behavior

I expect that when the system is checking to see if there are results loaded for an analysis that there is some message displayed.

### Actual behavior

When switching between viewing results (i.e. Pathways analysis) and then to the Incidence Rate results, there is a descent (> 5 sec) lag before results are shown which can be misleading when you know there are results.

Generate a pathway analysis and IR analysis on at least 1 data source. Open each analysis so the left-hand nav shows a green icon (indicating that each analysis is open). Navigate to the pathway analysis and view the results. Next, navigate to the open IR analysis, click the generate tab to view results. You may be faced with a delay before the results of each source are shown.

### To whomever is assigned

Please verify the same behavior - if you are unable to reproduce this, please let me know. It is consistently reproducible in our environment.

anthonysena:
Could use some feedback from @chrisknoll @pavgra on this.

Here is the issue: when the `js/pages/incidence-rates/ir-manager.js` component is loaded, the `init()` is called asynchronously. Before the changes in this PR, this would call the WebAPI's sources endpoint to retrieve the sources and find those that are configured with both a CDM & Results daimon. With the changes I've made in this PR, this information will be retrieved from the `sharedState` thus preventing a round-trip to the server. 

There is still a potential race condition here: in the case of #1883, the IR analysis is already loaded such that when the `onRouterParamsChanged` function is called, it will automatically start polling for updates to the source. Within the `pollForInfo` function, it requires that the sources with a CDM & results daimon exists in order to display the summary of results. So, my fix here is potentially incomplete in the rare case that the source list cannot be parsed before the polling for information starts.

I think the proper approach is to start polling after the sources are initialized and the selected IR analysis is loaded. There could be a computed observable that then kicks off the polling instead of having the upstream callers start it. What do you think?

### DIFF
--- a/js/pages/incidence-rates/ir-manager.js
+++ b/js/pages/incidence-rates/ir-manager.js
@@ -461,7 +461,7 @@ define([
 
 		async init() {
 			this.refreshDefs();
-			const sources = await sourceAPI.getSources();
+			const sources = sharedState.sources();
 			const sourceList = [];
 			sources.forEach(source => {
 				if (source.daimons.filter(function (daimon) {
--- a/js/pages/incidence-rates/ir-manager.js
+++ b/js/pages/incidence-rates/ir-manager.js
@@ -111,6 +111,12 @@ define([
 			this.sources = ko.observableArray();
 			this.stoppingSources = ko.observable({});
 
+			this.canStartPolling = ko.computed(() => {
+				if (this.sources().length > 0 && this.selectedAnalysis() && this.selectedAnalysis().id() > 0) {
+					this.startPolling();
+				} 
+			});
+
 			this.cohortDefs = ko.observableArray();
 			this.analysisCohorts = ko.pureComputed(() => {
 				var analysisCohorts = {targetCohorts: ko.observableArray(), outcomeCohorts: ko.observableArray()};
@@ -277,7 +283,6 @@ define([
 				this.selectedAnalysis(new IRAnalysisDefinition(analysis));
 				this.dirtyFlag(new ohdsiUtil.dirtyFlag(this.selectedAnalysis()));
 				this.loading(false);
-				this.startPolling();
 			});
 		}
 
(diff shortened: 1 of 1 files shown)


pavgra:
> I think the proper approach is to start polling after the sources are initialized and the selected IR analysis is loaded

Sounds reasonable for me

anthonysena:
Thanks @pavgra - I've updated this PR to use a computed: https://github.com/OHDSI/Atlas/pull/1890/commits/aa03d0480c635d2cc3c9dd551f37700e2a2db130. This approach is working well from my tests.

chrisknoll:
I'm concerned that there's a edge-case where the `computed` gets signaled multiple times and will cause multiple polling to occur.

chrisknoll:
@anthonysena , since sources is being put into shared state, and the concern here is that we're taking action before sources have been loaded, maybe we need to register a 'sourcesLoaded' promise in shared state such that once the sources have been loaded in the app, we can resolve the promise.

Then, in your IR `init` method, we `await sharedState.sourcesLoaded` so that we won't proceed until sources are loaded, and then we begin polling after the sources are complete.  Maybe using await isn't right here but somet hing like sharedState.sourcesLoaded.then(() => startPolling).

The idea is that we only wan tto start polling during init() and stop polling during destroy.  The key is that we now need to wait for sharedState's sources to be confirmed to be initialized before we start polling the sources.

If you do think the computed approach is best, just make sure you dispose it in the dispose() method.

anthonysena:
Thanks @chrisknoll for that feedback. I also shared your concern with having things wrapped in a `computed` and the potential that it could get triggered multiple times. 

Here is where I landed: the `sharedState.sources()` collection is initialized at app startup. So, I feel pretty comfortable using that when we are initializing the IR manager since no components are instantiated until the app signals it is ready. I decided to remove `async` from the `init()` function in the ir-manager since I want to make sure the sources are properly parsed before any of the other functions are invoked. 

Let me know if you have any concerns with this approach.

chrisknoll:
I think you are correct that the application bootstrap is designed to halt future component loads until the initial load is complete.  So, loading it there so that is is 'ensured' should do the trick. Let me pull it down, set some breakpoints and make sure that things load in order. I think the main test is navigating directly to an IR project URL.

anthonysena:
Excellent - thanks for the review and testing the changes. I'll merge this in. Thanks!

PR #1891, opened 2019-06-10
Update to use numeral for #1888

anthonysena:
Update to use numeral for number formatting as suggested in this comment: https://github.com/OHDSI/Atlas/pull/1889#discussion_r292158432

### DIFF
--- a/js/pages/cohort-definitions/components/cohortfeatures/tableConfig/FormattingOptions.js
+++ b/js/pages/cohort-definitions/components/cohortfeatures/tableConfig/FormattingOptions.js
@@ -1,4 +1,6 @@
 define(function (require, exports) {
+	
+	var numeral = require('numeral');
 
 	function FormattingOptions() {
 		var self = this;
@@ -8,7 +10,7 @@ define(function (require, exports) {
 		}
 
 		self.numberWithCommas = function (d) {
-			return d.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ",");
+			return numeral(d).format('0,0');
 		}
 	}
 
--- a/js/components/evidence/utils.js
+++ b/js/components/evidence/utils.js
@@ -1,4 +1,4 @@
-define(['knockout'], function(ko){
+define(['knockout', 'numeral'], function(ko, numeral){
 
     function hasEvidence(row){
         return (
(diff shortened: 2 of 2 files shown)


chrisknoll:
Code checks out, but, did we want to identify anywhere else in the codebase that may not be using this standard formatting function? That way this PR can have a broader purpose to use `numeral` for formatting.

Just a suggestion, I don't want to hold anything up.

anthonysena:
> js/pages/data-sources/components/reports/achillesHeel.js line 52.

I had looked at this one earlier. The challenge is that the replacement done is on the heel message which _may_ contain a number that requires formatting. I left this one alone as a result.

> /js/pages/cohort-definitions/cohort-definition-manager.js line 443
> /js/Model.js line 573
> /js/pages/pathways/components/tabs/pathway-results.js line 35

Best I can tell these could be replaced with `numeral` since the d3.format(",") is being passed to another pointer called `commaFormatted` which is then used to format the value. So we could replace commaFormatted as: `const commatFormatted = (val) => numeral(val).format("0,0")`. Or something close to that syntax....

chrisknoll:
> I had looked at this one earlier. The challenge is that the replacement done is on the heel message which may contain a number that requires formatting. I left this one alone as a result.

Gotcha, maybe the regex gurus can look into this and basically change the regex to identify the numbers in the string, and then replace the matched numeric-string with the result of numeral(val).format(). That way the regex is focused on identifying the location of numbers, an the numeral is focused on formatting the value.

js/pages/cohort-definitions/components/cohortfeatures/tableConfig/FormattingOptions.js
```diff
@@ -1,4 +1,6 @@
 define(function (require, exports) {
+	
+	var numeral = require('numeral');
```
pavgra: Do we still need `cohortfeatures` under cohort definitions? Why not to drop that code completely?

chrisknoll: Would be a good thing to do in a separate PR.

pavgra: @anthonysena , I'd propose then to open the appropriate issue and merge the current PR after that

anthonysena: Done in #1898

PR #1900, opened 2019-06-13
Allow to extend Cohort reports & Profile widgets via Atlas plugins

pavgra:


js/pages/profiles/profile-manager.js
```diff
@@ -309,6 +311,8 @@ define([
 				if (this.personId()) {
 					this.loadPerson();
 				}
+
+				this.plugins = pluginRegistry.findByType('atlas-profile-widget');
```
chrisknoll: Should we establish a set of constants for each plugin type we support?  so cohort reports and profiles are the two types I see so far...

js/pages/cohort-definitions/components/reporting/cohort-reports/const.js
```diff
@@ -1,10 +1,13 @@
 define([], () => {
+    const COHORT_REPORT_COMPONENT_TYPE = 'atlas-cohort-report';
```
chrisknoll: related to the other comment about establishing constants: should we group all the plugin type constants together, such that anyone who wants to register their own component for the plugin, would do something like `registry.add(pluginConst.COHORT_REPORT_COMPONENT,{...pluginDef...})`.

My point is I don't think the constant for cohort report or profile widget belongs to the cohort-reports/profile-manager components...those should refer to the same constants that were used to register the plugins, which should belong to a pluginConstants module or something like that.

js/services/PluginRegistry.js
```diff
@@ -0,0 +1,26 @@
...
+		let registeredPluginsOfGivenType = registry.get(type) || [];
+		registeredPluginsOfGivenType.push({ title, priority: priority || 999, html });
+		registry.set(type, registeredPluginsOfGivenType);
+	}
+
+	function findByType(type) {
+		return [ ...registry.get(type) || [] ].sort((a, b) => (a.priority > b.priority) ? 1 : -1);
+	}
+
+	document.addEventListener(REGISTER_PLUGIN_EVENT, e => {
```
chrisknoll: I'm interested in seeing what your plan is for using the event notification to lead to a plugin registration.  Initially, I thought this event was used to notify when a plugin has been added to atlas (so you can handle new plugins at run-time), but now i see it's responding to the event to add a plugin.  This was unexpected to me, and was wondering what the long-term goal of this function is.

pavgra: The goal is to be able to create a plugin which doesn't need Atlas as an explicit dependency (doesn't reference the `PluginRegistry` in `define`; what is more, the `PluginRegistry` could be loaded only via `optional` loader, otherwise bundling will fail due to lack of actual file in plugin's project) and decouple things. E.g. of usage:
```
document.dispatchEvent(new CustomEvent('registerAtlasPlugin', { detail: {
  type: ...,
  plugin: {
    title: ...,
    html: ...
  }
}}));
```

PR #1924, opened 2019-06-19
Fixed escape content in links

### LINKED ISSUE #1922: Atlas shouldn't allow special characters in input fields
Expected behavior: Should run without any failure or delay.

Actual behavior: Application gets hanged or some features don't work as expected

Steps to reproduce behavior:
1. Cohort > New Cohort
2. Enter the HTML text in cohort name
3. Save
4. Come back to cohort list and a button will be there to click.

I am reporting this because I have found this scenario which may cause issues with Atlas workflow and other multiple issues in WebAPI schema, to fix it, one has to delete the entries in DB.
Attached screenshot

-Ambuj
C1

C2

vlbe:
Fixes #1922

### DIFF
--- a/js/utils/CommonUtils.js
+++ b/js/utils/CommonUtils.js
@@ -3,16 +3,12 @@ define([
 		'atlas-state',
 		'pages/Page',
 		'urijs',
-		'xss',
-		'appConfig',
 	],
 	(
 		ko,
 		sharedState,
 		Page,
 		URI,
-		filterXSS,
-		appConfig,
 	) => {
 
 	const build = function (name, viewModelClass, template) {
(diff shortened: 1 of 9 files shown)


js/utils/CommonUtils.js
```diff
@@ -105,7 +107,7 @@ define([
 	function renderLink(s, p, d) {
 		var valid = d.INVALID_REASON_CAPTION == 'Invalid' ? 'invalid' : '';
 		var linkClass = getConceptLinkClass(d);
-		return '<a class="' + valid + ' ' + linkClass + '" href=\"#/concept/' + d.CONCEPT_ID + '\">' + d.CONCEPT_NAME + '</a>';
+		return '<a class="' + valid + ' ' + linkClass + '" href=\"#/concept/' + d.CONCEPT_ID + '\">' + filterXSS(d.CONCEPT_NAME) + '</a>';
```
chrisknoll: I think this should be using the form `if (p === 'display') { return {link text}; } else { return d.CONCEPT_NAME; }`.

This will return the formatted link for display purposes, but other cases it will return just the CONCEPT_NAME value. This will mean the sorting will be performed using just the COCNEPT_NAME, not the link text.

js/utils/CommonUtils.js
```diff
@@ -107,7 +107,10 @@ define([
 	function renderLink(s, p, d) {
 		var valid = d.INVALID_REASON_CAPTION == 'Invalid' ? 'invalid' : '';
 		var linkClass = getConceptLinkClass(d);
-		return '<a class="' + valid + ' ' + linkClass + '" href=\"#/concept/' + d.CONCEPT_ID + '\">' + filterXSS(d.CONCEPT_NAME) + '</a>';
+		const name = filterXSS(d.CONCEPT_NAME);
```
chrisknoll: I'd just apply the xss filter in the case of display because applying the filterXSS is an expensive operation (and will cause the sort mode of render to be very slow).

Second thing:  do we need to XSS concept name?  I'd expect the values in the vocabulary's concept names to be safe from XSS...

vlbe: @chrisknoll I removed filterXSS from this function. I've also added some filtration in linked-cohort-list and added strict xssOptions config to strip all html tags.

vlbe: @chrisknoll I remove filterXSS from this function. I've also added some filtration to linked-cohort-list with some strict xss options

PR #1951, opened 2019-07-05
Fixed issue with table not updated when adding group to role

vlbe:


olga-ganina:
related to issue #1952

Issue #1493, opened 2019-03-12
Tight coupling between search and concept sets


jhardin29:
### Enhancement request
When selecting a user created concept set to include in a characterization in the feature analysis section it would be useful to be able to view the concept set.  Flagging @anthonysena since we IMed about this issue.

AntonStepanof:
@jhardin29 @anthonysena Should it be additional column Concept Set in Feature analyses table? 
I guess, it can be column 'Concept Set' after 'Actions'. If there is a Concept Set for specific analysis, it will be clickable name of Concept Set(with routing to Concept Set view). Is it correct?

chrisknoll:
@jhardin29 : doesn't the criteria editor let you view the selected concept set by using the drop-down 'concept set quickview'?   But this doesn't really let you modify the concept set by adding/removing concepts.  Nor can you see a list of all concept sets in a feature analysis so you can get rid of unused concept sets.

Hi, @AntonStepanof,
  In a given feature analysis 'set' there are multiple concept sets that could be used across the series of criteria features.

I'd suggest we handle this like we do with concept sets attached to cohort definitions, or concept sets attached to incidence rate analysis.  

Cohort definition:
![image](https://user-images.githubusercontent.com/6818777/55561693-9da42280-56c0-11e9-9965-7c9ccf2fc26c.png)

The limitation here is:  adding/removing concept sets is done at a 'global' level (in Model.js).  Wen you pick a concept set to edit, it puts it into a 'global' variable where you go through the vocabulary search UI to add/remove concepts.

Incidence Rates:  where we edit the concept set locally in the incidence rate:

![image](https://user-images.githubusercontent.com/6818777/55562064-3c308380-56c1-11e9-8ba4-aaa79213e80d.png)

Vocabulary searching is using a very simplified ui:
![image](https://user-images.githubusercontent.com/6818777/55562086-481c4580-56c1-11e9-9f40-843d36f47964.png)

What we should have is a completely encapsulated search component that lets you search, navigate to concept hiearchy/ related concepts, and then select conepts to add to the 'target' concept set expression.  Could be simply adding a tab to the concept set editor:

![image](https://user-images.githubusercontent.com/6818777/55562352-bfea7000-56c1-11e9-80c3-f3e8b2637eb0.png)

[...]

PR #1962, opened 2019-07-12
Fixed issue with saving concept set modification in cohort definition

### LINKED ISSUE #1948: Cannot save concept set modification in cohort definition
### Expected behavior

When changing a concept set in a cohort definition, the selection/deselection of concepts will enable the cohort definition's save button allowing for those changes to be saved to the definition.

### Actual behavior

The changes to the concept set do not appear to register with the dirty flag handler on the cohort definition.

Import this concept set: [cd-concept-set-problem.txt](https://github.com/OHDSI/Atlas/files/3334743/cd-concept-set-problem.txt). Navigate to the concept set tab, open the Acute Myocardial Infarction concept set. De-select one of the concepts in the list and notice that the save icon for the cohort definition remains deactivated.

vlbe:
Fixes #1948

### DIFF
--- a/js/Application.js
+++ b/js/Application.js
@@ -174,7 +174,7 @@ define(
 									return i.concept.CONCEPT_ID === concept.CONCEPT_ID;
 								});
 							}
-							setTimeout(() => conceptSet.expression.items.valueHasMutated(), 0);
+							conceptSet.expression.items.valueHasMutated();
 							self.pageModel.resolveConceptSetExpressionSimple(conceptSet.expression)
 						}
 					});
@@ -191,7 +191,7 @@ define(
 						sharedState.selectedConcepts.remove(function (i) {
 							return i.concept.CONCEPT_ID == conceptSetItem.concept.CONCEPT_ID;
 						});
-
+						self.pageModel.currentCohortDefinition().expression().ConceptSets.valueHasMutated();
 						self.pageModel.resolveConceptSetExpression();
 					});
 
--- a/js/Application.js
+++ b/js/Application.js
@@ -175,7 +175,7 @@ define(
 								});
 							}
 							conceptSet.expression.items.valueHasMutated();
-							self.pageModel.resolveConceptSetExpressionSimple(conceptSet.expression)
+							self.pageModel.resolveConceptSetExpressionSimple(conceptSet.expression);
 						}
 					});
 
--- a/js/Application.js
+++ b/js/Application.js
@@ -191,7 +191,7 @@ define(
 						sharedState.selectedConcepts.remove(function (i) {
 							return i.concept.CONCEPT_ID == conceptSetItem.concept.CONCEPT_ID;
 						});
-						self.pageModel.currentCohortDefinition().expression().ConceptSets.valueHasMutated();
+						self.pageModel.currentCohortDefinition() && self.pageModel.currentCohortDefinition().expression().ConceptSets.valueHasMutated();
 						self.pageModel.resolveConceptSetExpression();
 					});
 
--- a/js/Application.js
+++ b/js/Application.js
@@ -35,7 +35,7 @@ define(
 				this.pageModel = model;
 				this.router = router;
 			}
-			
+
 			/**
 			 * Performs initial setup
 			 * @returns Promise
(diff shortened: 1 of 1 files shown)


js/Application.js
```diff
@@ -191,7 +191,7 @@ define(
 						sharedState.selectedConcepts.remove(function (i) {
 							return i.concept.CONCEPT_ID == conceptSetItem.concept.CONCEPT_ID;
 						});
-
+						self.pageModel.currentCohortDefinition().expression().ConceptSets.valueHasMutated();
```
chrisknoll: What about the case where there isn't a cohort definition open, but the shopping carts are being clicked (to modify an open concept set (ie: a stand-alone concept set))?  Wouldn't currentCohortDefinition() in this case be null?

vlbe: Thanks @chrisknoll 
I've added check for currentCohortDefinition. 

As for your question about sharedState.selectedConcepts. As far as i see through the code, sharedState.selectedConcepts depends on currentCohortDefinition().expression().ConceptSets(), so valueHasMutated on sharedState.selectedConcepts won't work.

chrisknoll: Hi, @vlbe, 
  Thanks for checking. I have an additional question about this code:

The 'selectedConceptSet' could refer to an open stand-alone concept set, or a nested concept set within a cohort definition.  I don't want to re-open old wounds, but the mechanism where the same UI handles editing a concept set inside a cohort definition as well as a stand-alone concept set is a bit complicated to manage, and is prone to errors.  In the context of this case:  there's checks in the code to understand if the concept set is within a cohort definition or the concept set being edited is 'stand alone'.  The code you changed is in the high-level application scope where clicking on a DOM element signals an update to the 'selected concept set'.  What that selected concept set actually is could either be cohort-def based or stand-alone based.  So, I don't think you can blindly check for the presence of a cohort definition when a concept set shopping-cart is clicked. I'm tagging in @anthonysena to talk about this specific element.

The bottom line for me on this issue is that altering the selected concept set via shopping cart should signal the dirty-flag that is attached to the concept set expression, which then should signal up to the dirty flag.  However, it might be the case (and I think you're demonstrating this) is that removing an element from the array _outside the interface for modifying an observable array_ requires the use of `valueHasMutated()`.  The strange thing is that the proper behavior should have been that when you modify one of the concept sets inside `expression().ConceptSets()`, the dirty flag should have been signaled. I wonder if that's the core behavior that we need to check here.  Adding `valueHasMutated` may 'make it work' but I feel there may be a more fundamental problem at play here.

Just remember: `expression().ConceptSets` is an array of concept set expresions...but clicking on a shopping cart manimpulates _one_ of the concept set expressions inside `expression().ConceptSets` is changed.  So, a work around could be: when you change a conceptSet inside `expression().ConceptSets`, signal `expression().ConceptSets.valueHasMutated` (which will signal a dirtyFlag check)....but still, the dirty flag should have handled this.

anthonysena: Just to add to this thread since @chrisknoll tagged me - we make the distinction between a 'repository' concept set and a 'cohort' concept set in the code. As an example of this, please take a look at:

https://github.com/OHDSI/Atlas/blob/master/js/Application.js#L165

The `self.pageModel.currentConceptSetSource()` will help guide which one is currently loaded in the global scope.

chrisknoll: Thanks @anthonysena.

@vlbe , i think you want to have somethign similar in the code you added before you call valueHasMutated:
`
if (self.pageModel.currentCohortDefinition() && self.pageModel.currentConceptSetSource() === "cohort")
`

If that's true, then the concept set that was modified was embedded in the selected cohortDefinition's concept sets, and you can call valuehasMutated on the expression().conceptSets.

I'm still trying to decide if this is the actual issue here, but if some event handler is making changes to an object value _outside of the observable interface to the value_ then, yes, calling valueHasMutated on the observable is the way to indicate that _something_ has changed in it.

vlbe: @chrisknoll @anthonysena thanks a lot for explanation. I could be mistaken, but is it possible to have self.pageModel.currentCohortDefinition() != null and self.pageModel.currentConceptSetSource() === "repository" at the same time?

chrisknoll: This is part of the confusion with the implementation:  I'm trying to remember if opening up a concept set will close the current cohort definition (and conversely opening a cohort definition closes the current concept set)...or is it the case that opening up a cohort definition's concept set is what will close the 'repository' concept set....but I'm pretty sure that once you open up a concept set, it will try to close a cohort definition (which is why you get a 'you have unsaved changes') message when you have an open + dirty cohort definition and you want to create a new concept set.  So, to answer your question: No, I do not think it's possible to have currentCohortDefinition() != null and currentConceptSet() === "repository" at the same time.  However, I do think that it could be possible that it changes in the future, and the 'correct' state to call `valueHasMutated()` is when the `currentConceptSetSource() === "cohort"`.  

Referencing #1493 since it relates to the updated design for editing concept sets. Once a general 'concept set/vocabulary search redesign' is created, we'll have to remember this block of code and remove it.

Issue #1320, opened 2019-02-08
Characterization - Filtered results


anthonysena:
The characterization results page takes a long time to load for characterizations with large result sets. Furthermore, the amount of information in the display is not filtered which can make the page response poor to user actions (i.e. scrolling, clicking, etc).

In the previous incarnation of characterization, the "generate with features" approach utilized a filter on the WebAPI call to pull back those covariates with an average_value > .005. This helped with the load time and responsiveness of the tables displayed. Furthermore, the tables were partitioned based on the domain again to help create smaller bundles of covariates to view together.

It would be nice if:

- The characterization screen utilized a filter, surfaced that to the user and provided results that indicate the filter. So if your viewing (i.e. "Viewing First 1,000 of 100,000 records) you know that you may be missing something. 
- The characterization page allowed for changing the filter (or perhaps it just allowed the full download of all results upon request) then it would help if people want to look at items that are less prevalent.

PR #1982, opened 2019-07-25
Adds ability to configure prevalence threshold for CC results

pavgra:
fixes https://github.com/OHDSI/Atlas/issues/1320


anthonysena:
@pavgra - functionally this looks to work well. A few suggestions noted below:

image

The label "Threshold level" should probably read "Filter results to covariates prevalence greater than ____%". Additionally, since we display this value as a % in the characterization results, I'd suggest allowing for users to enter values between 0.0-100.0% only since these are the only valid values and on the back-end just divide by 100 to pass the threshold to WebAPI. Also, just wondering if this area should be styled to make it clear to the user? My eyes missed it at first...

Also thinking it would be worthwhile to include this information in each of the relevant section headings in the characterization report where applicable. For example: 

### DRUG / Drug Group Era Long Term (prevalence > x%)

It is currently not obvious how these results are filtered and adding this description would be helpful in my opinion.

anthonysena:
@pavgra thanks for making these updates! Looking better from my test with a few more items:

- The prevalence box allows me to enter a value > 100% - I was able to pass 150% which we should prevent
- We should only display the `prevalence > x%` the analyses that have an analysis type == prevalence. At the moment, reports such as the Demographics Age also show this caption which does not apply.

anthonysena:
@pavgra - the filter text box should allow for decimal entry so I could put in something like 0.5. At the moment, the RegEx filter prevents that from happening.

Issue #1820, opened 2019-05-31
Notifications are not updating when tasks are completed


pbr6cornell:
### Expected behavior

When jobs are completed, then (at some reasonable time thereafter), Notifications is updated to say 'Completed'

### Actual behavior

Jobs are completed, but after many minutes, Notifications do not show that update.  Refresh was required to get proper status.
![image](https://user-images.githubusercontent.com/731930/58713857-7fd9fd80-8391-11e9-9f9f-47dee0d5d329.png)

### Steps to reproduce behavior

Generate across multiple functions, Check notification, then go into job and see its done even though generation says its 'started'

pbr6cornell:
If this is related to the discussion about polling vs. websockets, it may be connected to #1877, and can be wait to be resolved in ATLAS 3.0 as long as it doesnt result in performance or bug currently.

PR #1989, opened 2019-07-30
Made job parameters observable to trace changes

aklochkova:
Fixes https://github.com/OHDSI/Atlas/issues/1820


js/components/user-bar.js
```diff
@@ -40,6 +40,12 @@ define([
 			this.sortedJobListing = ko.computed(() => lodash.sortBy(this.jobListing(), el => -1 * el.executionId));
 			this.lastViewedTime=null;
 			this.permissionCheckWarningShown = false;
+			this.updateJobs = ko.computed(() => {
```
wivern: `updateJobs` is never used. Probably you want to use `state.jobListing().subscribe`, don't you?
And while having `jobListing` the `newJobState` is not necessary (Furthermore newJobState works only for cohort definitions, but other types of analyses remains issue).

js/components/user-bar.js
```diff
@@ -141,12 +142,11 @@ define([
 								executionUniqueId: ko.pureComputed(function () {
 									return job.type + "-" + job.executionId;
 								}),
-								duration,
+								duration: duration,
```
wivern: Why we need this change?

js/pages/cohort-definitions/cohort-definition-manager.js
```diff
@@ -530,7 +533,11 @@ define(['jquery', 'knockout', 'text!./cohort-definition-manager.html',
 			this.reportingAvailableReports = ko.observableArray();
 
 			this.pollId = null;
-
+			this.shouldUpdateJobs = ko.computed(() => {
+				if (this.generateReportsEnabled() == true) {
```
wivern: the `== true` is odd

js/services/JobDetailsService.js
```diff
@@ -27,6 +27,24 @@ define(['appConfig', 'services/job/jobDetail', 'atlas-state', 'services/http'],
 			}
 		}
 	}
+	
+	function setJobListMutated() {
+		sharedState.jobListing.valueHasMutated();
+	}
+
+	function addPoll(opts) {
```
pavgra: I still see direct usages of the `PollService.add`:
image

However, the general issue that concerns me is the following: it becomes unclear which service to use. Either the `PollService.add` should become "private" so that developers do not attempt to use it directly, OR, in my opinion, the preferrable solution is to throw event from the `PollService` and process the event in the JobDetails. So that we have loose coupling, avoid ambigioty in services and keep services atomic

js/pages/cohort-definitions/cohort-definition-manager.js
```diff
@@ -465,6 +465,9 @@ define(['jquery', 'knockout', 'text!./cohort-definition-manager.html',
 							if (source) {
 								// only bother updating those sources that we know are running
 									if (this.isSourceRunning(source)) {
+										if(source.status() !== info.status) {
+											jobDetailsService.setJobListMutated();
```
pavgra: Cannot we avoid imperative dirty declaration in favor of using pure observables?

js/components/user-bar.js
```diff
@@ -125,10 +126,10 @@ define([
 							if (job.status() !== n.status) {
 								job.status(n.status);
 								job.viewed(false);
-								job.duration = duration;
-								job.endDate = displayedEndDate;
+								job.duration(duration);
+								job.endDate(displayedEndDate);
 								job.url = jobDetailsService.getJobURL(n);
```
pavgra: Why is this variable remains not observable when all others were migrated to the observables?

js/services/Poll.js
```diff
@@ -6,14 +6,18 @@ define(['knockout', 'visibilityjs'], (ko, Visibility) => {
...
   });
 
   class PollService {
+    constructor(){
+      this.isJobListMutated = ko.observable();
+      this.isJobListMutated.extend({ notify: 'always' });
+    }
     add(opts = {}, ...args) {
-      const { callback = () => {}, interval = 1000, isSilentAfterFirstCall = false } = opts;
+      const { callback = () => {}, interval = 1000, isSilentAfterFirstCall = false, isJobListMutated = false } = opts;
```
pavgra: I would name the `isJobListMutated` in a different way for 2 reasons:
1. It sounds like a statement regarding what happened / what is present, whereas in fact it says whether something should happen in future, i.e. whether the callback should trigger jobs list rerender
2. A bit difficult to distinguish it from the `this.isJobListMutated`

wivern:
@aklochkova Could you resolve conflicts?
And also could you review this, since it seems like originated issue is more or lesser covered by this #2105
And I'd like to note when I tried to check this there were a lot of calls to notification service

js/services/Poll.js
YaroslavTir: I think this comment also applies to the `isJobListMutated` function in the `Poll.js`

js/services/JobDetailsService.js
YaroslavTir: The following points bother me 
1) PollService now has a new responsibility `isJobListMutated`. This contradicts his original idea, just executes callback every N seconds. These processes must be independent 
2) user-bar.js also uses a pollService which for some reason does not work. I would like to understand what is wrong with it or:
3) As an alternative for isJobListMutated it is possible to use events, as I know Vlad has added eventemitter3 framework( I guess we are not merge it yet), Events make the code clearer.

js/services/Poll.js
```diff
@@ -55,5 +61,5 @@ define(['knockout', 'visibilityjs'], (ko, Visibility) => {
     }
   }
 
-  return new PollService();
+  return PollService;
```
wivern, line 64: This change is required since we want to extend base class, am I right?
But should this be done without breaking current functionality?

Issue #1659, opened 2019-04-08
Estimation form enhancements


anthonysena:
PS matching: maximum number of persons dropdown should include large number (e.g. 100)

Change order of study population options to:
- Study start date
- Study end date
- Restrict the analysis to the period when both exposures are observed? 
- Should only the first exposure per subject be included?
- The minimum required continuous observation time prior to index date for a person to be included in the cohort.
- Remove subjects that have the outcome prior to the risk window start?
- Remove subjects that are in both the target and comparator cohort?
- If a subject is in multiple cohorts, should time-at-risk be censored when the new time-at-risk start to prevent overlap?
- If either the target or the comparator cohort is larger than this number it will be sampled to this size. (0 for this value indicates no maximum size)

anthonysena:
Additional feedback on the form changes per @schuemie: [EstimationUiRedesignV02.docx](https://github.com/OHDSI/Atlas/files/3386293/EstimationUiRedesignV02.docx)

anthonysena:
Revisions which supersede the original doc: [EstimationUiRedesignV02_Sena_Martijn.docx](https://github.com/OHDSI/Atlas/files/3503186/EstimationUiRedesignV02_Sena_Martijn.docx)

PR #2015, opened 2019-08-14
Estimation form enhancements

anthonysena:
Re-organized the estimation form per #1659 and feedback from @schuemie

### DIFF
--- a/js/pages/estimation/components/editors/cohort-method-analysis-editor.js
+++ b/js/pages/estimation/components/editors/cohort-method-analysis-editor.js
@@ -37,7 +37,6 @@ define([
 			this.showControlDisplay = ko.observable(false);
 			this.showPriorDisplay = ko.observable(false);
 			this.showConceptSetSelector = ko.observable(false);
-			this.showPsCovariateDisplay = ko.observable(false);
 			this.currentConceptSet = ko.observable(null);
 			this.trimSelection = ko.observable();
 			this.matchStratifySelection = ko.observable();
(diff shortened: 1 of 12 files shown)


js/pages/estimation/components/editors/cohort-method-analysis-editor.js
```diff
@@ -48,6 +47,11 @@ define([
 			this.bounds = ko.observable(this.analysis.trimByPsToEquipoiseArgs.bounds() && this.analysis.trimByPsToEquipoiseArgs.bounds().length > 0 ? dataTypeConverterUtils.percentArrayToCommaDelimitedList(this.analysis.trimByPsToEquipoiseArgs.bounds()) : '');
 			this.studyStartDate = ko.observable(this.analysis.getDbCohortMethodDataArgs.studyStartDate() !== null ? dataTypeConverterUtils.convertFromRDateToDate(this.analysis.getDbCohortMethodDataArgs.studyStartDate()) : null);
 			this.studyEndDate = ko.observable(this.analysis.getDbCohortMethodDataArgs.studyEndDate() !== null ? dataTypeConverterUtils.convertFromRDateToDate(this.analysis.getDbCohortMethodDataArgs.studyEndDate()) : null);
+			this.useRegularization = ko.observable(estimationConstants.isUsingRegularization(this.analysis.createPsArgs.prior) ? true : false);
+
+			console.log("DEBUG: this.analysis.createPsArgs.prior.priorType() == " + this.analysis.createPsArgs.prior.priorType())
```
pavgra: I would suggest using [console.debug](https://developer.mozilla.org/en-US/docs/Web/API/Console/debug)

anthonysena: Good catch @pavgra - I removed those console.log statements. I had thought I took them out.

anthonysena:
@chrisknoll - thanks for the review and your comment is a good catch. I apologize but Martijn and I had some offline conversation on these changes and I failed to update #1659 until just now: https://github.com/OHDSI/Atlas/issues/1659#issuecomment-521401990. I can confirm that the package is working properly in that the form is now properly setting the Control and Prior object as required. The other item you'll notice in these changes is the change to initialize the Control & Prior with a seed = 1 which is consistent with the changes made in CohortMethod: https://github.com/OHDSI/CohortMethod/commit/a2ed2047202af2bcf64d63f9780791e13660918d

PR #2018, opened 2019-08-26
Added type of source 'hive'

aklochkova:
Fixes https://github.com/OHDSI/WebAPI/issues/1168


aklochkova:
was fixed in https://github.com/OHDSI/Atlas/pull/2096

PR #2026, opened 2019-09-15
Added multi-select for cohort definitions

### LINKED ISSUE #1186: Multi-Select Lists
Looking at some of the new development work happening, I am really happy (and excited) to see a multi-select box for selecting features when designing a cohort characterization: 

image

For reference, the component is here: [js/pages/characterizations/components/feature-analyses/feature-analyses-browser.js](https://github.com/OHDSI/Atlas/blob/master/js/pages/characterizations/components/feature-analyses/feature-analyses-browser.js
)

I'd like to extend the [js/components/cohort/linked-cohort-list.js](https://github.com/OHDSI/Atlas/blob/master/js/components/cohort/linked-cohort-list.js) component to have the same behavior which is useful when defining prediction & estimation studies as it gets annoying to have to select a single cohort at a time. My thinking is that I can change the behavior of the `linked-cohort-list` component such that:

- It supports single selection (default) or multi-selection
- It supports the ability to detect items already selected in the list (it looks like the feature selector lack this at the moment unless I missed it)

In thinking about this, I felt that these enhancements might also be useful for the feature browser (potentially) so I wanted to ask: should we abstract out [js/pages/characterizations/components/feature-analyses/feature-analyses-browser.js](https://github.com/OHDSI/Atlas/blob/master/js/pages/characterizations/components/feature-analyses/feature-analyses-browser.js
) so that it might be shared? Or do we think that this would be better encapsulated in each component on its own?

Tagging @pavgra for input since I know your team wrote this component.

vlbe:
Fixes #1186


js/components/cohort-definition-browser.js
```diff
@@ -114,6 +106,20 @@ define([
 				callback();
 			}
 		}
+
+		async loadData() {
+			try {
+				this.isLoading(true);
+				const { data } = await httpService.doGet(`${config.api.url}cohortdefinition`);
```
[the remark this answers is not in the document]
vlbe: Fixed

js/components/entity-browser.js
```diff
@@ -0,0 +1,73 @@
...
+    class EntityBrowser extends Component {
+        constructor(params) {
+            super(params);
+            this.onSelect = params.onSelect;
+            this.multiChoice = params.multiChoice || false;
+            this.showModal = params.showModal;
+            this.scrollY = params.scrollY || '50vh';
+            this.scrollCollapse = params.scrollCollapse || false;
+            this.selectedDataIds =
+                this.multiChoice && params.selectedData
```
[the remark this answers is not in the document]
vlbe: Why should it be observable? We passed `selectedData`(items that were selected before and already presented in cohort-linked-list) to track which items have been already selected.  We load cohorts list and map through response data `this.data(data.map(item => ({ selected: ko.observable(this.selectedDataIds.includes(item.id)), ...item })))`.

js/components/cohort-definition-browser.js
```diff
@@ -94,14 +79,21 @@ define([
...
+				},
 			];
 
-			this.rowClick = this.rowClick.bind(this);
-		}
-
-		rowClick(data) {
-			this.action(() => this.selected(data));
+			if (!!this.multiChoice) {
+				this.columns = [
```
pavgra: Why wouldn't you move this into the EntityBrowser since it is a common wrapper which takes `multiChoice` parameter? This way the code block would appear just once and wouldn't be repeated across "client" components

vlbe: fixed

js/components/cohort-definition-browser.less
```diff
@@ -0,0 +1,11 @@
+.cohort-definition-browser {
+    &__button-panel {
+        display: flex;
+        flex-direction: row-reverse;
+        padding: 1rem 2rem;
+
+        .btn {
```
pavgra: Cannot we apply BEM for this?

js/pages/characterizations/components/feature-analyses/feature-analyses-browser.less
```diff
@@ -8,6 +8,15 @@
   .fe-btn {
     display: flex;
   }
+  &__button-panel {
+    display: flex;
+    flex-direction: row-reverse;
+    padding: 1rem 2rem;
+
+    .btn {
```
pavgra: Same question as above. Plus, it looks like code duplication (`button-panel`)

vlbe: fixed

js/components/cohort-definition-browser.less
vlbe: fixed

js/components/entity-browser.js
```diff
@@ -0,0 +1,85 @@
...
+            const selectedItems = this.data().filter(i => i.selected());
+            this.action(() => this.onSelect(selectedItems));
+        }
+
+        cancel() {
+            this.showModal(false);
+        }
+
+        toggleSelected(selected) {
+            commonUtils.selectAllFilteredItems(this.data, this.filteredData, selected);
```
anthonysena: I'm still facing an issue with the following scenario that appears to be related to the fact that `filteredData` is not properly capturing the filter applied to the data table. Here are the steps to reproduce:

- Create a new cohort characterization
- Import cohort definitions to reveal the modal dialog 
- Use the search filter to filter down the list to a subset of cohort definitions.
- Click the "Select All" button to select the items that are part of the filtered set
- Click the "Import" button to import the selected cohorts into the design

When I put a breakpoint into this line, `filteredData` contains the full set of table entries and not the ones that are filtered based on my search criteria.

vlbe: Fixed

anthonysena:
@vlbe - hope you don't mind but I merged in the latest code from master into this branch and also made a small update in https://github.com/OHDSI/Atlas/pull/2026/commits/c52b487b4be32625ae4933fe3bc5c4119adf1ae8. Overall this is ready - I had one (hopefully) small request. In my review, I noticed that the display of the cohort definition browser changed - a solid black line now appears at the bottom. I found that this due to the addition of the `scrollY` parameter in `js/components/entity-browser.js` (see below). I see that the use of the `scrollY` parameter was used to provide visibility for the Import button when using the modal cohort browser which makes sense. 

It would be preferable though if we could dynamically set the `scrollY` parameter - if it is always set, as it is now in this PR, then the DataTable component will always assume that we would like to have that type of fixed vertical scrolling which isn't always desirable. Perhaps we can re-use what was done here: https://github.com/OHDSI/Atlas/blob/master/js/components/faceted-datatable.js#L183? 

image

PR #2036, opened 2019-09-19
Fixed issue with blinking horizontal scroll during page navigation

### LINKED ISSUE #2035: Blinking horizontal scroll during page navigation
atlas-blinking-hor-scroll

vlbe:
Fixes #2035

### DIFF
--- a/js/styles/atlas.css
+++ b/js/styles/atlas.css
@@ -299,6 +299,7 @@ pre span.key {
   border-bottom: solid 1px #ccc;
   white-space: nowrap;
   text-align: right;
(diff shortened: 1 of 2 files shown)


js/styles/atlas.css
```diff
@@ -299,6 +299,7 @@ pre span.key {
   border-bottom: solid 1px #ccc;
   white-space: nowrap;
   text-align: right;
+  overflow: hidden;
```
pavgra: I would work on the cause of the issue instead of fixing the consequences. What do you think?

PR #2068, opened 2019-11-14
fixes updating of users after import 

ssuvorov-fls:
fixes https://github.com/OHDSI/WebAPI/issues/1155

### DIFF
--- a/js/pages/configuration/users-import/users-import.js
+++ b/js/pages/configuration/users-import/users-import.js
@@ -64,6 +64,7 @@ define(['knockout',
 				this.wizardStep = ko.observable();
 				this.hasPrevious = ko.computed(() => !!this.getStep('prev'));
 				this.hasNext = ko.computed(() => !!this.getStep());
+				this.stepMessage = ko.observable('User import from directory has started...');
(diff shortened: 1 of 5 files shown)


wivern:
@ssuvorov-fls 
1. Could you please make a default sort by the "last executed" field in descending order.
2. And there is some looking weird, why all last executed in all the jobs has the same value:
a766cde4-c1a4-4a7c-ace8-6f4ea933dcad

js/pages/configuration/users-import/users-import.js
```diff
@@ -154,6 +158,29 @@ define(['knockout',
...
+				if (this.pollId != null) {
+					PollService.stop(this.pollId);
+				}
+			};
+
+			async updateJobStatus(jobId) {
+				const data = await jobService.getJob(jobId);
+				if (data.closed) {
+					this.loading(false);
+                    userService.getUsers().then(data => this.model.users(data));
```
pavgra: Please fix indentation

ssuvorov-fls: done

PR #2079, opened 2019-12-09
fixed second login attempt in case of openid auth

### LINKED ISSUE #1984: OpenID login requires 2x login attempts
### Expected behavior

With security enabled using OpenID, I would expect to log in a single time and have ATLAS recognize my user account.

### Actual behavior

I have to make 2 login attempts before Atlas recognized my account.

Configure an OpenID provider with WebAPI & Atlas. In my case, I configured Elixir AAI in support of the EHDEN project (https://elixir-europe.org/services/compute/aai). I can share these details privately to reproduce the issue. Once configured, attempt to log in and you should experience:

- Login from Atlas using Open ID directs you to the Elixir AAI portal for authentication
- Upon successful authentication using Elixir AAI, you will be directed back to ATLAS
- The login area of ATLAS will still show the "Sign In" area on the top.
- Sign in through ATLAS again as mentioned above
- Once directed back to ATLAS, the user name will show up properly.

ssuvorov-fls:
fixes #1984

### DIFF
--- a/js/pages/home/routes.js
+++ b/js/pages/home/routes.js
@@ -16,8 +16,8 @@ define(
         '/welcome/:token': new Route((token) => {
           require(['welcome'], function () {
             authApi.token(token);
-            // application must be reloaded to call bootstrap function again
-            document.location.replace("/");
+            authApi.loadUserInfo();
+            document.location = "#/welcome";
           });
         }),
       };
--- a/js/pages/home/routes.js
+++ b/js/pages/home/routes.js
@@ -16,7 +16,8 @@ define(
         '/welcome/:token': new Route((token) => {
           require(['welcome'], function () {
             authApi.token(token);
-            document.location = "#/welcome";
+            // application must be reloaded to call bootstrap function again
+            document.location.replace("/");
           });
         }),
       };


js/pages/home/routes.js
```diff
@@ -16,7 +16,8 @@ define(
         '/welcome/:token': new Route((token) => {
           require(['welcome'], function () {
             authApi.token(token);
-            document.location = "#/welcome";
+            // application must be reloaded to call bootstrap function again
```
wivern: It's not clear to me why the bootstrap call is required during OpenID authentication. Could you explain with some more details?

js/pages/home/routes.js
```diff
@@ -16,7 +16,8 @@ define(
...
+            document.location.replace("/");
```
wivern: Anyway, this changes affects other OAuth providers such as GitHub or Facebook.

js/pages/home/routes.js
ssuvorov-fls: the problem is that bootstrap and user/me service is called before route and token is processed so user/me return 401 http code

js/pages/home/routes.js
ssuvorov-fls: in fact other oauth methods also need second attempt or reload to work correctly

PR #2084, opened 2019-12-18
Implements better handling for large volumes of concepts

wivern:
Resolves OHDSI/WebAPI#58

Things that are done by this pull request:
1. Reduced the redundant calls to the back-end
1. Removed draw callback of included concepts table  that looks unnecessary, remain one in the codebase since not completely sure there is no case when this callback is valuable
1. Turned on the knockout deferred updates and got up to ten times faster tables redraw, knowing this can be the reason for the existing application compatibility issues there were double-checked whole app features


chrisknoll:
I'll start to take a look at this.

js/components/conceptsetmodal/conceptSetSaveModal.html
```diff
@@ -1,19 +1,24 @@
...
+        showModal: $component.show,
+        iconClass: 'fa fa-save',
+        title: 'Save New Concept Set',
+        template: 'conceptSetSaveTmpl',
+        fade: $component.fade,
+        data: $component
+    }">
+</atlas-modal>
+
+<script type="text/html" id="conceptSetSaveTmpl">
```
chrisknoll: I believe we implemeted the modal type to allow you to embed the template between the `<atlas-modal>` tags. Can you try that out?  having a separate script tag that passes in the template vs just in-line the template in my opinion looks better.  Just remove the `template` param from the component and put the HTML directly inside.

chrisknoll:
Changes look ok, but two things:

1:  There's a conflict. Please check your conflicts: note the recent change in multi-select that allows changing the options before signaling an observable of the change (change is signaled in the 'drop-up' event).  Do not loose this change.

2:  I was testing some other parts of the app, and I went to check the CC results and I recieved a 'javax.rs.NotAllowedException'. I just am confirming that this change didn't lead to some sort of HTTP request for the wrong content type.

When I've checked on number 2, I'll let you know. Until then, please check on #1.

chrisknoll:
Ok, I have checked #2, and on master branch of atlas, I am able to open CC results without error.  However, when I switch to this branch, I recieve an HTTP 500 response, with this in the console:

image

Please test this on your end with CC results generation on WebAPI based on the master branch.

wivern:
@chrisknoll I've re-checked the issue with CC results. It seems the problem has gone after merging the master branch.

PR #2100, opened 2020-01-20
Added check for permissions in data sources tab

### LINKED ISSUE #2004: Sources list in Data sources tab does not respect user's permissions
### Expected behavior
Only sources with granted access are listed in the dropdown at Data Sources page.

### Actual behavior
A user observes a list of all sources in the system and can click any. Only then he/she will see a message telling about lack of access.

### Todo

Add permissions check similarly to how it is done in Profiles

vlbe:
Closes #2004

### DIFF
--- a/js/pages/data-sources/classes/Treemap.js
+++ b/js/pages/data-sources/classes/Treemap.js
@@ -4,6 +4,7 @@ define([
 	'utils/ChartUtils',
 	'const',
 	'services/http',
+	'services/AuthAPI',
 	'./Report',
 	'text!components/charts/datatableTemplate.html',
 	'faceted-datatable'
@@ -13,6 +14,7 @@ define([
 	ChartUtils,
 	constants,
 	httpService,
+	authApi,
 	Report,
 	datatableTemplate
 ) {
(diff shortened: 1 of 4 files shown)


js/pages/data-sources/classes/Treemap.js
```diff
@@ -34,12 +36,18 @@ define([
 			this.byValueAsConcept = false;
 			this.byOperator = false;
 			this.byQualifier = false;
-
+			this.handleConceptClick = node => {
+				if (authApi.isPermittedViewDataSourceReportDetails(this.context.currentSource().sourceKey)) {
```
wivern: This will block all report details when authentication is turned off

js/pages/data-sources/data-sources.js
```diff
@@ -104,8 +104,8 @@ define([
 			];
 
 			this.sources = ko.computed(() => sharedState.sources().filter(function (s) {
-                return s.hasResults && s.hasVocabulary;
-            }));
+				return s.hasResults && s.hasVocabulary && authApi.isPermittedViewDataSourceReport(s.sourceKey);
```
wivern: No sources would be available when authentication is turned off

js/pages/data-sources/classes/Treemap.js
vlbe: No need to check if auth turned off cause this check internally used in isPermittedViewDataSourceReportDetails function

js/pages/data-sources/data-sources.js
vlbe: See above

Issue #760, opened 2018-06-29
when cohort, concept set is opened and then closed, ATLAS resets the last page that was active


gklebanov:
When paging through concepts, cohorts or other design artifacts and concept set or cohort is open, the page that was current is lost and ATLAS goes back to page 1. Instead, the current page should be kept

pbr6cornell:
I agree this is only an 'enhancement', not 'bug', but I think it definitely falls squarely in the 'performance' territory of ATLAS 2.8, because it costs a lot of time to page through a list repeatedly to get back to where you were, and not having to do that would save a good bit of user time.

anthonysena:
I'm moving this item to 3.0 - the changes required for the underlying data table component are too complex and introduce other problems that we can better sort out with a new table component.

PR #2112, opened 2020-01-23
Adds persistence and reloading of datatable pagination params

pavgra:
fixes https://github.com/OHDSI/Atlas/issues/760

### DIFF
--- a/js/extensions/bindings/datatableBinding.js
+++ b/js/extensions/bindings/datatableBinding.js
@@ -6,8 +6,7 @@ define([
 	'xss',
 	'moment',
 	'services/MomentAPI',
-	'utils/CommonUtils',
-	'urijs',
+	'utils/DatatablePaginationUtils',
 	'datatables.net-buttons',
 	'colvis',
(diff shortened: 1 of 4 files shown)


pavgra:
@anthonysena , I do not think that this case can be implemented in a straightforward and clean way:
- after you navigated to a specific cohort definition, you may go over tabs back and forward, go into vocabulary search to attach concepts, etc - so, it is impossible to assign a browser's back button behavior to the close button
- therefore, we would need to store pagination parameters of the previous page:
  - either we can attach them to cohort definition URL - that would be `/cohortdefinition/111?searchParamOfTableFromPrevPage=...` - but that would make links really ugly and the logic very non-trivial
  - or we could put the params into sessionStorage; however, that would lead to the different behavior of the same page for different users, i.e. you wouldn't be able to share a link and expect the app to behave in the same way (since it would depend on your local data)

anthonysena:
I'd like to confirm with @gklebanov around the functional requirements for #760 before we merge this into master.

PR #2118, opened 2020-01-30
Fixed generate button on pathway analysis execution page

### LINKED ISSUE #2069: Generate button on pathway analysis execution page doesn't show generating
### Expected behavior
Clicking Generate in a pathway analysis should show a spinner immediately
### Actual behavior
If there is a job backlog, the genrate button shows a 'not-running' state even tho there is a 'starting' job in the list of executions:

image

1) Have a full backlog of spring batch jobs in the thread pool
2) Click generate on a pathway analysis so that the status goes into 'starting' (but not started) state
3) Observe, the gernate button is still a generate button, no indication of exeuction (can not cancel) and if you click it again, it will register another task to the backlog.

vlbe:
Fixes #2069


vlbe:
@anthonysena I agree. I'll prepare new commit.

anthonysena:
Excellent - thanks for that. I suspect that Estimation/Prediction also have this generation experience.

vlbe:
@anthonysena Hello! I've extracted execution tab content to separate component which will be shared between Cohort Pathway/Characterization/PLE/PLP and removed bunch of files that are not needed anymore. I've tested this on my local environment but i'll be very appreciated if you test it too.

There are some differences between Cohort Pathway/Characterization and PLE/PLP:
- Cohort Pathway/Characterization  execution tab - has an ability to cancel generation + has a  Design column + has link to view reports in Results column
- PLE/PLP execution tab - doesn't have an ability to cancel generation but you can start another one in parallel + doesn't have a Design column + has link to download results in zip instead of view reports in Results column.
- Some differences between conditions to execute, so some `extraExecutionPermissions` can be passed to the componentParams

And i have a question how will be better to implement error handing when loading of design failed. For now, i added alert to notify users that loading of design failed ( https://github.com/OHDSI/Atlas/pull/2118/files#diff-77e919ff16fbf32d1d863df621f0eb70R178 ). Maybe it will better to implement one of this:
1)  Open modal window with loading status(as now) and if error occurred - show error in this modal.
2) Do not open modal window before request will be fulfilled. Instead of this add some loading indicator to table cell and when request will be fulfilled - show modal with design/show alert error.
3) Leave as it is now.

What do you think about this?

anthonysena:
> And i have a question how will be better to implement error handing when loading of design failed. For now, i added alert to notify users that loading of design failed ( https://github.com/OHDSI/Atlas/pull/2118/files#diff-77e919ff16fbf32d1d863df621f0eb70R178 ). Maybe it will better to implement one of this:
> 
> 1. Open modal window with loading status(as now) and if error occurred - show error in this modal.
> 2. Do not open modal window before request will be fulfilled. Instead of this add some loading indicator to table cell and when request will be fulfilled - show modal with design/show alert error.
> 3. Leave as it is now.
> What do you think about this?

I think that using the alert mechanism as you have done in this PR is fine for now.

js/pages/characterizations/components/characterizations/characterization-view-edit/characterization-exec-wrapper.js
```diff
@@ -3,20 +3,37 @@ define([
...
 
             this.executionId = params.executionId;
-            this.componentParams = params;
+            const extraExecutionPermissions = ko.computed(() =>!params.designDirtyFlag().isDirty() && params.design().cohorts().length);
+            this.componentParams = {
+                tableColumns: ['Date', 'Design', 'Status', 'Duration', 'Results'],
+                runExecutionInParallel: false,
+                resultsPathPrefix: '/cc/characterizations/',
+                analysisId: params.characterizationId,
+                ExecutionService: CharacterizationService,
```
anthonysena, line 31: The `CharacterizationService` is lacking a 'generate' function which then causes a failure when attempting to generate a characterization.

PR #2119, opened 2020-01-31
checking of configuration jobs was changed as list of jobs now can be filtered 

ssuvorov-fls:
https://github.com/OHDSI/WebAPI/issues/1308

### DIFF
--- a/js/pages/configuration/configuration.js
+++ b/js/pages/configuration/configuration.js
@@ -10,6 +10,7 @@ define([
   'atlas-state',
   'const',
   'services/JobDetailsService',
+  'services/job/jobDetail',
   'services/Poll',
   'services/CacheAPI',
   'less!./configuration.less',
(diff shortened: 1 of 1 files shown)


js/pages/configuration/configuration.js
```diff
@@ -89,8 +91,16 @@ define([
       return this.sourceJobs.get(job.executionId);
     }
 
-    checkJobs() {
-      this.jobListing().forEach(job => {
+    async checkJobs() {
+      const notifications = await jobDetailsService.list([]);
```
anton-abushkevich, line 95: was: this.jobListing() - list of jobs from sharedState.jobListing.
now: jobDetailsService.list([]) which requests the same list only directly from backend.

It doesn't fix the problem. Either way, list doesn't contain "refresh jobs", started from configuration page by clicking on "refresh" button.

anton-abushkevich (changes requested):

PR #2122, opened 2020-02-04
Implements distribution of a field values in CC

wivern:
Resolves OHDSI/WebAPI#1220


anthonysena (dismissed):
image

Tagging @chrisknoll - I think we'll need to review this together since I'm unsure what had been discussed around this particular feature and how you envisioned this working. I have some ideas after reviewing these PRs but would appreciate your input.

chrisknoll:
Hi, all, After latest merges, something is wrong when trying to create new distribution criteria:
1. Click new Feature Analysis
2. Change analysis type to Distribution
3. Try to click 'Add New Criteria'.  No dropdown appears.

Note: I had to revert a few of the flyway migrations on my local side in order to get the latest migration script applied, I believe I did it correctly, but I will need to backtrack and see if I missed something. I'd like to ask if somoene else can pull down the latest PR and see if it is working for them.

wivern:
@chrisknoll I've fixed the appearance of `Distribute by` field. Now it showed only when distribution kind of feature analysis is selected.
It seems this behavior was lost during some merge from master.

chrisknoll:
One outstanding item on the UI per the latest changes on the WebAPI is the choices that are available on the Distribution UI:  The visit options of 'inpatient, outpatient, etc' was removed, and the 'length of era' was removed from both drug eara and condition ea because those are both handled by the 'duration' option available to all distribution types.   If someone can make that modification to the front end, that would be helpful, or if you can point me to where the options are stored, that would be helpful for me. Thanks.

anton-abushkevich (approved):

PR #2137, opened 2020-02-26
added confirmation when concept set expression is not empty

### LINKED ISSUE #669: importing concept set from repository overwrites concept set expression
As per @pbr6cornell - when in a cohort definition, on the concept set tab there is now an ability to import a concept set from the repository.  Doing so will overwrite the currently selected concept set with the one chosen.  

If this is the expected behavior then perhaps we rename the button with 'Replace' otherwise we should provide a warning to the user of what they are about to do.

ssuvorov-fls:
fixes #669


wivern:
There is a [PR](https://github.com/OHDSI/Atlas/pull/2132) that introduces the Concept Set editor as a separate component.
My suggestion is to postpone this until the mentioned PR merged.

PR #2180, opened 2020-04-06
added validation and highlighting of the required parameters

### LINKED ISSUE #2044: ATLAS designs / analyses must highlight required parameters before allowing users clicking generate / execute buttons
### Expected behavior
All ATLAS analyses (cohorts, characterization, IR, PLE, PLP, TxPathways) should highlight required parameters before allowing users clicking generate buttons. If user has not entered everything, he should be guided to complete design

I would propose to have a flag, or check that would allow to return analysis status as "valid/invalid" as well - to be by execution automation

### Actual behavior
In many cases, the analyses allow users to click generate button even if required parameters have not been completed. The execution says "completed"  but it actually fails - not obvious and the only way to see it is to decipher logs. So, exception handling should be documented and consistent across ATLAS as well (will create another GitHub issue)

image

image

image

@pbr6cornell  - this is causing a lot of headaches for many people today, I propose we target it in 2.8 but will seek your confirmation since it will impact our scope

ssuvorov-fls:
fixes #2044


js/pages/checks/style.css
```diff
@@ -0,0 +1,34 @@
+.warnings .fa-info {
```
wivern, line 1: Why component placed at pages rather than components dir? Looks like we can make it a true component formatted as other components

js/pages/checks/warnings.html
```diff
@@ -1,10 +1,10 @@
 <div role="tabpanel" class="warnings-button-pane pull-right">
```
wivern, line 1: Pls, replace classes with BEM

js/pages/checks/warnings.js
```diff
@@ -1,27 +1,29 @@
...
     'css!./style.css',
   ],
-  function (ko, view, cohortDefinitionApi, consts, utils, sharedState) {
-
-    function conceptSetWarnings(params){
-      var self = this;
-      this.currentCohortDefinition = sharedState.CohortDefinition.current;
-      self.cohortDefinitionId = this.currentCohortDefinition().id || ko.observable(-1);
-      self.count = params.count || ko.observable();
+  function (ko, view, consts, utils, sharedState) {
```
wivern, line 10: Why not to replace component function with class during such changes

js/pages/checks/warnings.js
```diff
@@ -1,27 +1,29 @@
...
+
+    function warningComponent(params){
+      const self = this;
+      self.current = params.current;
```
wivern, line 14: regarding comment above we don't need the `self` refs

js/pages/incidence-rates/ir-manager.css
```diff
@@ -0,0 +1,29 @@
+.nav-tabs li span.warning-alarm, .nav-tabs li span.warning-alarm:hover {
```
wivern, line 1: Why we need to duplicate the same styles for each page that uses warnings?

js/pages/pathways/components/manager.html
```diff
@@ -45,7 +45,15 @@
 				key: 'utilities',
 				componentName: 'pathway-utils',
 				componentParams: $component.componentParams,
-			}
+			},
+            {
+                title: 'Messages', 
+                key: 'messages',
+                componentName: 'warnings',
+                componentParams: $component.componentParams,
```
wivern, line 53: Pls note each tab has it's own `componentParams`, therefore we don't obliged to translate the whole set of parameters to every tab

js/pages/pathways/components/manager.js
```diff
@@ -66,12 +67,21 @@ define([
...
 
 			this.selectedTabKey = ko.observable("design");
-			this.componentParams = {
+			this.componentParams = ko.observable({
 				design: this.design,
 				analysisId: this.analysisId,
 				executionId: this.executionId,
 				dirtyFlag: this.dirtyFlag,
-			};
+				warningsTotal: ko.observable(0),
```
wivern, line 75: Same as above regarding param transmissions

js/services/Estimation.js
```diff
@@ -91,6 +103,8 @@ define(function (require, exports) {
 		generate,
 		listGenerations,
 		exists,
+        getWarnings,
```
wivern, line 106: formatting

js/services/Prediction.js
```diff
@@ -80,8 +80,20 @@ define(function (require, exports) {
...
+ 
+	function runDiagnostics(id, design) {
+		return httpService
+			.doPost(`${config.webAPIRoot}${predictionEndpoint}${id ||'-1'}/check`, design)
+			.then(res => res.data);
+	}
+
 
-    var api = {
+	var api = {
```
wivern, line 96: `const` or even `return`

wivern:
Additionally I have some questions regarding behavior:

1. Run check immediately after changes doesn't display warnings
2. Generation message
Prediction
3. Model settings empty - no warning, but execution is disabled - ???
4. Execution settings illegal patients value - the same as above
5. Training settings - an empty percentage value

js/services/Estimation.js
ssuvorov-fls: fixed

js/services/Prediction.js
ssuvorov-fls: fixed

js/pages/pathways/components/manager.html
ssuvorov-fls: fixed

js/pages/pathways/components/manager.js
ssuvorov-fls: fixed

js/pages/checks/warnings.js
ssuvorov-fls: fixed

js/pages/checks/warnings.js
ssuvorov-fls: fixed

js/pages/checks/style.css
ssuvorov-fls: fixed

js/pages/checks/warnings.html
ssuvorov-fls: fixed

PR #2183, opened 2020-04-14
merge geo spatial into 2.8

ssuvorov-fls:
fixes https://github.com/OHDSI/WebAPI/issues/1445


js/config/app.js
```diff
@@ -9,7 +9,7 @@ define(function () {
   appConfig.cacheSources = false;
   appConfig.pollInterval = 60000;
 	appConfig.cohortComparisonResultsEnabled = false;
-	appConfig.userAuthenticationEnabled = false;
+	appConfig.userAuthenticationEnabled = true;
```
anthonysena, line 12: Let's not change the default configuration for security

ssuvorov-fls: You're right.
Fixed

anthonysena:
@ssuvorov-fls @wivern : A couple of questions/comments as I'm reviewing this functionality:

- Thank you for adding the documentation to the README file. I'd suggest that you create a separate README file in `js/gis` and include that information in that folder since it is specific to the gis functionality and does not need to be on the main home page. We can also add a link to this in the ATLAS wiki. Also there is still a typo in the heading that reads "Geo spatial" which should be all one word.
- Why does this PR contain presumably the same files contained in https://github.com/OHDSI/atlas-component-geospatial/? I thought the idea here is that if you want to enable this plugin, you download from the other repo and set it up in ATLAS?
- `js/config-gis.js' references local libraries with hard-coded paths to localhost. Is this intended? 
- `js/gis/config-gis.js' has a "TODO" comment - please address this or remove the comment. 
- `js/gis/config-gis.js' references a `tilesServerUrl` but I do not see a corresponding setting in `js/config-gis.js`.

PR #2196, opened 2020-04-27
Removed a call to 'check' endpoint during deleting of cohort definition

### LINKED ISSUE #2172: Deleting a cohort definition calls the 'check' endpoint
### Expected behavior

When deleting a cohort definition, no REST errors are returned.

### Actual behavior

When deleting a cohort definition, the endpoint `/WebAPI/cohortdefinition/<id>/check` is called after the entity is deleted causing an error.

With chrome developer tools open, delete a cohort and observe the error.

ssuvorov-fls:
fixes #2172

### DIFF
--- a/js/pages/cohort-definitions/components/checks/conceptset-warnings.js
+++ b/js/pages/cohort-definitions/components/checks/conceptset-warnings.js
@@ -124,7 +124,7 @@ define(['knockout', 'text!./conceptset-warnings.html',
         }
       };
 
-      dispose = function() {
+      dispose() {
         this.warningSubscription.dispose();
       }
     }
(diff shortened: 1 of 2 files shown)


js/pages/cohort-definitions/components/checks/conceptset-warnings.js
```diff
@@ -91,19 +91,23 @@ define(['knockout', 'text!./conceptset-warnings.html',
       };
 
       self.getWarnings = function() {
-        if (parseInt(self.cohortDefinitionId(), 10) <= 0 || isNaN(self.cohortDefinitionId())) {
-          return false;
+        if (this.cohortDefinitionId()) {
```
vlbe, line 94: cohortDefinitionId becomes undefined after you try to save newly created concept so the check will not fire any more after changing the cohort definition

ssuvorov-fls: fixed

Issue #2191, opened 2020-04-24
Cohort Definitions: Remove the ability to change a design if you lack permissions


olga-ganina:
Parent task #2174 

### Expected behavior
If a given entity is locked since I do not have rights, I should not be able to make changes that would require a save action.

### Actual behavior
Cohort Definitions: I can change "Included Concepts" for a concept set I do not own included in a cohort definition

PR #2200, opened 2020-04-29
Disabled concepts in cohort if lack of permissions

aklochkova:
Fixes https://github.com/OHDSI/Atlas/issues/2191

### DIFF
--- a/js/const.js
+++ b/js/const.js
@@ -45,7 +45,7 @@ define([
 		
 		const getIncludedConceptsColumns = (sharedState, context, commonUtils, conceptSetService) => [
 			{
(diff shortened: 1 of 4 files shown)


js/const.js
```diff
@@ -42,6 +42,72 @@ define([
 				}
 			}]
 		};
+		
+		const getIncludedConceptsColumns = (sharedState, context, commonUtils, conceptSetService) => [
```
vlbe, line 46: I think it is the same as in this PR https://github.com/OHDSI/Atlas/pull/2199/files. Canyou combine them into one constant for use on both pages?

Issue #2194, opened 2020-04-24
Incident Rates: Remove the ability to change a design if you lack permissions


olga-ganina:
Parent task #2174 

### Expected behavior
If a given entity is locked since I do not have rights, I should not be able to make changes that would require a save action.

### Actual behavior
IR: Target/Outcome cohort lists should disable the ability to remove. The Time At Risk settings should be read-only. The Add Study Window/New stratify buttons should be disabled. On the concept set tab, the buttons should be disabled. The generation tab should ensure all buttons are disabled to remove results and to re-generate.

aklochkova:
@anthonysena in my PR I didn't disable `re-generate` and `generate` buttons but only hid a `remove` one. Users can't run a generation with a changed design before it's saved and they can't save it due to lack of permissions. So it's safe from changes.
Is it enough or should I hide generation buttons and add new permissions to WebAPI to protect it?

aklochkova:
@anthonysena could you please review my previous comment about `generate` buttons?

anthonysena:
Let's handle it in the review of #2220 - I suspect your approach will be fine as you described.

PR #2220, opened 2020-05-28
Disabled incidence rate components if lack of permissions

aklochkova:
Fixed https://github.com/OHDSI/Atlas/issues/2194


anton-abushkevich:
@anthonysena Due to changes in OHDSI/WebAPI#1503 any IR is now can be edited by anyone. Therefore, it is impossible now without DB hack to reproduce a lack of permissions for Atlas user to edit IR.

Question is: OHDSI/WebAPI#1503 is about "Role "atlas user" lacks rights to create incidence rates", but for creation there is another permission - ir:post, not ir:*:put. And for "Atlas user" ir:post was there from the very beginning. I suggest to reopen OHDSI/WebAPI#1503, make sure "Atlas user" has the rights to create IR without ir:*:put (and if not - research why), and rollback those migrations.

anthonysena:
> Question is: OHDSI/WebAPI#1503 is about "Role "atlas user" lacks rights to create incidence rates", but for creation there is another permission - ir:post, not ir:*:put. And for "Atlas user" ir:post was there from the very beginning. I suggest to reopen OHDSI/WebAPI#1503, make sure "Atlas user" has the rights to create IR without ir:*:put (and if not - research why), and rollback those migrations.

@anton-abushkevich - sorry for this regression. I agree with this approach.

PR #2222, opened 2020-05-29
Fix loading cohort definition

### LINKED ISSUE #2221: Fix loading cohort definition

ssuvorov-fls:
fixes #2221

### DIFF
--- a/js/services/CohortDefinition.js
+++ b/js/services/CohortDefinition.js
@@ -60,7 +60,7 @@ define(function (require, exports) {
 	function getCohortDefinition(id) {
 		return httpService
 			.doGet(config.webAPIRoot + 'cohortdefinition/' + id)
-			.then(res => JSON.parse(res.expression))
+			.then(res => res.data)
 			.catch(error => {
 				console.log("Error: " + error);
 				authApi.handleAccessDenied(error);


chrisknoll:
This PR doesn't work:  the result from the web service stores the expression as a string in the cohort definition result, meaning that you need to parseJson() from the response before passing it back to the caller.

The confusion here is that originally: cohort definitions wrapped an 'expression' as a string, with the intent that a cohort definition could be defined by a JSON expression, or a SQL statement, etc.  Those other modes (the nonJSON types) never materialized, but the interface to the API was to return a string.

There was a PR around the time treatment pathways and CC were introduced that broke this rule, and I had to implement a backwards-compatible work-around where the endpoint /cohortdefinition/{id} returns a string for the expression (seen [here](https://github.com/OHDSI/WebAPI/blob/d6e010c0e1573c01c9df03e3966ba38be2bfed0e/src/main/java/org/ohdsi/webapi/service/CohortDefinitionService.java#L416)),  Note the use of CohortRawDTO.  Other elements of the application (that performs exports) will return a CohortDTO with an expression of type 'CohortExpression'.

Details of this change can be found here: https://github.com/OHDSI/WebAPI/pull/1261

The fix for this PR is to just parse the expression string into a JSON object before returning the caller.

Issue #2193, opened 2020-04-24
Pathways: Remove the ability to change a design if you lack permissions


olga-ganina:
Parent task #2174 

### Expected behavior
If a given entity is locked since I do not have rights, I should not be able to make changes that would require a save action.

### Actual behavior
Pathways: the cohort definitions, feature analyses lists contain "Remove" links. The buttons to Add/Import things into the design should be disabled. Additionally, the options at the bottom should be read-only.

PR #2225, opened 2020-06-02
Disabled Pathway and Feature analyses components if lack of permissions

aklochkova:
Fixes https://github.com/OHDSI/Atlas/issues/2193

### DIFF
--- a/js/pages/characterizations/components/feature-analyses/feature-analysis-view-edit.html
+++ b/js/pages/characterizations/components/feature-analyses/feature-analysis-view-edit.html
@@ -27,98 +27,99 @@
         </div>
     </div>
 
-    <div data-bind="css: {'content': true, 'fe-design-disabled': !$component.canEdit()}, visible: !$component.loading()">
+    <div data-bind="css: {'fe-design-disabled': !$component.canEdit()}">
+        <div data-bind="css: classes('content'), visible: !$component.loading()">
 
-        <div data-bind="css: classes({ element: 'descr-panel', extra: 'panel panel-primary' })">
-            <div class="panel-heading">
-                Description
-            </div>
-            <div data-bind="css: classes({ element: 'descr-panel-body' })">
-                <!-- ko if: $component.canEdit() -->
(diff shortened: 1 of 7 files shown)


js/pages/characterizations/components/feature-analyses/feature-analysis-view-edit.html
```diff
@@ -27,7 +27,7 @@
         </div>
     </div>
 
-    <div data-bind="css: classes('content'), visible: !$component.loading()">
+    <div data-bind="css: {'content': true, 'fe-design-disabled': !$component.canEdit()}, visible: !$component.loading()">
```
wivern, line 30: Take a look, `classes('content')` isn't the same as `'content': true`, this change breaks the panel style

js/pages/characterizations/components/feature-analyses/feature-analysis-view-edit.less
```diff
@@ -110,4 +110,15 @@
...
     margin-bottom: 1rem;
   }
+}
+
+
+.fe-design-disabled {
+  pointer-events: none;
+  opacity: 0.7;
+
+  .btn.btn-primary {
```
wivern, line 120: Could you describe what is the goal of this construction? Not sure CSS selector used in a right way

js/pages/characterizations/components/feature-analyses/feature-analysis-view-edit.less
```diff
@@ -110,4 +110,15 @@
     font-weight: 500;
     margin-bottom: 1rem;
   }
+}
+
+
+.fe-design-disabled {
+  pointer-events: none;
+  opacity: 0.7;
```
wivern, line 118: It's better to remove opacity from whole component since this makes harder to read large analyses.
Let's just prevent event handling and opacity for buttons.

js/pages/characterizations/components/feature-analyses/feature-analysis-view-edit.less
aklochkova: it simulates the css style 'disabled', if I make the button just more transparent its color is too light

Issue #2151, opened 2020-03-11
cohort definition text view does not show terms "All", "any", etc


jamieweaver:
### Expected behavior
Cohort definition text view shows:

Inclusion Criteria #1: age >= 18
Having **all** of the following criteria:
with the following event criteria:
with age >= 18s:

### Actual behavior
Cohort definition text view shows:

Inclusion Criteria #1: age >= 18
Having [object Object] of the following criteria:
with the following event criteria:
with age >= 18

### Steps to reproduce behavior
View a cohort definition in text view

PR #2226, opened 2020-06-02
Parse raw cohort definition's expression.

chrisknoll:
Fixes PR #2222 by parsing the raw JSON value in the CohortDefinition.expression.


chrisknoll:
@anton-abushkevich , since you approved the other PR, could you review this one and ensure that it works in your test case that you used to review the other PR?  I'm concerned that this change would break the test case that you used in the other PR (assuming the other PR worked) and want to understand if I missed something.

chrisknoll:
While I was at it, I fixed #2151 in this PR.

Issue #2182, opened 2020-04-09
Restrict use of characters not part of windows/linux file names


gowthamrao:
### Expected behavior
Dont accept atlas names (cohort, concept set) that has '/' in it.

### Actual behavior
Allows names

### Steps to reproduce behavior
create a cohort definition and give it a name with '/'

### why is this important
we are increasing the use of tools like ROhdsiWebApi to pull definition names to local. If a name has '/' in it, it will throw an error because Windows cannot create file names with '/' in it.

This is a known issue and right now we are just being careful. it would be nice to not allow characters that are not accepted by windows.

anthonysena:
Linking this to OHDSI/WebAPI#1510 since these both touch on ways that we want to format/restrict the name of assets in the system. Additionally, I believe there is a length limit for file names when exported that should be considered here.

gowthamrao:
In addition - recommend creating a max length to the file name.
- we commonly create files that use the names of the cohort. e.g. fileName.json or fileName.sql . 
- These names should not only be valid windows/linux file names, but should also be not more than a certain reasonable file length, as defined by number of characters. 
- I dont know what the max number of characters should be - but it should allow for adding '.json' and '.sql' to the end + maybe five characters for adding any prefix/suffix to the webApi file name.

PR #2232, opened 2020-06-09
Added length and symbol check for an analysis name

aklochkova:
Fixes https://github.com/OHDSI/Atlas/issues/2182

### DIFF
--- a/js/const.js
+++ b/js/const.js
@@ -109,21 +109,6 @@ define([
 			];
 		};
 		
-		const isValidName = function(name) {
-			
-			const forbiddenSymbols = ['\\', '/', ':', '*', '?', '"', '<', '>', '|'];
-			const maxFileNameLength = 100;			
-			if (forbiddenSymbols.some(symbol => name.includes(symbol))) {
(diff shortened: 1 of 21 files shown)


js/const.js
```diff
@@ -109,6 +109,21 @@ define([
 			];
 		};
 		
+		const isValidName = function(name) {
```
wivern, line 112: Could you move it to the `CommonUtils` which looks more suitable?

js/pages/characterizations/components/characterizations/characterization-view-edit.js
```diff
@@ -189,6 +189,9 @@ define([
             // Next check to see that a characterization with this name does not already exist
             // in the database. Also pass the id so we can make sure that the current characterization is excluded in this check.
             try {
+                if (!constants.isValidName(this.design().name())) {
```
wivern, line 192: In general, best practices tells the good function should have a single entry point and single out.
Regarding this case it's better to use the following sample:
```
if (isValidName(someName)) {
  ...
}
```

Issue #2117, opened 2020-01-27
Provide ability to export pathway sunburst plots and data 


jhardin29:
### Requested behavior
it would be super helpful if we could export the sunburst plots (jpg, bmp?) with all of the drugs included via an export button.
Presently the mechanism to include the plots is to cut and paste manually

anthonysena:
@jhardin29 - I'm hijacking this issue a bit - completely agree that having the ability to export the pathway image is useful. Additionally, we do not currently provide a mechanism to export the underlying data set that is used to construct the pathway sunburst which is equally as valuable for reporting.

Issue #1554, opened 2019-03-18
Cohort Pathways - Save Sunburst as Figure


davekern:
### Expected behavior
Save the sunburst as a jpg or other figure format so that it can be added to external reports.

### Actual behavior
Cannot export the sunburst figure

Issue #1552, opened 2019-03-18
Cohort Pathways - Download pathways data


davekern:
### Expected behavior
Ability to download the pathways data so that additional analysis can be performed by user

### Actual behavior
Unable to do this

chrisknoll:
I am providing a tabular view of the pathway data, that lets you see the paths rolled up based on a user-selected 'Maximum Path Length':
![image](https://user-images.githubusercontent.com/6818777/55365982-bb04a100-54b4-11e9-9105-0b064d59111c.png)

By setting the path length lower, you can see a roll-up of fewer paths:
![image](https://user-images.githubusercontent.com/6818777/55366041-05861d80-54b5-11e9-9c19-3721fa30027f.png)

You can even set the max-paths to '1' to see the first-line path summary:
![image](https://user-images.githubusercontent.com/6818777/55366109-51d15d80-54b5-11e9-9e40-12b3fc829265.png)

Clicking 'export' above each report group will export the same columns displayed in the report.

If there are no objections to this approach, I can work on a PR.

chrisknoll:
Closed in #2235

Issue #1451, opened 2019-03-07
Cohort pathways - export results - enhancement request


jhardin29:
Could there be a feature similar to what is in cohort characterization (in beta) to allow the user to export the results of the pathway analysis? 

The following columns would be useful: 

Treatment (Exposure)	Treatment line	Rank in tx line	Patient count	% of all pts in cohort*	Cnt of patients in tx line	% of pts w/exposure in that tx line*

Treatment   line | Count of those having at least   this line of therapy | % of all patients in cohort | % of treated

anthonysena:
To add to Jill's note above, here are the tables (the table # and headings are geared toward drugs and need to be generalized to "events"). Just put any questions or items that need clarification into this issue. Thanks!

----


### Table 1a. Patient counts by treatment within each line of therapy

Treatment (Exposure) | Treatment line | Rank in tx line | Patient count | % of all pts in cohort* | Cnt of patients in tx line | % of pts w/exposure in that tx   line
-- | -- | -- | -- | -- | -- | --

### Table 1b.  Patient counts by treatment line

Treatment   line | Count of those having at least this line of therapy | % of all patients in cohort | % of treated
-- | -- | -- | --


### Table 1c. Patient counts by final treatment line (mutually exclusive)

  | Count | % of all patients in cohort
-- | -- | --
Untreated | 0 | 0%
Of the treated |   |  
Exactly 1 tx | 0 | 
Exactly 2 | 0 | 
Exactly 3 | 0 | 
Exactly 4 | 0 | 
Exactly 5 | 0 | 
Exactly 6 | 0 | 
Exactly 7 | 0 | 
Exactly 8 | 0 | 
Exactly 9 | 0 |

johnSamilin:
Result page have 3 panels: Legend, Sunburst plot and Path details.
Do I understand it correctly that there should be also 3 separate buttons somewhere which will get Pathway analysis data, transform it and export as csv? Or it should be one button that will generate and export a zip file with 3 csv files?

pavgra:
I believe, the desired result is the 3 CSVs in one zip by a single button click and this more about back-end. @anthonysena, any concerns?

ssuvorov-fls:
Hello.
Sorry, I don't know much about the subject area yet. Could you give me a screenshot with indication which value from it should be substituted into the tables columns?

anthonysena:
@ssuvorov-fls - let me follow up with @jhardin29 to make sure we've got clear requirements for this item.

[...]

PR #2235, opened 2020-06-11
Implementation of exporting pathway analysis data.

chrisknoll:
References issues: #2117, #1552, #1554. #1451.

This PR's primary focus is #1552 (I gave this a bad branch name), where a tabular view of the sunburst is presented and can be exported one at a time.  I think it would be useful to support a 'export all' button but I did not get to that point.

Other issues I would like to address, but could use some help: 
#1554/#2117: I thought we had the ability to grab a SVG from the browser and save it as a PNG (like we can save a .csv file from tabular data) but I can't find where we did this.  If anyone can provide a function to do this, we can make exporting the sunburst SVG part of this PR.
_**Edit**_: I think I found it in `ChartUtils.downloadAsPng()`.

#1451/#1557: This request is more of a summary statistics view of the sunburst data, and I have some ideas about implementing this, and I'm happy to put it into this PR if time allows.

### DIFF
--- a/js/components/Chart.js
+++ b/js/components/Chart.js
@@ -4,17 +4,22 @@ define([
 	'utils/AutoBind',
 	'utils/ChartUtils',
   'const',
+	'utils/BemHelper',	
   'components/empty-state',
+	'less!./charts/chart.less'
 ], function (
   ko,
   Component,
 	AutoBind,
 	ChartUtils,
   constants,
+	BemHelper
 ) {
   class Chart extends AutoBind(Component) {
     constructor(params, container) {
       super(params);
+			const bemHelper = new BemHelper('Chart');
+      this.chartClasses = bemHelper.run.bind(bemHelper);
       this.renderer = null; // atlascharts
       this.rawData = ko.observable();
       this.format = {};
--- a/js/components/Chart.js
+++ b/js/components/Chart.js
@@ -1,15 +1,19 @@
 define([
   'knockout',
   'components/Component',
+	'utils/AutoBind',
+	'utils/ChartUtils',
   'const',
   'components/empty-state',
 ], function (
   ko,
   Component,
+	AutoBind,
+	ChartUtils,
   constants,
 ) {
-  class Chart extends Component {
-    constructor(params) {
+  class Chart extends AutoBind(Component) {
+    constructor(params, container) {
       super(params);
       this.renderer = null; // atlascharts
       this.rawData = ko.observable();
@@ -19,6 +23,8 @@ define([
       this.data = ko.computed(() => {
         return this.prepareData(this.rawData());
       });
+			this.container = container;
+			this.filename = params.filename || 'untitledChart.png';
     }
 
     prepareData(rawData) {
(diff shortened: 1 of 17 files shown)


chrisknoll:
I've added the functionality to export the SVG as PNG. This implementation [from here](http://bl.ocks.org/Rokotyan/0556f8facbaf344507cdc45dc3622177) is replacing the existing `ChartUtils.downloadAsPng()` since the new implementation handles external styles.

Also, all charts will now provide an export button if a paramater of `filename` is passed to the component.  Currently, I am only doing this for the Sunbursts on pathways.

chrisknoll:
This PR should be complete.  I implemented the tables defined in #1451.

Things I do not have time to implement:
- Server side export where all cohorts/all sources are combined into a single download.  Currently you need to export tables and images separately.

js/components/Chart.js
```diff
@@ -30,6 +41,12 @@ define([
       this.format = params.format;
     }
 		
+		export() {
+			console.warn('Export not implemented');
```
anthonysena, line 45: I'm not sure why this warning is here? It looks like the export is implemented - maybe it was a placeholder?

js/extensions/bindings/dynamicDatatableBinding.js
```diff
@@ -0,0 +1,13 @@
+define(['jquery', 'knockout', './datatableBinding'], function($,ko) {
```
anthonysena, line 1: I'm not really sure I understand the purpose of this binding - could you help me to understand? I see it is used in the pathways-tableview component but not sure why. No objections to it - just trying to understand.

js/components/Chart.js
chrisknoll: My bad.  I thought at one point that subclasses would need to implement their own export, but later realized it can be a Chart-level function. I'll remove this.

js/extensions/bindings/dynamicDatatableBinding.js
chrisknoll: Yes. Fun story:  I spent a lot of time trying to re-define the columnDef and columns using the normal 'datatableBinding', but could not get it to work.  Then I found information on the datatables.net forums: https://datatables.net/forums/discussion/34654/add-or-remove-column-dynamically.  This thread calls for calling `destroy()` on the datatable if you are going to make those types of changes.  In the case of the pathway table being dynamic (based on number of steps you want to show), i needed to destroy the table.  I didn't want to hack into our existing datatable binding (it's complicated enough), so instead I 'extended' the current datatableBinding, called it `dynamicDatatableBinding`,  and the entire table is destroyed when the underlying datatable observable (including column defs) changes.

anthonysena: Sorry to bring up these painful coding memories during the review! But this is helpful to know and makes sense given the dynamic nature of that data table.

chrisknoll: No worries, was a fair question, and it is good to document the reasoning behind how it was implemented.  Keep 'em coming.

PR #2243, opened 2020-07-06
Added authorship to design header

### LINKED ISSUE #2170: Design heading when security is enabled
When security is enabled, a user is not allowed to make changes to a design unless they are the owner or have been granted permission. This is not clearly stated in the interface. 

One idea is to adopt the approach done in cohort definitions and use this across the other designs in the application: 

image

Adding the owner + date created for all design entities would help to identify the owner properly. Additionally, we should add some text near the design identifier to show when a design is read-only such as:

**Design #x** _(Read only)_

vlbe:
Closes #2170

### DIFF
--- a/js/pages/concept-sets/conceptset-manager.html
+++ b/js/pages/concept-sets/conceptset-manager.html
@@ -4,7 +4,9 @@
 
 <!-- ko if: currentConceptSet() != null && !loading() -->
 <heading-title params="name: conceptSetCaption(), icon: 'shopping-cart', theme: 'dark'"></heading-title>
-
+<!-- ko if: currentConceptSet().id != 0 -->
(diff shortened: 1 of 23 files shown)


js/pages/concept-sets/conceptset-manager.html
```diff
@@ -4,7 +4,9 @@
 
 <!-- ko if: currentConceptSet() != null && !loading() -->
 <heading-title params="name: conceptSetCaption(), icon: 'shopping-cart', theme: 'dark'"></heading-title>
-
+<!-- ko if: currentConceptSet().id != 0 -->
```
anton-abushkevich, line 7: everywhere else it is double exclamation mark: .
why dont do the same here?

PR #2267, opened 2020-08-03
Fixes issue with opening conceptSet after relogin

wivern:
Fixes OHDSI/WebAPI#1595

### DIFF
--- a/js/components/atlas-state.js
+++ b/js/components/atlas-state.js
@@ -1,12 +1,22 @@
 define(['knockout', 'lscache', 'services/job/jobDetail', 'assets/ohdsi.util', 'const'], function (ko, cache, jobDetail, ohdsiUtil, constants) {
 	var state = {};
-	state.resultsUrl = ko.observable();
-	state.vocabularyUrl = ko.observable();
-	state.evidenceUrl = ko.observable();
 	state.jobListing = ko.observableArray();
 	state.priorityScope = ko.observable('session');
 	state.roles = ko.observableArray();
 	state.sources = ko.observableArray([]);
+
+	state.vocabularyUrl = ko.observable(sessionStorage.vocabularyUrl);
+	state.evidenceUrl = ko.observable(sessionStorage.evidenceUrl);
+	state.resultsUrl = ko.observable(sessionStorage.resultsUrl);
+	state.vocabularyUrl.subscribe(value => sessionStorage.vocabularyUrl = value);
+	state.evidenceUrl.subscribe(value => sessionStorage.evidenceUrl = value);
+	state.resultsUrl.subscribe(value => sessionStorage.resultsUrl = value);
+
+	state.resetCurrentDataSourceScope = function() {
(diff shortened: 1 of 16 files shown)


js/components/welcome.js
```diff
@@ -3,6 +3,7 @@ define([
     'text!./welcome.html',
     'appConfig',
     'services/AuthAPI',
+	  'services/SourceAPI',
```
YaroslavTir, line 6: formating

js/components/welcome.js
```diff
@@ -71,6 +73,9 @@ define([
 
         self.onLoginSuccessful = function(data, textStatus, jqXHR) {
             sharedState.resetCurrentDataSourceScope();
+            // if (!sharedState.vocabularyUrl()) {
+            //     await sourceApi.initSourcesConfig();
+            // }
```
YaroslavTir, line 78: :+1:

js/services/ConceptSet.js
```diff
@@ -62,7 +62,8 @@ define(function (require, exports) {
 	}
 
 	function loadConceptSetExpression(conceptSetId) {
-		return httpService.doGet(config.api.url + 'conceptset/' + conceptSetId + '/expression/' + sharedState.sourceKeyOfVocabUrl()).then(({ data }) => data);
+		const sourceKey = sharedState.sourceKeyOfVocabUrl();
+		return httpService.doGet(config.api.url + 'conceptset/' + conceptSetId + '/expression' + (sourceKey ? `/${sourceKey}`: '')).then(({ data }) => data);
 	}
```
YaroslavTir, line 67: this is used to be `/expression/` but now just '/expression'. Is this correct?

js/components/atlas-state.js
```diff
@@ -8,14 +8,22 @@ define(['knockout', 'lscache', 'services/job/jobDetail', 'assets/ohdsi.util', 'c
...
+	state.defaultResultsUrl.subscribe((value) => state.resultsUrl(value));
 
 	state.resetCurrentDataSourceScope = function() {
-		state.resultsUrl(null);
-		state.vocabularyUrl(null);
-		state.evidenceUrl(null);
+		state.resultsUrl(state.defaultResultsUrl());
+		state.vocabularyUrl(state.defaultVocabularyUrl());
+		state.evidenceUrl(state.defaultEvidenceUrl());
 	}
```
YaroslavTir, line 27: can you keep the same order as above?
```
		state.vocabularyUrl(state.defaultVocabularyUrl());
		state.evidenceUrl(state.defaultEvidenceUrl());
		state.resultsUrl(state.defaultResultsUrl());
```

js/services/ConceptSet.js
wivern: It was resolved as `/expression/null` when `sourceKey` was `null`. And that's a mistake.
If talking regarding the tail slash, I should say the `/expression` and the `/expression/` has different meanings from HTTP's point of view.

PR #2270, opened 2020-08-05
Disabled hyperlinks for system jobs

### LINKED ISSUE #2168: Notifications area with security enabled
### Expected behavior

I can see notifications for my own jobs and throughout the system.

### Actual behavior

When security is disabled, the default behavior of the notifications area shows all completed jobs in the system which is useful for seeing what is running across the system. When security is enabled, this display defaults to showing the user's jobs only.

### Ideas to enhance this

Having a tabbed display that shows "my notifications" and "system notifications" so that we can view both when security is enabled. Additionally the "bell" icon should only activate when a person's jobs have finished versus any job in the system.

ssuvorov-fls:
fixes #2168

### DIFF
--- a/js/components/userbar/user-bar.js
+++ b/js/components/userbar/user-bar.js
@@ -65,6 +65,7 @@ define([
 			} else {
 				this.selectedTabKey(constants.jobTypes.ALL_JOB.title);
 				this.jobNotificationsPending = ko.computed(() => this.allJobParams.jobListing().filter(j => !j.viewed()).length);
+				this.allJobParams.jobNameClick = this.jobNameClick.bind(this);
 			}
 			this.tabs.push({
 				title: 'All jobs',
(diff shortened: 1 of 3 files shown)


js/components/userbar/user-bar.js
```diff
@@ -46,11 +46,13 @@ define([
 				jobListing: ko.computed(() => lodash.sortBy(this.jobListing(), el => -1 * el.executionId)
 					.filter( j => j.ownerType === constants.jobTypes.USER_JOB.ownerType)),
 				jobNameClick: this.jobNameClick.bind(this),
+				showAsHyperlinks: true,
 			};
 			this.allJobParams = {
 				jobListing: ko.computed(() => lodash.sortBy(this.jobListing(), el => -1 * el.executionId)
 					.filter( j => j.ownerType === constants.jobTypes.ALL_JOB.ownerType)),
 				jobNameClick: this.jobNameClick.bind(this),
```
anton-abushkevich, line 54: You do not need "jobNameClick" parameter if "showAsHyperlinks" is false. Maybe even use it as flag instead of additional parameter "showAsHyperlinks"?

PR #2280, opened 2020-08-13
Date is converted to string with local timezone

### LINKED ISSUE #2148: Occurrence end date decreases by one day after saving
Before saving:
image
After saving:
image

wrong date in PUT request:
{CodesetId: 0, OccurrenceStartDate: {Value: "2019-12-31", Extent: "2020-03-29"

vlbe:
Fixes #2148

### DIFF
--- a/js/components/cohortbuilder/InputTypes/Range.js
+++ b/js/components/cohortbuilder/InputTypes/Range.js
@@ -1,4 +1,4 @@
-define(['knockout'], function (ko) {
+define(['knockout', 'moment'], function (ko) {
 	var debug = false;
 
 	function Range(data) {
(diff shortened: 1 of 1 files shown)


chrisknoll:
Hi, @wivern  and @vlbe,
  Eventually the cohort builder will be extracted out to a stand-alone component, and I want to minimize external dependencies on this.. Looks like this fix introduced a dependency on moment, and I'm wondering if there's a way to address this problem without resorting to another dependency?  As I said in the original issue report: we did experience this problem dealing with certain timezones that would swap the mm/dd to the day earlier due to some wierd timezone offset, and I thought we worked out a way to have it just use a universal timezone and then use the verbatim mm/dd/yyyy that was entered.  I'd almost rather drop the use of the JQuery calendar control and just have the input box take a validated yyyy-mm-dd value....

chrisknoll:
Just found a probem with this PR where the import 'moment' was not assigned a variable to reference, and you get a 'moment not defined' error.

I'll file a PR for this.

Issue #2272, opened 2020-08-06
Concept Sets: Clicking checkbox in header clears out all checks in columns without all boxes checked


katy-sadowski:
### Expected behavior
When I check the checkbox in the header for Descendants, Mapped, or Exclude, it should not affect which boxes are checked in the other columns.

### Actual behavior
If a column has only some of its checkboxes checked and I then check or uncheck the header checkbox in another column, all of the boxes in that other column will be unchecked.  Note that columns with all of their checkboxes checked are _not_ affected by this issue.

### Steps to reproduce behavior
See attached video

[Screen Recording 2020-08-06 at 9.19.11 AM.zip](https://github.com/OHDSI/Atlas/files/5035032/Screen.Recording.2020-08-06.at.9.19.11.AM.zip)

chrisknoll:
Hi, thank you for your help on this, but with the changes to the concept set editor, I think this is out of date. But, I have made the changes to fix this relative to the new concept set editor, and so we can close this out.

PR #2309, opened 2020-09-07
issue-2272: When check the checkbox in the header for Descendants, Ma…

hongwonjun:
…pped, or Exclude, it does not affect which boxes are checked in the other columns. (#2272)


chrisknoll:
@hongwonjun , thank you for your PR, but I am not sure this code applies to the latest version of the concept set editor. However, we may not have implemented it properly, so would you mind resolving your conflicts with the latest commits in master?  It's possible your PR needs to change based on the new UI.

chrisknoll:
Hi, thank you for your help on this, but with the changes to the concept set editor, I think this is out of date. But, I have made the changes to fix this relative to the new concept set editor, and so we can close this out.

PR #2310, opened 2020-09-08
createPs is set to true when user specifies inverse probability of treatment weighting (IPTW) for the outcome model

### LINKED ISSUE #1657: Inverse probability treatment weighting bug
### Expected behavior

When user specifies inverse probability of treatment weighting (IPTW) for the outcome model, `createPs` should be set to true. 

### Actual behavior

`createPs` is not properly set which then causes an error when executing the study

ssuvorov-fls:
fixes #1657


anton-abushkevich:
If PLE has multiple Analysis Settings, "Messages" tab is not pointing to one that has aforementioned problem. And if erroneous analyses more than one, messages for them are identical:
изображение

anton-abushkevich (approved):

Issue #2334, opened 2020-09-23
Utilities tab is not available for new estimation analyses


KSadomtsev:
### Expected behavior
![image](https://user-images.githubusercontent.com/56290675/94009566-5080a880-fdad-11ea-8ad1-fa0e2ae029e1.png)

### Actual behavior
in url we can see: #/estimation/cca/0/utilities
![image](https://user-images.githubusercontent.com/56290675/94009523-3f379c00-fdad-11ea-93b6-1800bca6209b.png)

### Steps to reproduce behavior 
1)Estimation -> New Population Level Estimation Analysis button
2)Click to utilities tab

PR #2338, opened 2020-09-28
fixed error during saving estimation(#2334)

ssuvorov-fls:


anton-abushkevich:
Still reproducible. 

See js/components/utilities/export.js:74
```
        async loadExportJSON() {
            if (this.entityId() !== 0) {
                this.loading(true);
                const res = await this.exportService(this.entityId());
                this.exportEntity(res);
                this.loading(false);
            }
        }
```

Method should not call exportService because of `this.entityId()` is supposedly `0`. But it is not. `this.entityId()` in this case is string "0". Problem is in `sharedState.estimationAnalysis.selectedId`. It contains string "0" when creating a new entity.

anton-abushkevich (approved):

PR #2358, opened 2020-10-14
Get print-friendly from WebAPI service.

### LINKED ISSUE #1181: Add R Markdown export for cohort definitions
### Expected behavior
Provide an R markdown format friendly export.

### Actual behavior
Current text export is hard to integrate into R Markdown.

chrisknoll:
Fixes #1181.

### DIFF
--- a/js/pages/cohort-definitions/cohort-definition-manager.js
+++ b/js/pages/cohort-definitions/cohort-definition-manager.js
@@ -628,6 +628,10 @@ define(['jquery', 'knockout', 'text!./cohort-definition-manager.html',
 			});
 
 			this.pollForInfoPeriodically();
+
+			this.subscriptions.push(
+				this.currentCohortDefinition.subscribe(() => ("export" === this.tabMode()) && this.currentCohortDefinition() && this.refreshPrintFriendly())
+			);
 		}
 
 		// METHODS
@@ -1317,7 +1321,7 @@ define(['jquery', 'knockout', 'text!./cohort-definition-manager.html',
 				this.printFriendlyLoading(true);
 				try {
 					const printFriendlyHtml = await cohortDefinitionService.getCohortPrintFriendly(ko.toJS(this.currentCohortDefinition().expression()));
-					this.printFriendlyHtml(printFriendlyHtml);
+					this.printFriendlyHtml(printFriendlyHtml.data);
 				} catch(error) {
 					console.error("Problem loading print-friendly output.", error);
 				}finally {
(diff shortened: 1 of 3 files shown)


chrisknoll:
This may not be complete: need to decide if we should show both the HTML content along side markdown, or if just the HTML is enough.

chrisknoll:
I think we should go with this as-is:  adding the additional elements to put the markdown output in a separate element pushes the codebase in cohortDefinitionmanager.js to be even more unwieldy, so let's push the markdown output in atlas to a later enhancement when we have more UI elements encapsulated.

wivern (approved):

js/pages/cohort-definitions/cohort-definition-manager.js
```diff
@@ -628,6 +628,10 @@ define(['jquery', 'knockout', 'text!./cohort-definition-manager.html',
 			});
 
 			this.pollForInfoPeriodically();
+
+			this.subscriptions.push(
```
chrisknoll, line 632: Thanks for the update: looks like this will refresh print friendly if the underlying cohort definition changes (like if you change the cohort def in the route). Good catch.

I don't think I see this subscription being cleaned up?  Am I missing it?

js/services/CohortDefinition.js
```diff
@@ -151,17 +151,7 @@ define(function (require, exports) {
...
-			method: 'POST',
-			contentType: 'application/json',
-			data: JSON.stringify(cohortExpression),
-			error: function (error) {
-				console.log("Error: " + error);
-				authApi.handleAccessDenied(error);
-			}
-		});
-		return getCohortPrintFriendlyPromise;
+		return httpService.plainTextService.doPost(config.webAPIRoot + 'cohortdefinition/printfriendly/cohort?format=html', cohortExpression);
```
chrisknoll, line 154: Thanks for updating this service. I initially tried to use the httpService (as per the other places in the file) but I coudln't figure out how to allow it to handle plain-text responses. I didn't realizet here was a `plainTextService` api of httpService.

js/pages/cohort-definitions/cohort-definition-manager.js
wivern: Subscriptions are disposed inside the base `Component`. js/components/Component.js
```
dispose() {
        this.subscriptions.forEach(sub => sub.dispose());
    }
```

chrisknoll: Thank you thank you...I missed the prevous line of the push! haha, man, I'm getting old :)

PR #2364, opened 2020-11-02
Count criteria now supports different count columns.

### LINKED ISSUE #400: Add distinct options for cohort correlated criteria
### Current Behavior
When using correlated criteria (ie: criteria that is wrapped in 'with {at least} {1..n} using all occurrences of: {domain criteria}), the current choice is to count all events, or use distinct. The confusion is that it's not clear what is being used to make the events distinct.  In reality it is the {domain}_concept_id, but there are a few limitations:
- it's not clear, and therefore confusing
- the UI is a toggle, implying that there's only a Boolean choice between one option or another, and adding more options breaks this paradigm.
- The current language of 'using all' makes it sound like a distinct count isn't being made, but in fact it is a distinct count happening, just at the occurrence level (which is by definition unique, and therefore distinct)

Here is the a screenshot of the current behavior:
usingall

usingdistinct

### Proposed behavior
This proposal changes the behavior of the toggle-button into a dropdown with the following choices:
- distinct occurrence
- distinct concept
- distinct date

This allows you to count things that are occurrences (will count duplicate concepts, even on same day), by distinct concepts (but could happen on the same day), or by distinct dates (ensuring that the counting of things happens on different days).

The UI change would look like this:
dropdownlist

### Implementation considerations

This is not only a UI change: the WebAPI cohort builder will need to be updated to produce the correct queries based on this new input type.  Also, backwards compatibility must always be considered, so the implantation will need to handle prior cases where the 'using distinct' boolean value is in effect and translate it over to the 'distinct concepts' case in this proposal.

chrisknoll:
Fixes #400

### DIFF
--- a/js/components/cohortbuilder/InputTypes/Occurrence.js
+++ b/js/components/cohortbuilder/InputTypes/Occurrence.js
@@ -7,6 +7,7 @@ define(['knockout'], function (ko) {
 		self.IsDistinct = ko.observable(data.IsDistinct || false);
 		self.Type = ko.observable(data.Type===0 ? 0 : data.Type ||  2); // default to At Least
 		self.Count = ko.observable(data.Count===0 ? 0 : data.Count || 1); // default to 1 count
+		self.CountColumn = ko.observable(data.CountColumn);
 	}
 	
 	return Occurrence;
--- a/js/components/cohortbuilder/components/CriteriaGroup.js
+++ b/js/components/cohortbuilder/components/CriteriaGroup.js
@@ -237,6 +237,15 @@ define([
 					return true;
 			}
 		}
+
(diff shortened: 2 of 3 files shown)


wivern:
@chrisknoll I've checked this. Generally It works. 
Only one thing looks odd. When I change criteria from "using all" to "using distinct",
and then press the save button cohort definition saved properly without any errors 
but the save button doesn't go to be disabled.
This is not affect a newly created cohort definition, just when editing existing one.

chrisknoll:
Ok, that's strange behavior, possibly related to dirty flag?  Or something related to observables...I'll look into this, thanks for checking.

wivern (approved):

PR #2373, opened 2020-11-09
Added simplified cohort sample UI.

chrisknoll:
Depends on OHDSI/WebAPI#1657.

Thank you Andries Purnal (Andries.Purnal@ucb.com),  Florence Giesen (Florence.Giesen@ucb.com), @MaximMoinat and The Hyve for their implementation, which this PR is based on.

This is the simplified cohort sample functionality which allows a sample to be pulled from a source, and then navigated to the current patient profile UI.

### DIFF
--- a/js/pages/cohort-definitions/cohort-definition-manager.js
+++ b/js/pages/cohort-definitions/cohort-definition-manager.js
@@ -1775,7 +1775,7 @@ define(['jquery', 'knockout', 'text!./cohort-definition-manager.html',
 					alert('Error when deleting sample, please try again later');
 				})
 			} else if (e.target.className == 'sample-list fa fa-refresh') {
-				this.sampleDataLoading(false);
+				this.sampleDataLoading(true);
 				this.refreshSample({sampleId, sourceKey, cohortDefinitionId})
 				.then(res => {
 					this.showSampleDataTable(res.elements);
@@ -1807,12 +1807,7 @@ define(['jquery', 'knockout', 'text!./cohort-definition-manager.html',
 		}
 
 		refreshSample({sampleId, sourceKey, cohortDefinitionId}) {
-			sampleService.refreshSample({cohortDefinitionId, sourceKey, sampleId})
-			.then(res=>{
-				if(res.ok) {
-					console.log("todo, refresh result table with new sample");
-				}
-			})
+			return sampleService.refreshSample({cohortDefinitionId, sourceKey, sampleId})
 			.catch(() => {
 				alert('Error when refreshing sample, please try again later');
 			})
--- a/js/pages/cohort-definitions/cohort-definition-manager.js
+++ b/js/pages/cohort-definitions/cohort-definition-manager.js
@@ -225,22 +225,17 @@ define(['jquery', 'knockout', 'text!./cohort-definition-manager.html',
 			this.sampleName=ko.observable('');
 			this.patientCount=ko.observable();
 			this.sampleAgeType = ko.observable('');
-			this.isAgeRange =ko.observable(false);
 			this.firstAge = ko.observable();
 			this.secondAge = ko.observable();
 			this.isMaleSample=ko.observable(false);
 			this.isFeMaleSample=ko.observable(false);
 			this.isOtherGenderSample=ko.observable(false);
 			//error state
-			this.isAgeRangeError = ko.observable();
-			this.firstAgeError = ko.observable();
-			this.sampleNameError=ko.observable();
-			this.patientCountError=ko.observable();
-			//reset sample state after closing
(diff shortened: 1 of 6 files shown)


chrisknoll:
Back-end has been merged (so the endpoints are now available in WebAPI.  Need a review on the UI side.

blootsvoets:
The create sample form validation creates some errors now. Expected behaviour:

**Situation 1**: no values have been filled and no button has been pressed.

**Expected behaviour**: All fields are empty, no warnings are shown.
**Actual behaviour**: The _Name_, _Number of patients_, and _Age_ text inputs show error messages and they are indicated as being erroneous.
**Possible resolution**: don't do form validation until someone presses the submit button or possibly when they leave a form input field.

**Situation 2**. Sampling without age criteria

**Expected behaviour**: if no value is filled for age criteria, a sample is created with no age criteria applied.
**Actual behaviour**: form validation prohibits an empty age criteria, meaning that all samples now need age criteria.
**Possible solution**: allow an empty value in age criteria OR add a dropdown option "No age criteria" that removes the age field from view.

chrisknoll:
Hello,
 For the first item, I think showing validation errors as the form is populated is good UX:  you don't want to have to wait for the user to click a button in order to point out the things that are wrong.  In addition, the submit button is disabled until the form is valid, so I can't validate-on-submit.  

Good catch on the second one, I misunderstood the validation logic for a required field.  I can fix that.

chrisknoll:
Ok, I fixed the optional age range validation, and also noticed that the validation called for non-equal values, so I added that check too.

Thanks for checking!

js/pages/cohort-definitions/cohort-definition-manager.js
```diff
@@ -231,11 +231,11 @@ define(['jquery', 'knockout', 'text!./cohort-definition-manager.html',
...
 			this.isOtherGenderSample=ko.observable(false);
 			//error state
-			this.isAgeRange =ko.pureComputed(() => ['between','notBetween'].includes(this.sampleAgeType()));
-			this.isAgeRangeError = ko.pureComputed(() => this.isAgeRange() && (this.firstAge() == null || this.secondAge() == null));
-			this.firstAgeError = ko.pureComputed(() => this.firstAge() == null || this.firstAge() < 0);
 			this.sampleNameError=ko.pureComputed(() => this.sampleName().trim() == "");
 			this.patientCountError=ko.pureComputed(() => !(this.patientCount() > 0)); // this works because null == 0
+			this.isAgeRange =ko.pureComputed(() => ['between','notBetween'].includes(this.sampleAgeType()));
+			this.firstAgeError = ko.pureComputed(() => this.firstAge() != null && this.firstAge() < 0);
+			this.isAgeRangeError = ko.pureComputed(() => this.isAgeRange() && (this.firstAgeError() || (this.secondAge() != null && this.secondAge() < 0) || this.firstAge() == this.secondAge()));
```
blootsvoets, line 238: Not sure if this is encoded, it think that another age range condition is `firstAge <= secondAge`.

chrisknoll: Actually, the age order doesn't matter (I was surprised to find this) that's why the validation check says that it's an error if they are equal.  When the form is submitted, the smaller value is put as firstAge, and the larger is secondAge.

blootsvoets (dismissed):
LGTM

blootsvoets (commented):
I retested with the new configuration, there's still some issues there
1. If no age is provided, but the dropdown has been modified, the form sends the comparison mode selected without an age. The backend rejects this input. Either an input should be mandatory, with an additional dropdown option "No age criterion", or if no input is given in a single text input, also no age criterion is sent to the backend (`"age": null`).
2. If one age field is provided in `between` or `notBetween`, the form accepts the input, but the backend rejects it because it requires two ages. I think this input should be rejected in the form.
3. If no age was selected, the sample table shows "Random age" as age selection. Perhaps "Any age" would be more appropriate. Likewise, instead of "Mix", "Any gender" could be shown.

chrisknoll:
Thanks for the hints, I'll work on these changes.

chrisknoll:
Ok, I've made the changes you've suggested.  This includes moving to an 'async..await' form of handling promises so that try..catch works predictably.  The issue with prior code was how .catch() handlers were being attached to the promises, leading to then() being invoked even in case of error.  Now, the then() will only get fired when the request was successful, and the catch() and finally() will be handled within he context of the await.

chrisknoll:
I'm hoping that the final reviews are completed, and can get sign-off on this PR. Please let me know if you have any additional comments, as I'm anxious/excited to close this out.

blootsvoets:
Yes, looks good to me. Thanks for your hard work in making this suitable for merge!

anthonysena (approved):

PR #2386, opened 2020-11-24
Added ConditionStatus to UI.

### LINKED ISSUE #404: Support for condition_status_concept_id
### Expected behavior
Able to restrict to condition_occurrence that meet condition_status_concept_id criteria

**Examples of condition_status_concept_id's:**
- Admitting diagnosis: use concept_id 4203942
- Preliminary diagnosis: use concept_id 4033240
- Final diagnosis: use concept_id 4230359 – should also be used for ‘Discharge diagnosis’

### Actual behavior
No option to restrict by condition_status_concept_id

http://www.ohdsi.org/web/wiki/doku.php?id=documentation:cdm:condition_occurrence

chrisknoll:
Add the default domain to the domain options in ConceptPicker.js
Move default domain to top of domain options.
Fixes #404.

### DIFF
--- a/js/components/conceptpicker/ConceptPicker.js
+++ b/js/components/conceptpicker/ConceptPicker.js
@@ -27,7 +27,9 @@ define(['jquery','knockout', 'text!./ConceptPickerTemplate.html', './InputTypes/
 		self.dtApi = ko.observable();
 
 		VocabularyProvider.getDomains().then(function (domains) {
-			self.DomainOptions(domains);
+			const domainList = domains.filter(d => d != self.SelectedDomain); // moving selected domain to top
+			self.DomainOptions([self.SelectedDomain].concat(domains));
+
 			if (params.DefaultQuery != null) {
 				self.search().then(() => self.ProviderReady(true));
 			} else {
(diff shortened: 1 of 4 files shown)


js/components/conceptpicker/ConceptPicker.js
```diff
@@ -27,7 +27,13 @@ define(['jquery','knockout', 'text!./ConceptPickerTemplate.html', './InputTypes/
 		self.dtApi = ko.observable();
 
 		VocabularyProvider.getDomains().then(function (domains) {
-			self.DomainOptions(domains);
+			if (self.SelectedDomain != null) {
+				const domainList = domains.filter(d => d != self.SelectedDomain); // moving selected domain to top
+				self.DomainOptions([self.SelectedDomain].concat(domains));
```
anton-abushkevich, line 32: You mean `[self.SelectedDomain].concat(domainList)`? 

And in the first place, why do you want selected domain on top of the list? Popup already opens with selected domain.

anton-abushkevich (changes requested):

js/components/conceptpicker/ConceptPicker.js
chrisknoll: Yes, I shoudl ahve put domainList in the concat. Good catch, thank you!

I am doing this operation to handle the case where if the domain doesn't exist yet (as in the case of the domain 'Condition Status'):  if a default domain is specified, I still want it to appear in the list, even if it's not found from the vocabulary search.  (the specific case of the 'Condition Status' domain is more complicated than I want to go into here, but I'm trying to accommodate the random-feeling changes that are happening in the vocabulary).  Putting in the 'selected' domain on the top was just a nice side-effect.

I'll push up the fix to the right variable, but I hope the overall approach here makes sense.

anton-abushkevich (approved):

PR #2392, opened 2020-12-01
do not allow to create copy of FA preset

anton-abushkevich:
to prevent situations as in OHDSI/WebAPI#1706

### DIFF
--- a/js/pages/characterizations/components/feature-analyses/feature-analysis-view-edit.js
+++ b/js/pages/characterizations/components/feature-analyses/feature-analysis-view-edit.js
@@ -123,7 +123,7 @@ define([
             });
             this.canDelete = this.isDeletePermittedResolver();
             this.isNewEntity = this.isNewEntityResolver();
-            this.canCopy = ko.pureComputed(() => PermissionService.isPermittedCopyFa(this.featureId()));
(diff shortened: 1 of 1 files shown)


SergeySukhanov (approved):

js/pages/characterizations/components/feature-analyses/feature-analysis-view-edit.js
```diff
@@ -123,7 +123,7 @@ define([
             });
             this.canDelete = this.isDeletePermittedResolver();
             this.isNewEntity = this.isNewEntityResolver();
-            this.canCopy = ko.pureComputed(() => PermissionService.isPermittedCopyFa(this.featureId()));
+            this.canCopy = this.isCopyPermittedResolver();
```
chrisknoll, line 126: Just a question on implementation here: what is the purpose of the ...resolver() pattern here?  When you have the pureComputed in-line, the code is readable:  the logic of canCopy is immediately seen.  The ...Resolver() means you're redirect to some other place in code, for a function that isn't actually reused for anything...I would have thought that putting that logic into function makes sense when you have the same function used in multiple places, but in this case, is it a needless redirection?

PR #2413, opened 2020-12-16
issue-2410 - fix included concepts tab

### LINKED ISSUE #2410: Included concepts are not shown until Concept Set is saved and the page is refreshed
### Expected behavior
Included concepts are shown

### Actual behavior
Conepts are not shown
image
Same for Included Source Codes

1)Create a concept set
2)Save
3)Go To Search
4)Find some valid concepts
5)Select and add them
6)Back to concept set page

ekaterinakrivets:
fixes #2410

### DIFF
--- a/js/components/conceptset/ConceptSetStore.js
+++ b/js/components/conceptset/ConceptSetStore.js
@@ -80,10 +80,7 @@ define([
 			this.source = props.source || "unnamed";
 			this.title = props.title || "unnamed";
       
-      this.resolveCount = counter(); // handle out of order resolves
-
-			this.observer = ko.pureComputed(() => ko.toJSON(this.current() && this.current().expression.items()))
-				.extend({ rateLimit: { timeout: 1000, method: "notifyWhenChangesStop" } });
+      		this.resolveCount = counter(); // handle out of order resolves
 
 			this.isEditable = ko.observable(false);
 		}
--- a/js/components/conceptset/ConceptSetStore.js
+++ b/js/components/conceptset/ConceptSetStore.js
@@ -80,19 +80,17 @@ define([
 			this.source = props.source || "unnamed";
 			this.title = props.title || "unnamed";
       
-      		this.resolveCount = counter(); // handle out of order resolves
+			this.resolveCount = counter(); // handle out of order resolves
 
 			this.observer = ko.pureComputed(() => ko.toJSON(this.current() && this.current().expression.items()))
-				.extend({ rateLimit: { timeout: 1000, method: "notifyWhenChangesStop" } });
-			this.selectedTab  = ko.observable(ViewMode.EXPRESSION);
+				.extend({ rateLimit: { timeout: 500, method: "notifyWhenChangesStop" } });
 
 			// watch for any change to expression items (observer has a delay)
 			this.observer.subscribe(async () => {
 			    try {
 				    await this.resolveConceptSetExpression();
-				    await this.refresh(this.selectedTab());
 			  	} catch (err) {
-				  	if (err != RESOLVE_OUT_OF_ORDER)
+				  	if (err !== RESOLVE_OUT_OF_ORDER)
 					  	console.info(err);
 				  	else
 					  	throw(err);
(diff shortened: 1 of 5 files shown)


js/pages/concept-sets/conceptset-manager.js
```diff
@@ -266,7 +266,8 @@ define([
 				} finally {
 				}
 			}));  
-			
+
+			this.conceptSetStore.resolveConceptSetExpression();
```
wivern, line 270: resolveConceptSetExpression is highly expensive operation. It could be dozens of concepts inside a conceptSet.
The idea was to load included concepts only when `Included` tab is explicitly selected or conceptSet was modified.

ekaterinakrivets: moved this call to included concepts tab

SergeySukhanov: so better

SergeySukhanov (approved):

wivern (changes requested):
source codes doesn't work properly

js/components/conceptset/ConceptSetStore.js
```diff
@@ -18,7 +18,7 @@ define([
 	JSZip,
 ) {
 
-	const {ViewMode} = constants;
+	const { ViewMode,  RESOLVE_OUT_OF_ORDER } = constants;
```
ekaterinakrivets, line 21: fixed typo

js/components/conceptset/ConceptSetStore.js
```diff
@@ -82,6 +82,24 @@ define([
       
       		this.resolveCount = counter(); // handle out of order resolves
 
+			this.observer = ko.pureComputed(() => ko.toJSON(this.current() && this.current().expression.items()))
+				.extend({ rateLimit: { timeout: 1000, method: "notifyWhenChangesStop" } });
+			this.selectedTab  = ko.observable(ViewMode.EXPRESSION);
```
ekaterinakrivets, line 87: added this observable to support previous logic

js/pages/concept-sets/components/tabs/conceptset-expression.js
```diff
@@ -121,7 +121,7 @@ define([
       this.conceptSetItems().forEach(conceptSetItem => {
         conceptSetItem[key](!areAllSelected);
       })
-      await this.conceptSetStore.resolveConceptSetExpression();
```
ekaterinakrivets, line 124: this one became redundant after fix

wivern (approved):

js/pages/concept-sets/components/tabs/conceptset-expression.js
chrisknoll: The reason why we don't want to put the subscription to the observer inside the concept set store is that we don't really need to constantly resolve the concept set expression unless someone is going to be 'looking' at it.  That's why we attach the subscription in the UI components, but also have to do a 'resolve' before we hookup the subscription because we don't know if it is up to date.

You removed the subscription in the 'pages/concept-set' but not from the embedded UI (like cohort definitions, cohort characterization, ethich both use the 'shared' concept set editor UI...the repository concept set works slightly differently).   My suggestion is:  take the subscription out of concept set store, and put the resolveConceptSet expression and subscription to the conceptSetStore.observer back.

js/pages/concept-sets/conceptset-manager.js
chrisknoll: I think this has lead to the issue reported in https://github.com/OHDSI/Atlas/issues/2421.  Previously, there was an 'await resolveConceptSetExpression()' which did hold the UI up until the initial conceptIDs for the concept set were loaded, and it's probably the case that this did not need to be an await (ie: load asynchronously) but I also thought there was a reason I did this (I can't remember now).

chrisknoll: Also, just to note: resolveConceptSetExpression just load the set of included conceptIDs, not the full-blown concept objects (with names/domains, etc) which is a more expensive operation.  You are correct that resolveConceptSetExpression is an expensive operation, but it's only half the work to actually load the 'includedConcepts' (which coverts conceptIDs -> Concept objects)

chrisknoll:
Hey, all,
  After reviewing this to see if there was a way to limit the calls to `resolveConceptSetExpression()` until necessary, but I was finding that it was more complicated to work out the logic of calling `resolve()` at the "right time" so that the individual tab results are updated, and I think what you've done here is simplest:  always resolve the concept expression as it changes, but then only reload the included concepts / mapped concepts if it is being viewed.  So, I think we should leave what you guys did in place, it makes sense and looks good,

PR #2417, opened 2020-12-18
issue-2410 - fix included concepts tab (#2413)

ssuvorov-fls:
* issue-2410 - fix included concepts tab

* added condition to load included concept sets only when tab is opened

* removed logs

* issue-2410 - moved observer to conceptset-manager.js

* issue-2410 - fixed observer for conceptsets, removed redundant func call

* separate view logic from store

Co-authored-by: wivern <vitaly.koulakov@gmail.com>

(cherry picked from commit 0c7fb56)

### DIFF
--- a/js/components/conceptset/ConceptSetStore.js
+++ b/js/components/conceptset/ConceptSetStore.js
@@ -18,7 +18,7 @@ define([
 	JSZip,
 ) {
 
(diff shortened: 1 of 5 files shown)


konstjar (approved):
Lgfm

PR #2444, opened 2021-01-20
Configure access modal fails on cohort definition load #2441

### LINKED ISSUE #2441: Configure access modal fails on cohort definition load
### Expected behavior

When loading a cohort definition on https://atlas.ohdsi.org, the cohort definition loads and a read-only version appears.

### Actual behavior

An error is thrown which stops the page from properly rendering:

image

Navigate to https://atlas.ohdsi.org/#/cohortdefinition/317 with the developer tools enabled to view the error.

anton-abushkevich:
fixes #2441

@anthonysena 
Concerning expected behavior. Read-only version available for user with "Atlas users" role. If user have only "public" role, then after the fix Atlas will show him this screen: 
изображение

### DIFF
--- a/js/pages/cohort-definitions/cohort-definition-manager.html
+++ b/js/pages/cohort-definitions/cohort-definition-manager.html
@@ -1,12 +1,13 @@
 <loading data-bind="visible: $component.isLoading()"></loading>
+<access-denied params="isAuthenticated: isAuthenticated, isPermitted: hasAccess"></access-denied>
+
+<!-- ko if: hasAccess -->
 <div data-bind="if: currentCohortDefinition() && $root.router.currentView() == 'cohort-definition-manager' && !$component.isLoading()" class="flexed">
 	<heading-title params="name: cohortDefinitionCaption(), description: canEdit() ? '' : '(Read only)', icon: 'users', theme: 'dark'"></heading-title>
 	<!-- ko if: currentCohortDefinition() && currentCohortDefinition().id() != 0 -->
 	<!-- ko component: {name: 'authorship', params: getAuthorship()} --> <!-- /ko -->
 	<!-- /ko -->
-	<access-denied params="isAuthenticated: isAuthenticated, isPermitted: hasAccess"></access-denied>
 
-	<!-- ko if: hasAccess -->
 	<div data-bind="currentCohortDefinition() != null">
 		<div class="asset-heading">
 			<div class="input-group">
@@ -557,7 +558,7 @@
 		<b>Full Analysis</b>
 		<p>Runs a complete set of characterization reports. (estimated: 60 minutes)</p>
 	</atlas-modal>
-	<!-- /ko -->
+
 	<atlas-modal params="{
 		showModal: cohortLinkModalOpened,
 		iconClass: 'fa fa-group',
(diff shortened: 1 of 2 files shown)


anthonysena:
Thanks @anton-abushkevich - I'm working with Lee to verify that this resolves the issue on atlas.ohdsi.org before I approve/merge.

anthonysena (changes requested):
@anton-abushkevich - after review, these changes to not achieve the intended behavior. I traced the code and found that the root cause of this problem is this code:

https://github.com/OHDSI/Atlas/blob/317d1975e43fb2f017e224cea173e67a6194cf71/js/pages/cohort-definitions/cohort-definition-manager.js#L413

This line should be removed since the `PermissionService` call later on will then enforce the proper access controls:

https://github.com/OHDSI/Atlas/blob/317d1975e43fb2f017e224cea173e67a6194cf71/js/pages/cohort-definitions/cohort-definition-manager.js#L812-L817

I've tested this change on my side and it works well. Can you revise this PR? From there we should be set to merge.

anton-abushkevich:
@anthonysena 
I agree that line 443 prevents `PermissionService.decorateComponent()` and should be removed. I updated the branch. 

But if we only remove this line, users with **default** public role will see just a blank page when trying to assess corort via direct link.

изображение

Meanwhile, in console:
изображение
which is ok, since this user does not have access to cohort. And `currentCohortDefinition()` remains undefined in cohort-definition-manager.js. So I suggest to keep my changes as well: 
`<access-denied params="isAuthenticated: isAuthenticated, isPermitted: hasAccess"></access-denied>` should be above `<div data-bind="if: currentCohortDefinition()...` in DOM in order to show "access denied" text (see screenshot in the description).

Btw, i cannot reproduce the issue in https://atlas.ohdsi.org. Did you change public role's rights there?

anthonysena:
Thanks @anton-abushkevich for this analysis - it is very helpful. To your question: yes, we've modified https://atlas.ohdsi.org so that the public role contains a subset of permissions to provide read-only access for anyone who logs into that environment. The default 'public' role that comes out of the box with WebAPI was a bit too restrictive for what we needed in atlas.ohdsi.org.

One concern I had with the changes here: by moving the `access-denied` component above in the DOM: if a user has read-only access to a cohort, they will be unable to navigate away from that cohort definition. I'm not sure of the best way to test this out with the default roles? Perhaps the cohort reader role is useful for this type of test?

anton-abushkevich:
@anthonysena - thank you for your reply.
If a user has read-only access to cohort, they will see the cohort in read-only mode and will be able to navigate from it. Contents of `<access-denied params="isAuthenticated: isAuthenticated, isPermitted: hasAccess"></access-denied>` doesn't appear in this case, since user `isAuthenticated` and `hasAccess` to cohort, though read-only. I tested it locally for default user role "cohort reader" and separately for user with role "Atlas users". The result is the same:
изображение

And if a user has only default public role, then Atlas shows them screen you see in the description. Only way to navigate from there is the left menu. If necessary, I can add page title ("Cohort Definitions" or "Cohort #X") above the access-denied text.

anton-abushkevich:
I added another commit, preventing error reading this.currentCohortDefinition() when it's undefined.

anthonysena:
Thanks @anton-abushkevich for all of the information here! These changes seem to make sense upon review - I'd like to deploy them to verify that the behavior is as we intend. Thank you!

anthonysena (approved):

PR #2461, opened 2021-01-27
Loader gets stuck when we adding Included Concepts in Concept Sets #2446

### LINKED ISSUE #2446: Loader gets stuck when we adding Included Concepts in Concept Sets
Atlas 2.8

### Actual behavior

Loader gets stuck. Eventually, concept was added but if we don't switch the tab loader is still getting stuck. No errors were found in the console.
loader Included Concepts

1. Create a new Concept Set
2. Add some concept (example "aspirin") -> Save
3. Add descendants to make Included Concepts available
4. Included concepts -> Choose some concept
5. Add concepts to Consept Set

anton-abushkevich:
fixes #2446

### DIFF
--- a/js/components/conceptset/included.js
+++ b/js/components/conceptset/included.js
@@ -55,8 +55,8 @@ define([
 				conceptSetStore: this.conceptSetStore,
 			});
 			commonUtils.clearConceptsSelectionState(this.includedConcepts);
-    }
-	
+			this.conceptSetStore.loadingIncluded(false);
+		}
 	}
 
 	return commonUtils.build('conceptset-list-included', IncludedConcepts, view);
--- a/js/components/conceptset/included.js
+++ b/js/components/conceptset/included.js
@@ -47,7 +47,6 @@ define([
 		}
 
 		addConcepts(options) {
-			this.conceptSetStore.loadingIncluded(true);
 			const concepts = commonUtils.getSelectedConcepts(this.includedConcepts);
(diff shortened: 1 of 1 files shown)


js/components/conceptset/included.js
```diff
@@ -55,8 +55,8 @@ define([
 				conceptSetStore: this.conceptSetStore,
 			});
 			commonUtils.clearConceptsSelectionState(this.includedConcepts);
-    }
-	
+			this.conceptSetStore.loadingIncluded(false);
```
chrisknoll, line 58: Little confused here: adding the concepts tot he concept set store will signal a reload which should handle clearing the `loadingIncluded` flag (once all the async requests are complete).  But this looks like it will set the `loadingIncluded` to true and then false almost immediately due to the async calls?

anton-abushkevich: You are right. We do not need to place loader here at all, Because if/when we apply changes from https://github.com/OHDSI/Atlas/pull/2460/files - there wiil be no async calls here to wait for. So i suggest to just remove this.conceptSetStore.loadingIncluded(true); from line 50.

anton-abushkevich: In any case, if we need to put a loader on the page, we rather not do it in this function, but where async calls are. Which is already there: js/components/conceptset/ConceptSetStore.js:158

chrisknoll:
Fixed in #2460.

PR #2487, opened 2021-03-09
Update StdDiff logic.

chrisknoll:
Remove std mean diff calculations from javascript (they were unused).
Handle Infinity/-Infinity response from server.

Depends on OHDSI/WebAPI#1812.


chrisknoll:
Although this is for the changes in WebAPI (to handle Inf response values), this code will work if the server isn't returning those responses.

anthonysena (approved):

Issue #2497, opened 2021-03-25
User display name not shown, login is


alondhe:
### Expected behavior
When using authentication, the display name of a user should be shown, rather than their login.

### Actual behavior
We currently set the login name as the name to show.

### Steps to reproduce behavior
1. Using authentication, log in. 
2. Names are all logins

anton-abushkevich:
Fixed in 2.11 (see #2670).

PR #2498, opened 2021-03-25
Changed createdBy to use name instead of login

alondhe:
#2497 - Atlas is showing login values instead of display names

### DIFF
--- a/js/utils/DatatableUtils.js
+++ b/js/utils/DatatableUtils.js
@@ -44,8 +44,8 @@ define(['services/MomentAPI', 'xss', 'appConfig', 'services/AuthAPI', '../const'
             d.hasOwnProperty('createdBy') && !!d.createdBy
                 ? typeof d.createdBy === 'string'
                     ? d.createdBy
-                    : typeof d.createdBy === 'object' && !!d.createdBy.login
-                    ? d.createdBy.login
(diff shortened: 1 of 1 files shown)


js/utils/DatatableUtils.js
```diff
@@ -54,7 +54,7 @@ define(['services/MomentAPI', 'xss', 'appConfig', 'services/AuthAPI', '../const'
         const getFacetForCreatedBy = getCreatedByLogin;
 
         const getFacetForDesign = d =>
-            d.hasWriteAccess || (d.createdBy && authApi.subject() === d.createdBy.login)
+            d.hasWriteAccess || (d.createdBy && authApi.subject() === d.createdBy.name)
```
ssuvorov-fls, line 57: In case of different login and name we will always get "Other designs" because authApi.subject() contains user login Subject is set in AuthAPI.loadUserInfo

chrisknoll:
We'll target this PR for 2.10 (since 2.9 is closing in now) but it would be good if we could get resolution on the 2 items in this pr:
- Resolve conflicts
- Answer the question about the following from @ssuvorov-fls : 

> In case of different login and name we will always get "Other designs" because authApi.subject() contains user login. Subject is set in AuthAPI.loadUserInfo

anton-abushkevich:
Fixed in 2.11 (see #2670).

PR #2519, opened 2021-04-21
Included concepts table is not shown #2518

### LINKED ISSUE #2518: Included concepts table is not shown
Atlas 2.9
### Expected behavior
Included concepts table is shown

### Actual behavior
image

1)Open any concept set
2)Included concepts

anton-abushkevich:
Fixes #2518

### DIFF
--- a/js/components/conceptset/utils.js
+++ b/js/components/conceptset/utils.js
@@ -67,6 +67,7 @@ define(['knockout','utils/CommonUtils', 'utils/Renderers', 'services/http','atla
 	];
 
 	const includedConceptsOptions = {
+		xssSafe: true,
 		autoWidth:false,
 		Facets: [
 			{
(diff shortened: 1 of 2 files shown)


chrisknoll:
Sorry, I am confused about this change: why does xss safe flag lead to a hidden table?  The xsssafe field is used to improve performance: none of the values in the included concepts tab is user-input, so there's no XSS vulnerability there, but there are thousands of rows and even more columns where xss safety will attempt to 'clean' all the cells of malformed html.  That's a lot of cost, and doesn't quite explain why the table is hidden (where there are other tables where this flag is set and the table renders properly).

anton-abushkevich:
Yes, there is another way: if xssSafe enabled then only unwrap the title and don't do any other 'cleaning'. I returned xssSafe to 'Included Concepts' table and made necessary changes in datatableBinding.js.

chrisknoll (approved):

anton-abushkevich:
@chrisknoll, why you reverted the commit? Please comment if you found any problems with changed code.

chrisknoll:
I'm sorry, I clicked accidently the revert, but did not actually 'commit' the revert i canceled that!  Sorry, I'll revert the revert.

chrisknoll:
I've deleted the branch, but the 'revert' message in this commit log remains. Sorry for the confusion, but this PR was not reverted.

PR #2575, opened 2021-07-01
Improve docker config

blootsvoets:
Changes:
- Added an optional config file path in the Docker configuration to put alternative config-local.js
- Bumped Dockerfile dependencies


blootsvoets:
@chgl this might also be a usability improvement for the edit you did earlier.

anthonysena (approved):

PR #2589, opened 2021-08-06
Tagging capability

anton-abushkevich:
Description: OHDSI/WebAPI#1917

Backend PR: OHDSI/WebAPI#1914


chrisknoll (approved):
Functionality seems to work, and I checked that the new subscriptions were being disposed (they are).  What I didn't check is that bundling works, so if you could confirm that there's no issues with bundling, then we should e good to go for the UI portion of this functiion.

Issue #2491, opened 2021-03-11
Atlas Cohort Definition Validation Function


jduke99:
This function intended to support evaluation of cohort definition performance via manual review of patient data in in the Profile Page in Atlas. The manual case validation process involves having a domain expert review the clinical data for a patient and assert if the patient is a true case or a false positive.  In some cases, additional information may need to be annotated beyond the case/no-case assessment.

This function will enable:
1) Creation of a "Question Set" for a cohort definition

![image](https://user-images.githubusercontent.com/5069947/110850487-d6ba7980-827d-11eb-873a-e81bf96a879f.png)

2) Pairing a cohort sample (from the Sample tab) with a question set to create a Validation study which will enable annotating patients in that sample using the questions in the question set.

3) Answering of study questions while on the patient profile page

![image](https://user-images.githubusercontent.com/5069947/110850743-26994080-827e-11eb-94a8-b2663b938c1b.png)

4) Reviewing or downloading the case/no-case metrics for the cohort sample.

PR #2593, opened 2021-08-22
Atlas Cohort Definition Validation Feature

cahilton:
This pull request creates the Atlas features for cohort definition validation. There are primarily three functional areas impacted by this new feature. 

1. **Cohort Definition**

- On the Samples tab, provides an option to create a validation study, which essentially links the sample to the validation question set.
- Creates the new 'Validation' tab, which allows for the creation of question sets, and the linking of question sets to sample in order to create studies.

2. **Profile**
- Adds a UI feature to annotate validation studies, which appears on the right of the page
- Adds optional routing parameters that link explicitly to samples and question sets (the validation study).

#2491 describes this feature in greater detail

@jduke99 can provide additional details.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,9 @@
 /js/assets/fonts
 /js/assets/images
 js/config-local.js
+js/config-gis.js
 js/version.js
 package-lock.json
 .vscode
 .DS_Store
+Broadsea-WebTools/
--- a/js/components/analysisExecution/analysis-execution-list.js
+++ b/js/components/analysisExecution/analysis-execution-list.js
@@ -47,7 +47,11 @@ define([
     }) {
       super();
       this.analysisId = analysisId;
-      this.analysisId.subscribe(() => this.loadData());
+      this.analysisId.subscribe((id) => {
+        if (id) {
+          this.loadData();
+        }
+      });
       this.resultsPathPrefix = resultsPathPrefix;
       this.downloadApiPaths = downloadApiPaths;
       this.downloadFileName = downloadFileName;
--- a/js/pages/characterizations/routes.js
+++ b/js/pages/characterizations/routes.js
@@ -34,7 +34,7 @@ define(
 						router.setCurrentView('characterizations-list');
 					});
 				}),
-				'c/:id:': characterizationViewEdit,
+				'cc/characterizations/:id:': characterizationViewEdit,
 				'cc/characterizations/:id:/:section:': characterizationViewEdit,
 				'cc/characterizations/:id:/:section:/:subId:': characterizationViewEdit, // for executions
 				'cc/feature-analyses': new AuthorizedRoute(() => {
--- a/js/pages/characterizations/routes.js
+++ b/js/pages/characterizations/routes.js
@@ -34,7 +34,7 @@ define(
 						router.setCurrentView('characterizations-list');
 					});
 				}),
-				'cc/characterizations/:id:': characterizationViewEdit,
+				'c/:id:': characterizationViewEdit,
 				'cc/characterizations/:id:/:section:': characterizationViewEdit,
 				'cc/characterizations/:id:/:section:/:subId:': characterizationViewEdit, // for executions
 				'cc/feature-analyses': new AuthorizedRoute(() => {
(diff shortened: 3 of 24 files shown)


cahilton:
I have some additional tweaks to make on my end after the WebAPI PR was updated. Thanks!

chrisknoll:
I'm getting an error when trying to create/modify a question set.:
image

In this image, I'm seeing a console error about how 'type' is not a function. when looking at how a `Question` is created, I do see that the type property of `Question` is set to `ko.observable('')`.  So, i believe that would be the third element in the `questions()` collection.  Somewhere, the `type` property of the object is losing the observable status and becomming the raw type...possibly after the question set is saved...I'm not sure yet, but there is an error here.

chrisknoll:
Looking further into the implementation, I think the problem stems from QuestionSetForm.js, around lines 128-151:

```
            for (i = 0; i < self.questions().length; i++) {
                if (self.questions()[i].type() === 'Text') {
                    self.questions()[i].answers.push(new Answer());
                    self.questions()[i].answers()[0].text = '';
                    self.questions()[i].type = 'TEXTAREA';
                } else if (self.questions()[i].type() === 'Numeric') {
                    self.questions()[i].answers.push(new Answer());
                    self.questions()[i].answers()[0].text = '';
                    self.questions()[i].type = 'NUMERIC';
                } else if (self.questions()[i].type() === 'Date') {
                    self.questions()[i].answers.push(new Answer());
                    self.questions()[i].answers()[0].text = '';
                    self.questions()[i].type = 'DATE';
                } else if (self.questions()[i].type() === "Checkbox") {
                    self.questions()[i].type = 'MULTI_SELECT';
                    for (j = 0; j < self.questions()[i].answers().length; j++) {
                        currentAnswer = self.questions()[i].answers()[j];
                        currentAnswer.value = currentAnswer.text();
                    }
                } else if (self.questions()[i].type() === "Radio Button") {
                    self.questions()[i].type = 'SINGLE_SELECT';
                    for (j = 0; j < self.questions()[i].answers().length; j++) {
                        currentAnswer = self.questions()[i].answers()[j];
                        currentAnswer.value = j;
                    }
                }
            }
```

If you notice, each of the `else` clauses assigns the `type` property back to a native type, and not an observable value:

```
                    self.questions()[i].type = 'TEXTAREA';
...
                    self.questions()[i].type = 'NUMERIC';
,,,
                    self.questions()[i].type = 'DATE';

```

Once that happens, you can't call `questions()[i].type()` anymore (since it's not an observable).  So I think that is where the data corruption is coming from (corruption = observable converted to non-observable).

Perhaps part of an issue at play here is a misunderstanding between a 'viewmodel' and the raw data, but we have a few places in code that could serve as an example about how you can define a JavaScript object to represent the object as a viewmodel.  It looks like Question and Answer have an implementation:

```
        function Question() {
            const self = this;
            self.text = ko.observable('');
            // self.helpText = ko.observable('');
            self.type = ko.observable('');
            self.caseQuestion = ko.observable('');
            self.required = ko.observable('');
            self.answers = ko.observableArray([]);

            self.addAnswer = function () {
                if ((self.type() !== 'Text' && self.type() !== 'Numeric' && self.type() !== 'Date') && self.type() !== undefined) {
                    self.answers.push(new Answer());
                }
            };
            self.removeAnswer = function (item) {
                self.answers.remove(item);
            };

            self.questionTypeChanged = function (obj, evt) {
                if (obj !== 'Text' && obj !== 'Numeric' && obj !== 'Date') {
                    if (self.answers().length === 0) {
                        self.addAnswer();
                    }
                }
            };
        }
```

However, I would have put this into a separate file for annotation 'types', and passed a constructor object into the function (ie: `new Answer(data) {...}`) such that the answer object can be initialized based on some other JSON/JSObject source (such as something that comes back from a web service endpiont).

Examples of where we create viewmodels to represent data objects:
https://github.com/OHDSI/Atlas/blob/master/js/components/cohortbuilder/InputTypes/Occurrence.js
https://github.com/OHDSI/Atlas/blob/master/js/components/cohortbuilder/InputTypes/Text.js

Etc. I use these as examples because these elements are the ones used to capture user input for cohort definitions, and so it might be a good guide on creating objects meant to capture user input about question set definitions.

PR #2597, opened 2021-08-23
 Add Visit Criteria - shows place of service attribute #2594

### LINKED ISSUE #2594: Add Visit Criteria - shows place of service attribute
### Expected behavior
When pulling in a visit criteria, no attributes should be populated by default.

### Actual behavior

This odd "with distance from a Place of Service to Patient" attribute appears by default:

image

Add a visit criteria to a new cohort in Atlas v2.9x. Found in our version and in the Atlas demo site: http://atlas-demo.ohdsi.org/#/cohortdefinition/1777598/definition

anton-abushkevich:
Fixes #2594. 

PlaceOfServiceDistance() is undefined, not null. So strict comparison `PlaceOfServiceDistance() !== null` returned `true`.

### DIFF
--- a/js/components/cohortbuilder/components/VisitOccurrenceTemplate.html
+++ b/js/components/cohortbuilder/components/VisitOccurrenceTemplate.html
@@ -135,7 +135,7 @@
 					</div>
 				</td>
 			</tr>
-			<tr data-bind="if: PlaceOfServiceLocation() !== null, visible: PlaceOfServiceLocation() !== null">
+			<tr data-bind="if: PlaceOfServiceLocation() != null, visible: PlaceOfServiceLocation() != null">
 				<td><i data-bind="click: function() { $component.removeCriterion('PlaceOfServiceLocation') }, title: ko.i18n('common.remove', 'Remove')" class="fa fa-times" /></td>
 				<td>
 					<div>
@@ -147,7 +147,7 @@
 					</div>
 				</td>
 			</tr>
-			<tr data-bind="if: PlaceOfServiceDistance() !== null, visible: PlaceOfServiceDistance() !== null">
+			<tr data-bind="if: PlaceOfServiceDistance() != null, visible: PlaceOfServiceDistance() != null">
 				<td><i data-bind="click: function() { $component.removeCriterion('PlaceOfServiceDistance') }, title: ko.i18n('common.remove', 'Remove')" class="fa fa-times" /></td>
 				<td>
 					<div>


chrisknoll:
But what is `PlaceOfServiceDistance` ?  it doesn't appear anywhere in the Circe object model.

anton-abushkevich:
Maybe it's obsolete?

PlaceOfServiceDistance was added in this commit (18 apr 2019): https://github.com/OHDSI/Atlas/commit/593cf937fc05700523aeedba0d5f533e2930223e
And disabled by default two days later: https://github.com/OHDSI/Atlas/commit/cc3e450981807cd5b1ce696126982b918062d604

(timeline: https://github.com/OHDSI/Atlas/commits/master?after=fc9edd9b101dc5d778be851aa42eb6447df7d29b+314&branch=master)

chrisknoll:
Ok, so if it can be hidden behind a config flag, Good to know.  I wish I understood what it was doing, I wonder if this has something to do with the 'overridable' query builders that @pavgra implemented.  I don't know if anyone is using that.

Ok so I feel like your fix for the `!==` is correct since that's how the other criteria checks are handled, so I can merge this in, but the demo site should probably be configured to disable this (I'm not sure why the atlas-demo site is showing this field when the default is false...unless @leeevans enabled it?)  

In any case, could you confirm that when that coniguraton is set to false by default, the element does not show up in the UI?  That might be another bug that could be fixed in this PR.

alondhe:
@anton-abushkevich  -- I see locationDistance: false in our app.js file. Do we need to explicitly call this out in the config-local.js file?

anton-abushkevich:
@alondhe No, you do not need to explicitly set locationDistance to false in config-local.js.

anton-abushkevich:
@chrisknoll Yes, if locationDistance is set to false by default, the element does not show up in the UI.

chrisknoll:
Ok, so, let merget this PR in because it does fix that problem, and you confirm the flag is working correctly in the config, so there's no worries there.

The real question is why is that flag enabled in atlas-demo.ohdsi.org, but that's a separate thread I'll follow up on.

chrisknoll (approved):

chrisknoll:
@alondhe : I should have probably asked this of you before I merged:  are you seeing the UI element in your instance even tho your app.js shows false?  I'm a little confused about which enviornment (atlas-demo or something you are hosting) shows the UI element when the config says `false`. Can you just confirm that you're seeing the element even tho the config is false?

alondhe:
@chrisknoll -- yep, I'm seeing the element in our instance even though app.js shows false. This is a deployment without the changes from this PR.

chrisknoll:
Ok (dammit!) @anton-abushkevich , it sounds like there's some disagreement in your statement that the UI doesn't show when the value is `false` but @alondhe is seeing it and his value is `false`.   Could be a problem and if so, Ajit, can you make a new issue about that configuration variable not working (ie: setting it false leads to the field being visible in the UI).?

Also, Ajit, you may want to check to see if the docker config somehow has a config-local.js setting that overrides the default.  That would explain how the 'default' value of false is being overridden by some docker configuration setting.

anton-abushkevich:
@chrisknoll Sorry for unclear explanation. 

The value in the configuration is for dropdown menu, and its working with or without the fix. See https://github.com/OHDSI/Atlas/blob/master/js/components/cohortbuilder/components/VisitOccurrence.js#L130
```
    if (config.features.locationDistance) {
      self.addActions....
    }
```

With `locationDistance: true`:
изображение

With `locationDistance: false`:
изображение

This fix is for criteria, that shows up (ignoring configuration) when you choose `Add visit...`. See https://github.com/OHDSI/Atlas/blob/master/js/components/cohortbuilder/components/VisitOccurrenceTemplate.html#L150

Was:
`<tr data-bind="if: PlaceOfServiceDistance() !== null, visible: PlaceOfServiceDistance() !== null">`
`PlaceOfServiceDistance()` is `undefined`, not `null`, so strict comparison makes the criteria visible.
изображение

Now:
`<tr data-bind="if: PlaceOfServiceDistance() != null, visible: PlaceOfServiceDistance() != null">`
The `if` statement works as intended now.
изображение

Ajit just needs to have this fix on their environment. I will merge the changes there today.

PR #2610, opened 2021-09-06
FHIR terminology import

johngrimes:
This change adds a tab to the Import section of the Concept Sets area.

A user can enter:

1. The URL of a [FHIR terminology server](https://hl7.org/fhir/R4/terminology-service.html), and;
2. A URI describing a FHIR [ValueSet](https://www.hl7.org/fhir/R4/valueset.html) known to that terminology server.

A request will be made to the ValueSet [expand](https://hl7.org/fhir/R4/valueset-operation-expand.html) operation on that server, and the resulting codes will be imported.

#### Limitations

- Matching is done on code only, in the same way the "Source Codes" import currently works. This could be enhanced when a proper mapping of FHIR code system identifiers to OMOP vocabulary identifiers is available, which is [currently being worked on](https://chat.fhir.org/#narrow/stream/179202-terminology/topic/OMOP.20Standard.20Terminologies).

### DIFF
--- a/js/services/http.js
+++ b/js/services/http.js
@@ -3,7 +3,7 @@ define(function(require, exports) {
   const config = require('appConfig');
   const sharedState = require('atlas-state');
   const { Api:OHDSIApi, STATUS } = require('ohdsi-api');
-  const JSON_RESPONSE_TYPE = 'application/json';
+  const FHIR_JSON_RESPONSE_TYPE = 'application/fhir+json';
   const TEXT_RESPONSE_TYPE = 'text/plain';
   const EventBus = require('services/EventBus');
 
@@ -100,6 +100,18 @@ define(function(require, exports) {
     }
   }
 
+  class FhirApi extends Api {
+    getHeaders(requestUrl) {
+      return {
+        'Accept': FHIR_JSON_RESPONSE_TYPE,
+      };
+    }
+
+    parseResponse(json) {
+      return JSON.parse(json);
+    }
+  }
+
   const singletonApi = new Api();
   singletonApi.setAuthTokenHeader('Authorization');
   
@@ -108,7 +120,10 @@ define(function(require, exports) {
   plainTextService.setUnauthorizedHandler(() => singletonApi.handleUnauthorized());
   plainTextService.setUserTokenGetter(() => singletonApi.getUserToken());
 
+  const fhirService = new FhirApi();
+
   singletonApi.plainTextService = plainTextService;
+  singletonApi.fhirService = fhirService;
 
   return singletonApi;
 });
--- a/js/components/conceptset/import.js
+++ b/js/components/conceptset/import.js
@@ -12,6 +12,7 @@ define([
 	'./import/identifiers',
 	'./import/sourcecodes',
 	'./import/conceptset',
+	'./import/fhir',
 ],function(
 	ko,
 	view,
@@ -60,6 +61,12 @@ define([
 				{
 					title: ko.i18n('components.conceptSet.repository', 'Repository'),
 					key: 'repository',
+				},
+				{
+					title: ko.i18n('components.conceptSet.fhir', 'FHIR'),
+					key: 'fhir',
+					componentName: 'conceptset-list-import-fhir',
+					componentParams: {...params, appendConcepts: this.appendConcepts}
 				}
 			];
 		}
--- a/js/components/conceptset/import/fhir.html
+++ b/js/components/conceptset/import/fhir.html
@@ -1,0 +1,18 @@
+<div>
+    <div class="heading">
+        <span data-bind="text: ko.i18n('components.conceptSet.fhirHeading', 'Import a value set from a FHIR terminology server')"></span>
+    </div>
+    <label>
+        <span data-bind="text: ko.i18n('components.conceptSet.fhirServer', 'FHIR server URL')"></span>
+        <input class="form-control" type="text" name="fhirServer" data-bind="textInput: fhirServer"/>
+    </label>
+    <label>
+        <span data-bind="text: ko.i18n('components.conceptSet.valueSet', 'Value set URI')"></span>
+        <input class="form-control" type="text" name="valueSet" data-bind="textInput: valueSet"/>
+    </label>
+    <concept-add-box params="{
+        isActive: canAddConcepts,
+        onSubmit: doImport,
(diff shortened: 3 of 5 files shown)


johngrimes:
Just following up on this, has anybody had a chance to take a look?

ablack3:
@anthonysena, @alex-odysseus, @chrisknoll - Who is responsible for reviewing this PR?

anthonysena:
@ablack3 thanks for tagging me here. @johngrimes very sorry for the long delay here. I will take on the code review.

@johngrimes - let me make sure I understand the use case you are trying to address. You have a value set in FHIR and you'd like to import that as a concept set in ATLAS. Then you mention in the limitations:

> Matching is done on code only, in the same way the "Source Codes" import currently works. This could be enhanced when a proper mapping of FHIR code system identifiers to OMOP vocabulary identifiers is available, which is [currently being worked on](https://chat.fhir.org/#narrow/stream/179202-terminology/topic/OMOP.20Standard.20Terminologies).

I don't have access to the chat link you referenced and given the age of this PR, I'm unsure of the status of the mapping of FHIR to OMOP vocabularies. What happens if/when a value set is imported and no source codes are found in the vocabulary? There was some work done in #2653 that may be of interest to allow for importing of codes and then selecting of the vocabulary.

After quickly reviewing the code I had a few thoughts:

- Configuration of the FHIR endpoint URL in the `config/app.js` is the right approach. I'm wondering though if this should be set to an empty string by default so that the FHIR import is hidden when that value is not set? Having that tab displayed when there is no FHIR endpoint configured will be confusing.
- The cohort definition editor has a "Concept Sets" tab that allows for import of concept sets; we may need to include the FHIR import there as well.
- Since you mentioned that "Authenticated FHIR terminology endpoints are not yet supported" what are your thoughts on that level of support? I'm unsure how many people are using FHIR endpoints without security?

js/services/http.js
```diff
@@ -3,7 +3,7 @@ define(function(require, exports) {
   const config = require('appConfig');
   const sharedState = require('atlas-state');
   const { Api:OHDSIApi, STATUS } = require('ohdsi-api');
-  const JSON_RESPONSE_TYPE = 'application/json';
```
chrisknoll, line 6: This looks like it replaced JSON_RESPONSE_TYPE with FHIR_JSON_RESPONSE_TYPE, and I don't think you should remove JSON_RESPONSE_TYPE...can you confirm?

js/services/http.js
```diff
@@ -100,6 +100,18 @@ define(function(require, exports) {
     }
   }
 
+  class FhirApi extends Api {
```
chrisknoll, line 103: I don't think FhirApi should be a nested class in http.js.  The http.js is a module which facilitates http communication.  I think the FHIR class handles communicating to a FHIR service.   We have a separate folder of `services` which you can implement your class there, and then you can import the class in the javascript compoennt you want to use (via the define())

chrisknoll (changes requested):

js/services/http.js
chrisknoll: Apologies, my understanding about the http.js module was that it just wrapped some lower-level HTTP calls to handle passing specific headers, but not something that made some kind of 'client api' to a specific type of service (like the FHIR one or the PlainText one)...I'm not sure what the design decision was about introducing the PlainText api into http.js but it seems that whoever put that forward thought that you would put these type of API clients directly into the http.js module...so, I can't argue with that, so my only comment then is how the JSON_RESPONSE_TYPE  was removed....although not sure where that is being used (but possibly it's a constant that is refernced elsewhere in the app).

johngrimes:
@anthonysena Thanks very much for your time and detailed response.

> I don't have access to the chat link you referenced and given the age of this PR, I'm unsure of the status of the mapping of FHIR to OMOP vocabularies. What happens if/when a value set is imported and no source codes are found in the vocabulary? There was some work done in #2653 that may be of interest to allow for importing of codes and then selecting of the vocabulary.

The change looks great and really improves the import source codes feature. Once we have a solid vocabulary identifier mapping, I can imagine that you could enhance this implementation to remove the intermediate step in the user interface for FHIR source codes, as they are self-describing. Unfortunately I'm not sure that the work on the vocabulary mapping is complete, or I don't have visibility of it.

>     * Configuration of the FHIR endpoint URL in the `config/app.js` is the right approach. I'm wondering though if this should be set to an empty string by default so that the FHIR import is hidden when that value is not set? Having that tab displayed when there is no FHIR endpoint configured will be confusing.

It would be nice if there was a functional default for this, maybe "https://tx.fhir.org/r4" as the official reference server for the current version of the spec?

>     * The cohort definition editor has a "Concept Sets" tab that allows for import of concept sets; we may need to include the FHIR import there as well.

Is that a different component, or another instantiation of the same component? If there is a different one, could you point me towards where the source for that lives?

>     * Since you mentioned that "Authenticated FHIR terminology endpoints are not yet supported" what are your thoughts on that level of support? I'm unsure how many people are using FHIR endpoints without security?

I think that this change adds significant benefit even without this feature, as it allows users to interact with well-known terminologies that are available on public reference servers (e.g. tx.fhir.org, tx.ontoserver.csiro.au). I'd be keen to get your thoughts on what you think would be involved in adding authentication capability. I am not as familliar with the code base and any existing infrastructure or features of the framework that could be used for this purpose.

chrisknoll:
> Is that a different component, or another instantiation of the same component? If there is a different one, could you point me towards where the source for that lives?

I'm pretty sure there's one component being reused, similar how you defined your `conceptset-list-import-fhir`.  I can pull the branch and see how it appears, it may be the case that you need to add that component in another place...i seem to recall the concept set tabs under the main 'concept sets' nav is slightly different than the other plaes we have embedded cocnept sets (like Incidence, Cohort Defs and Cohort Characterization).

anthonysena:
@johngrimes - just reviewing the upcoming v2.13 release and we are going to move this item to v2.14 since we were unsure if you still wanted to contribute this to the codebase?

johngrimes:
I definitely still want to get it in, I just haven't had a chance to address the comments yet 😔

johngrimes:
There's probably not too much work remaining.

anthonysena:
No worries @johngrimes - we'll put it into v2.14 and we can check in with you next year. Thanks!

PR #2637, opened 2021-11-14
An improved `removeCriterion` (attribute) x-mark button

ehudkr:
**Problem:**
Current red cross-mark button used for removing added attributes in cohort criteria may be confused for an exclusion criteria - which is the opposite intent. See the warning in Book of OHDSI chapter 10.7.3 [[book](https://ohdsi.github.io/TheBookOfOhdsi/Cohorts.html#additional-initial-event-criteria)] [[source](https://github.com/OHDSI/TheBookOfOhdsi/blob/81a3897f10fea16355ce7285ed647393e889369c/Cohorts.Rmd#L203)].

**Proposed solution:**
Suggested solution may look and feel more like an ordinary "cancel" button, therefore avoiding the "exclusion" connotation.
(Additional tweaking is welcomed)

image atlas-remove_criterion

*Disclaimers:*
* Untested proposal
* I am not a UI/UX person (a better solution might exist)
* I have limited frontend experience (a better implementation might exist)


chrisknoll:
I like this, but I think there's some minor adjustments to where the assets are stored (ie: keep the image resource relative to the css).  Let me grab the branch and try some things, and if it works out we'll adjust the PR to go to a modified branch, but your commits will still be accepted.

anthonysena:
Hi @ehudkr we're just following up here to see if this PR is still needed? The graphic we see in ATLAS appears to be the same as what is currently in ATLAS. Could you help us see the visual difference in what you proposed vs. what is in ATLAS? In the meantime we'll move this item to the v2.14 release.

ehudkr:
Hi Anthony, thanks for checking up on this. 

Currently when you add an attribute to the inclusion criteria in Cohort Definitions, it appears next to a big red cross mark. The red cross mark is a cancel button, but it might also signal that the added attribute is actually an exclusion criterion (the exact opposite of what the user required). It was probably a source for enough confusion if The Book of OHDSI found it right to warn about it. 

The change suggested in this PR replaces the big red cross mark with a softer cross mark (that should resemble more a pop-up window closing icon that is hopefully more familiar).
The first message compares the current state (left) to what the suggested implementation looks like (right).

My (subjective) argument is that the softer cross mark conveys the "this is a cancel button" better and without invoking the "this is an exclusion criterion" connotation.

chrisknoll:
Hi, @ehudkr ,

I appreciate your work on this and taking the time to contribute.   There are some principles I'd like to adhere to and one of them is that we leverage assets from our icon library (font-awesome) for these icons and not introduce other things. This is so that we can upgrade one thing and everything involved will carry along with it.

I would like to propose an alternative to your approach here for improving the user experience.   Your instinct is supported by other feedback we've received from the community about these icons and how it may be confusing that these icons don't represent an 'action button' to remove but potentially some 'exclusion criteria' based on the icon.   My proposal is that we apply CSS rules like `:hover` to only show the icons for removal if the mouse is hovered. Then the troublesome crosses do not appear by default, but only when users hover over the element to remove it.  Would this satisfy the concern about the presence of the buttons and the look?

ehudkr:
Thanks for sharing your broader considerations, Chris, I was not aware of them, but as an open-source maintainer myself I do understand the desire to have a smallest possible set of dependencies. 

As for your proposed solution, I'm in no way a UX/UI expert, so I have no intuition as to whether having no indication that cancelation is possible by default (until hover) is preferred over the current or the proposed solution.
 
I was also unaware of Font-Awesome. However, running a quick "cross mark" search query there did bring up some candidates in line with my proposal that might be better than the current big red cross mark. See for example:
<a href="https://fontawesome.com/icons/circle-xmark?f=classic&s=solid">image </a> <a href="https://fontawesome.com/icons/circle-xmark?f=classic&s=duotone">image </a> <a href="https://fontawesome.com/icons/circle-xmark?f=sharp&s=solid"> image </a>
Which you may consider using instead of the svg I added if it alleviates your dependency concerns (given there are no additional restrictions I'm not aware of).

chrisknoll:
Circle x-mark looks pretty good (we can style it to be colored red) and I think if we can also apply styles such that the element does not appear unless hovered, that should give us an improvement to the look-and-feel of the form.  Is this something you'd be interested in trying?

chrisknoll:
Hi, @ehudkr .   Thanks for your attempts, but changing the definition of the style is definitely not something we want to do. Unfortunately, if we want to use a different icon from one to another, we need to refer to the new icon.

This simplest thing we can do on this effort is to apply a style that will hide the 'x' until there's a hover.   This way, we don't have the 'glaring' x-marks in the UI until someone hovers over and sees the prompt for delete.

chrisknoll:
Hello,
   We're going to close this as the approach depends on changing font-awesome.min.js.   The hover css is nice that adjusts the css to give feedback on hover, but I think if we want to change the 'remove' button from the glaring red x to something more subtle, we should just choose a new icon and change the reference.  At the same time, maybe it would make sense to define a global constant for the remove icon so that once we replace the hard-coded reference to the global constant, we can change all icons with 1 change.  That would be a good PR to work on (if we feel we will want to make changes to those icons again).

PR #2659, opened 2022-02-04
 Create Reusables module #2623 

### LINKED ISSUE #2623: Create Reusables module
_Targeting v2.11, more enhancements from Amgen & Odysseus. We've also discussed with other OHDSI sites. We intend to refine this in future releases._

**A new “Reusables” module will be added to the left navigation (with the recycle icon) between Prediction and Jobs module so as to not disturb the existing placement of key modules**

image

It will consist of:
1.	A library of available reusables
2.	A designer for new reusables

**Much like the “Custom Feature” option in Cohort Characterizations, in which users can build a reusable piece of cohort logic, we will allow the same in Reusables**
1.	Access to this interface could be limited based on Shiro access rules
2.	Tagging and versioning should be enabled for each reusable

**Parameters could be left open for design-time changes**

image

1. Meaning, we could have static concept sets baked into the reusable, or leave criteria concept sets unselected so that after injecting the reusable into a study asset, the user can then specify what to assign
2. When first importing a reusable with open parameters, a new UI should be present to highlight the parameters available to fulfill
3. We can also specify that a parameter value should be implemented in multiple places. For instance, if we wanted the same concept set to be used for more than 1 condition occurrence criteria in the reusable, we can select this option. Then in the import of the reusable, the user just selects 1 concept set in a new modal UI that would then be cascaded to all condition occurrence criteria prompts in the reusable.	

**Reusables could then be injected (by value) into places that use Circe logic:**
1.	Cohort Definitions (cohort entry events, restrict initial events, inclusion rules, censoring)
2.	Characterizations (custom features, subgroups)
3.	Incidence Rates (stratification criteria)
 
**The resulting JSON object and OHDSql should still conform to the current standards of Circe / Atlas**

@anton-abushkevich @alex-odysseus @gklebanov @chrisknoll @anthonysena

anton-abushkevich:
Resolves #2623


alex-odysseus:
@chrisknoll

Unfortunately the server-side implementation part has been already merged
https://github.com/OHDSI/WebAPI/pull/1948

PR #2683, opened 2022-03-22
Pinned and updated dependency versions

### LINKED ISSUE #2363: Include package-lock.json
### Expected behavior

The lockfile should be included to ensure consistent and performant builds, especially for CI.

```sh
time npm i
# ...
real    0m23.724s
user    0m0.000s
sys     0m0.077s
```

```sh
time npm ci
# ...
real    0m9.855s
user    0m0.015s
sys     0m0.046s
```

### Actual behavior

The package-lock.json was removed in https://github.com/OHDSI/Atlas/pull/1157/file. I haven't been able to replicate the issues described there. Any builds running on my local windows machine and within a docker build ran fine.

I can definitely create a PR with an updated .gitignore and package-lock.json.

chgl:
Given the recent challenges to the supply chain security of the NPM ecosystem (https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-23812, https://blog.sonatype.com/npm-libraries-colors-and-faker-sabotaged-in-protest-by-their-maintainer-what-to-do-now), ATLAS's security might benefit from pinning and locking dependency versions.

This PR updates all dependencies specified in the package.json to the latest non-major version, adds a package-lock.json, changes the Dockerfile build to use `npm ci` (https://docs.npmjs.com/cli/v8/commands/npm-ci), updates Node to the latest LTS version (16), and removes the unused `express.js` from the dependencies.

Closes #2363


anthonysena:
@chgl thanks for this work here. Upon review, we noted a few items:

- This PR re-introduces the package-lock.json which we had excluded previously. When doing an npm install, it would re-generate the package-lock.json and make it look like it changed and it would cause problems if installing on Mac/Windows. Let me find any related issues for this particular problem.
- We (me) want to take a look at some previous issues to see if `express` is being used here. 

If you could also resolve the conflicts on your branch, that would be helpful for our review. Thank you!

chgl:
Thanks for the feedback! I believe the first issue is a result of using unpinned dependencies with npm install, which will install the most recent version of a package within the semver range specified in package.json and update the lock file as a result. Using npm Clean-Install (https://docs.npmjs.com/cli/v9/commands/npm-ci) should use only the lock file and avoid needless changes to lock file. If there are issues with this on some plarforms, then it might be more convenient not to use the lock file of course. I'd still recommend pinning the dependencies.

I'll go ahead and rebate and resolve the conflicts later this week.

anthonysena:
After discussion with the Atlas/WebAPI WG, we'd like to move this forward but with a change in direction. We agree that using a fixed version for each dependency in package.json makes sense and we support that change. However, we do not want to include the package-lock.json in the repository as this file is not necessary and usually leads to confusion as it can vary between developers, operating systems, etc. We had encountered this issue with package.lock previously (I can find the issue if needed) so if you can put package-lock.json back into the .gitignore than these changes will work. Thank you!

chgl:
Sure, I stumbled upon a strange esprima parse error with some version which I'll have to resolve first, but other than that I'm still on it. I agree that diverging package lock isn't ideal but it should be avoidable when using npm ci (https://blog.npmjs.org/post/171556855892/introducing-npm-ci-for-faster-more-reliable). I believe the initial error from https://github.com/OHDSI/Atlas/pull/1157 (2018) may not actually be relevant nowadays anymore. I'll go ahead and test it on windows and in a Linux Container an report back to be sure. I'm ok with refactoring this PR to stick to pinning dependency versions and keep git-ignoring the lock file.

Note that the use of a lock file is generally recommended by the ossf as well, should we eventually decide to adopt one in a future pr if no issues remain; https://github.com/ossf/package-manager-best-practices/blob/main/published/npm.md#use-a-lockfile

chgl:
I've re-added package-lock.json to gitignore, I've split dependency-pinning and updates in two commits for better visibility - the major updates only affect dev dependencies, the other should be the actual versions installed as a result of the semver constraint anyways so there shouldn't be any regression (hopefully).

Just for completeness: I couldn't reproduce the issue from https://github.com/OHDSI/Atlas/pull/1157 or detected any unexpected changes to the package-lock file when installing using npm clean-install - on either windows or linux.

anton-abushkevich (approved):

anton-abushkevich (changes requested):
After clean rebuild: cohort builder not working.

1. create new cohort
2. try to add initial event or inclusion criteria - nothing happens, log empty too

Checked on node v12 and v14. Some library update breaks cohort builder functionality, need to investigate.

chgl:
@anton-abushkevich sorry about that! And thanks for double-checking! I've tested it locally and could reproduce the issue. I've now updated *all* libraries to the latest version accepted by the former semver range.

anthonysena:
Thanks @chgl could you merge master into your feature branch to resolve the conflicts here? Thanks!

anton-abushkevich (approved):

PR #2685, opened 2022-03-23
 Add warning message to reusables about using types #2684

### LINKED ISSUE #2684: Add warning message to reusables about using types
Add warning message to reusables: "You can switch between both types in one reusable but only one type can be active at one time (the one you save Reusable with)."

anton-abushkevich:
Resolves #2684


alex-odysseus (approved):
The message being added will support users not to loose their work

PR #2700, opened 2022-05-05
Fixed invalid redirect for /atlas without trailing slash

### LINKED ISSUE #2641: Invalid redirect when accessing ATLAS on a URL not ending in a trailing slash
### Expected behavior

When running the atlas container exposed on a port other than 8080 or when accessing the container from a reverse proxy and the `/atlas` URL does not end in a trailing slash, one should still be able to see the ATLAS UI.

### Actual behavior

The nginx server responds with a redirect to the server host and port with the trailing slash added to the URL. This seems to be a common issue: https://serverfault.com/questions/739498/nginx-rewrite-on-docker-machine-when-host-port-container-port

I was able to resolve it by using the following /etc/nginx/default.conf:

```nginx
server {
    listen       8080;

    root /usr/share/nginx/html;

    location ~ ^.*[^/]$ {
        try_files $uri @rewrite;
    }

    location @rewrite {
        return 302 $scheme://$http_host$uri/; # <- this uses the request header ('Host:') to determine the hostname+port
    }

    error_page   500 502 503 504  /50x.html;
    location = /50x.html {
        root   /usr/share/nginx/html;
    }
}
```

If it seems reasonable, I can create a PR for this.

1. `docker run --rm -it -p 9090:8080 ohdsi/atlas:2.10.1`
2. 
  ```sh
  curl -vvv localhost:9090/atlas
  *   Trying ::1...
  * TCP_NODELAY set
  * Connected to localhost (::1) port 9090 (#0)
  > GET /atlas HTTP/1.1
  > Host: localhost:9090
  > User-Agent: curl/7.55.1
  > Accept: */*
  >
  < HTTP/1.1 301 Moved Permanently
  < Server: nginx/1.20.1
  < Date: Fri, 26 Nov 2021 11:09:30 GMT
  < Content-Type: text/html
  < Content-Length: 169
  < Location: http://localhost:8080/atlas/ # <---- see here
  < Connection: keep-alive
  <
  <html>
  <head><title>301 Moved Permanently</title></head>
  <body>
  <center><h1>301 Moved Permanently</h1></center>
  <hr><center>nginx/1.20.1</center>
  </body>
  </html>
  * Connection #0 to host localhost left intact
  ```
3. thus this would redirect to localhost on port 8080 but the container is exposed on port 9090

chgl:
Closes #2641

When accessing the Atlas UI on a URL *not* ending in a trailing slash, NGINX would always automatically redirect you to `host:8080/atlas/`.

This change modifies the default NGINX configuration to use the host header for redirection.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,7 +26,7 @@ RUN find . -type f "(" \
       | xargs -0 -n 1 gzip -kf
 
 # Production Nginx image
-FROM nginxinc/nginx-unprivileged:1.20-alpine
+FROM nginxinc/nginx-unprivileged:1.23.0-alpine
 
 LABEL org.opencontainers.image.title="OHDSI-Atlas"
 LABEL org.opencontainers.image.authors="Joris Borgdorff <joris@thehyve.nl>, Lee Evans - www.ltscomputingllc.com"
@@ -41,6 +41,7 @@ ENV WEBAPI_URL=http://localhost:8080/WebAPI/ \
   CONFIG_PATH=/etc/atlas/config-local.js
 
 # Configure webserver
+COPY ./docker/nginx-default.conf /etc/nginx/conf.d/default.conf
 COPY ./docker/optimization.conf /etc/nginx/conf.d/optimization.conf
 COPY ./docker/30-atlas-env-subst.sh /docker-entrypoint.d/30-atlas-env-subst.sh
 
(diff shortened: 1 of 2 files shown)


chgl:
I didn't want to overload this PR by adding too much, but I would be happy to add a basic E2E test (using eg https://playwright.dev/) to at least make sure that the UI loads correctly.

leeevans (approved):

PR #2735, opened 2022-08-15
 Add selection of inclusion rule permutations in Intersection View #2726 

### LINKED ISSUE #2726: Add selection of inclusion rule permutations in Intersection View
### Expected behavior
The inclusion rule report in Intersection View should have the ability to explicitly select different permutations of rules. 

### Actual behavior
The user needs to click on blocks in the treemap. Often, these blocks get pretty tiny, making it a challenge to find the exact permutation of interest. 

### Steps to development
1. The data is already available to the client, so should not require any enhancement to the endpoint.
2. Add checkbox interface to the view to allow for selecting specific inclusion rule permutations.
3. Add a summary line item to the table indicating the permutations's statistics; this is already shown above the treemap, but having it near the table helps make the stats clearer. 
4. Highlight the permutation in the treemap with a color that isn't in the treemap's current color palette (perhaps blue?).

anton-abushkevich:
Resolves #2726

### DIFF
--- a/js/components/cohortbuilder/components/FeasibilityAttritionReport.html
+++ b/js/components/cohortbuilder/components/FeasibilityAttritionReport.html
@@ -16,7 +16,7 @@
 	</tr>
 </table>
 <div class="inclusionRuleType">
-	<table style="border-collapse: collapse; width:100%; max-width: 1190px; margin-left: auto;">
+	<table>
 		<thead>
 			<th style="width: 15px;"></th>
 			<th class="header">
--- a/js/components/cohortbuilder/components/FeasibilityIntersectReport.html
+++ b/js/components/cohortbuilder/components/FeasibilityIntersectReport.html
@@ -53,9 +53,9 @@
 							<td></td>
 							<td></td>
 							<td class="rules-summary-title">Summary</td>
-							<td>???</td>
-							<td>???</td>
-							<td>???</td>
+							<td data-bind="text: summaryValue().toLocaleString()"></td>
+							<td data-bind="text: summaryPercent().toFixed(2) + '%'"></td>
+							<td data-bind="text: summaryPercentToGain"></td>
 						</tr>
 					</tfoot>
 				</table>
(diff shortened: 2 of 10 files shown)


chrisknoll:
The only comment I have is:   the default selection of the inclusion rule filter is to specify 'any' that has 'passed'.  It should default to 'all' that have 'passed', because that's the default view that people are interested in., which also matches the 'matched' statistic at the top of the report.

chrisknoll:
I take it back, I played with this locally and when you set it to 'all passed' by default, the only box that lights up is the 'all passed' box (makes sense, right??) Instead, 'any passed' makes sense from a backwards-compatable perspective,  because each box represents a case where at least 1 passed, so the visualization looks the same between versions when you start with 'any'.  So, we'll leave it with any, unless we get some feedback where the 'all' default makes sense such that only the lit part of the graph represents the 'matching' population,....I honestly could go either way, but i'll change it back to 'any' because it makes the visualization look the same by default.

chrisknoll:
@anton-abushkevich , I've pushed up a commit as a suggestion on some restructuring of the code. I noticed the use of subscriptions and the need to call out to `greyRectsInTreemap` at strategic places, but there's a more 'automatic' way to handle this using the observables.  I kept much of our code, but with the following changes:

1. Removed Subscriptions:  instead of subscribing to different observables and taking actions based on subscriptions, instead I introduced an `afterRender` hook to the population treemap binding which is called after the SVG is rendered.   It is in this handler that we can do the work of graying out checkboxes.  Because this is invoked in the data-binding `update()`, any observable that is referenced (such as the selected inclusion rules, any/all and pass/fail options) will automatically force a refresh on the treemap when their observable udpates.  This also removes the need to remember to call `greyRectsInTreemap` because of the automatic depenency tracing will signal the refresh if any dependencies update.
2. Removed updates to observables in a loop:  I noticed there was some obserable updates (like a running total of events) during the rectangle loop, and instead I just declared a local variable and made one observable update after the calculation.  

Those are the main changes.   I'm sorry to push directly into your branch, but I wanted to try to present a best practice when dealing with databinding and observables.  If you have any objections, we can always revert this commit and come at it a different way.

chrisknoll:
Sorry about the force push (it was a quick bugfix I made in the loop of rectangles).  It shouldn't impact you.

chrisknoll (approved):

PR #2751, opened 2022-09-20
Phoebe 2.0 UI

### LINKED ISSUE #2728: Phoebe 2.0 Integration
### Overview

This issue tracks the features and requirements for PHOEBE 2.0 integration into Atlas.

###  Functions

1. Add a "Recommendation' tab to Concept set editors which will provide a list of suggested concepts from PHOEBE 2.0 based on the current concept set's 'included concepts'.

2. Add a sort of 'I feel Lucky' search which will take a search string in the vocabulary search, but when the user clicks the Phoebe search, the concept selected by PHOEBE will be automatically placed into a new concept set expression (with descendants selected).   The user is then redirected to the concept set editor with the 'recommended' tab selected.

### Mockups

To Be Added....

chrisknoll:
This PR provides the UI elements to support PHOEBE 2.0:

### Recommended Concepts tab

Conceptset editors show a recommended concept tab that will provide concept recommendations based on the included concepts in the concept set expression.  If the required tables do not exist in the CDM schema, an instructional message will be presented to the user on how to install the additional tables.

### Lexical Search

A new search button was added to vocabulary which will create a concept set based on the 'best' lexical concept that matches the search terms the user provided.

Fixes #2728.


js/services/Vocabulary.js
```diff
@@ -55,32 +61,19 @@ define(function (require, exports) {
...
 		}
 		return sourceKey ?
-			CDMResultAPI.getConceptRecordCount(sourceKey, searchResultIdentifiers, results, false) :
-			CDMResultAPI.getConceptRecordCountWithResultsUrl(sharedState.resultsUrl(), searchResultIdentifiers, results, false);
+			CDMResultAPI.getConceptRecordCount(sourceKey, searchResultIdentifiers, results, false, formatter) :
+			CDMResultAPI.getConceptRecordCountWithResultsUrl(sharedState.resultsUrl(), searchResultIdentifiers, results, false, formatter);
 	}
 
-	function search(searchString, options) {
+	async function search(params) {
```
anton-abushkevich, line 68: no point put 'async' here without 'await' in the body of the function

js/services/Vocabulary.js
```diff
@@ -159,11 +152,16 @@ define(function (require, exports) {
 			method: 'POST',
 			contentType: 'application/json',
 			error: authAPI.handleAccessDenied,
-	});
+		});
 
 		return getMappedConceptsByIdPromise;
 	}
 
+	async function  getRecommendedConceptsById(identifiers, url, sourceKey) {
```
anton-abushkevich, line 160: no point put 'async' here without 'await' in tthe function

js/services/Vocabulary.js
chrisknoll: I think there is because async means that we're returning a promise....right?

chrisknoll: So, let me clarify:   it's not required to define async when you don't use await, however async does indicate that a promise is returned, so I say we leave it in because it's more informative that the function is an async function vs. not knowing if it would return a promise or not.

anton-abushkevich: async is simply ignored by the JS interpreter, if there's no await. 
As a developer, when I see async, I expect await in the function, so if there's none it's confusing. So I suggest to keep the code clean.

js/services/Vocabulary.js
chrisknoll: I'm sorry i accidently clicked resolve conversation, but to your comment about 'if the function is async, there shoudl be an await',

This is not true, and the only thing about async that you should expect is that the function returns a promise. It's not true that the JS interpreter just 'ignores' async when there is no await:  if the function doesn't return a promise, then the function will wrap a promise around the 'raw' value so that the async function returns a promise.

See this for details: https://javascript.info/async-await.

Because this function returns a promise, and I want developers to know that this function returns a promise, I'm leaving this function as 'async'.

js/services/Vocabulary.js
chrisknoll: This is not true, and the only thing about async that you should expect is that the function returns a promise. It's not true that the JS interpreter just 'ignores' async when there is no await:  if the function doesn't return a promise, then the function will wrap a promise around the 'raw' value so that the async function returns a promise.

See this for details: https://javascript.info/async-await.

Because this function returns a promise, and I want developers to know that this function returns a promise, I'm leaving this function as 'async'.

anton-abushkevich: Yes, sorry, you are right.

anton-abushkevich (approved):

anton-abushkevich (approved):

PR #2789, opened 2022-12-06
Generation / Executions UI with many data sources #2776 

### LINKED ISSUE #2776: Generation / Executions UI with many data sources
- design one standard UI for the generation / execution of a study asset (Cohort Definition, Characterization, Incidence Rate, Cohort Pathway, Patient Level Prediction, Population Level Effect Estimation)
- allow for search or filter boxes to allow users to key in the name and limit the number of options

anton-abushkevich:
Resolves #2776

1. All Generation/Execution sources lists now work as datatable:
-- user can filter sources by name
-- datatables display items (sources) count and add pagination when necessary
2. Added checkbox to hide sources without generation/execution results
3. Two types of generation/execution tables 
-- for Characterizations, Pathways, PLP and PLE (with results versioning)
-- for Cohorts and Incidence Rates (without versioning, displaying results in the table)


chrisknoll (changes requested):
I noticed the use of subscriptions, but I'd like to remove subscriptions in favor of pureComputed (subscriptions will need to be disposed of, and the 'flow' is more logical to use computed instead of responding to subscriptions.  here is how:

Define a prueComputed to yield the `filteredExecutionGroups`, something like:

```
  this.filteredExecutionGroups = ko.pureComputed(() => {
    return (this.showOnlySourcesWithResults() ? 
                  this.executionGroups().filter(eg => eg.submissions().length > 0)
                  :  this.executionGroups());
  });
```

In this way, the pureComputed re-calculates automatically if the checkbox changes (the `showOnlySourcesWithResults`) or if the executionGroups() changes.  Note, may need to add logic that calls this.executionGroups.valueHasMutated() if there is a background process that refreshes the execution groups when jobs are running.

Since your datatables are being bound to filteredExecutionGroups, that code won't need to be changed if filteredExecutionGruops becomes a computed observable.

anton-abushkevich:
@chrisknoll 
Subscription is replaced with pureComputed as you suggested - please review: 1a888df7f1bd18c5d5cf3a016de29000796b7c5c

chrisknoll (approved):
Great PR:  the changes were made exactly where this feature request would expect them to be, it's following good coding conventions, and is a great feature with minimal configuration or administrative burden.  Good job.

PR #2790, opened 2022-12-06
Add highlights sql output

### LINKED ISSUE #2775: Cohort SQL output – UI enhancements
- emulate the SqlRender Shiny app to allow easy parameter entry
- add color coding of keywords like typical SQL IDEs have

anton-abushkevich:
Resolves #2775 

- added SQL syntax highlight
- added variables (parameters) replacement (switch between parameters names and parameters values)

### DIFF
--- a/js/components/utilities/sql/sqlExportPanel.html
+++ b/js/components/utilities/sql/sqlExportPanel.html
@@ -1,22 +1,12 @@
 <div>
     <loading data-bind="visible: loading"></loading>
     <!-- ko ifnot: loading -->
-    <!-- ko if: sourceSql -->
-    <h5 class="title">
-        Source output:
-    </h5>
-    <!-- /ko -->
-    <!-- ko ifnot: sourceSql -->
-    <h5 class="title">
-        Rendered translation:
-    </h5>
-    <!-- /ko -->
     <div class="flex-container">
         <div class="sql-text">
-            <!-- ko if: sourceSql -->
+            <!-- ko ifnot: sourceSql() -->
             <pre  data-bind="html: sqlText, attr:{ id: clipboardTarget }"></pre>
             <!-- /ko -->
-            <!-- ko ifnot: sourceSql -->
+            <!-- ko if: sourceSql() -->
             <pre  data-bind="html: paramsTemplateSql(), attr:{ id: clipboardTarget }"></pre>
             <!-- /ko -->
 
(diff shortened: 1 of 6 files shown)


chrisknoll (changes requested):
When the view initially presents, the options text input fields are read only.  However, you can click 'replace values' (which will replace the defaults in the SQL output, and then the fields become enabled.   I think these text fields should always be enabled (ie: never read-only).

chrisknoll (changes requested):
Let's make some changes to behave more like a knockout application where we depend on observables instead of managing our own subscriptions:

1.  Replace ko.compuated() with ko.pureComputed when the observable is just a computed alue (ie: the sqlParams).
2. Remove subscriptions and instead just store user-entered fields as observables.
3. sqlParamsList should return a key-value where the key is the param name, and the value is an observable(). This way, changing the value of the input will automatically trigger the refresh of the SQL.
4. sqlText should be a pureCompuated()) calculated by replacing the sqlText() with the specified paramater values in sqlParamsList (if checkbox is checked) and then render the html with highlightJS().

The flow of the component should be:

1. The root 'sql' comes is as a component param and is observable (which will update when cohort definition changes or a new cohort definition is loaded)
2. sqlParams (pureCompted returning an array) depends on `sql`, which will yield a name-value pairs of paramater names and observables() which store the user input.  This pureCompted will be used in the `foreach: sqlParams`.  If the sql canges in some way, this will refresh these params. and re-render the foreach elements.
3. sqlText() will be a pureComputed() which will generate the HTML to render in the `pre` element.  This can be bound to a pureCompuated which will do the appropraite find/replace and SQL highlighting from sql.  This means that if sql, sqlParams, or the observables inside sqlParams change, this computed will refresh and the `<pre>` element will be updated.
4. The input fields can be bound to the `value:` databinding so that they don't update the observables until the textbox is unfocused (via  tabbing off the field).

This should eliminate the need to maintain your own subscriptions, and handle any 'input' events in the event handler.   

In addition, is the spread `...` operator necessary, can we not just return a 'Set' here:

```
		sqlParamsList(templateSql) {
			const regexp = /@[-\w]+/g;
			const params = templateSql.match(regexp);
			const paramsList = new Set(params);
			return [...paramsList];
		}
```

TitrS:
@chrisknoll Hello Chris! Thank you for your comment. I changed the logic of component after that.  I tried to realize the full flowm but i didn't. 
`sqlText` and `sqlParams` depend on not only from `sql`. It depens on from dialect also, we have to change it after async request `await cohortService.translateSql ` so i kept them as `observable`. Especially thank you for help with subscriptions. It looks better now! Let'me know if i have to change some else or if i misunderstood something.

chrisknoll (approved):

PR #2792, opened 2022-12-06
Add conceptset preview

### LINKED ISSUE #2791: Add Concept Set preview
Users aren’t confident with the impacts of adding/subtracting a concept to their concept set in terms of included source codes. This is only possible by opening each concept in a new tab and exploring related concepts.

Solution: add a preview button next to the Add to Concept Set button that then brings up a modal window with a preview of: expression, included concepts, included source codes, and export for the set with the newly checked concepts.

anton-abushkevich:
Resolves #2791


chrisknoll (changes requested):
When I bring up the preview, and change some of the existing elements in the preview, but then click 'cancel', the existing elements that I changed remain changed.

Should the preview use a 'copy' of the concept set expression for preview, and then 'apply' copies the copy into the final concept set expression?

anton-abushkevich:
@chrisknoll 
You are absolutely right. Please review the last commit - I added copying of existing CS items, instead of including them into the preview by reference.

Please note the `ConceptSetItem.js` changes. It turns out, that when creating a new object on the basis of existing one, the observable fields `isExcluded`, `includeDescendants` and `includeMapped` were always true, because for example `data.isExcluded` is a reference to a function, and in the expression `data.isExcluded || false` interpreted as true, so the result was always true. I fixed it by adding a call to extract the values from there observables, like `data.isExcluded && data.isExcluded() || false`.

js/components/conceptset/InputTypes/ConceptSetItem.js
```diff
@@ -5,9 +5,9 @@ define(['knockout', 'conceptpicker/InputTypes/Concept'], function (ko, Concept)
 		var self = this;
 		
 		self.concept = data.concept && new Concept(data.concept);
-		self.isExcluded = ko.observable(data.isExcluded || false);
-		self.includeDescendants = ko.observable(data.includeDescendants || false);
-		self.includeMapped = ko.observable(data.includeMapped || false);
+		self.isExcluded = ko.observable(data.isExcluded && data.isExcluded() || false);
+		self.includeDescendants = ko.observable(data.includeDescendants && data.includeDescendants() || false);
+		self.includeMapped = ko.observable(data.includeMapped && data.includeMapped() || false);
```
chrisknoll, line 10: Don't do this, data should never come in as an observable to the constructor.

If you want to make a 'copy' and re-observe it, take the origional, do a ko.toJS() on the object, then re-instantiate.

ie:

```
var origional; // this is the existing thing to clone

var clone = new ConceptSetItem(ko.toJS(origional))
```

```

File: js/pages/vocabulary/components/search.js [hunk 2]
```diff
@@ -540,7 +537,10 @@ define([
 			const concepts = commonUtils.getSelectedConcepts(this.data);
 			const items = commonUtils.buildConceptSetItems(concepts, options);
 			const itemsToAdd = items.map(item => new ConceptSetItem(item));
-			this.previewConcepts(itemsToAdd.concat(this.conceptSetStore.current() ? this.conceptSetStore.current().expression.items() : []));
+			const existingConceptsCopy = this.conceptSetStore.current()
+				? this.conceptSetStore.current().expression.items().map(item => new ConceptSetItem(item))
+				: [];
+			this.previewConcepts(itemsToAdd.concat(existingConceptsCopy));
```
chrisknoll, line 543: Here, you would just say:

```
const existingConceptsCopy = new ConceptSet(ko.toJS(this.conceptSetStore.current()));
```

Although what i'm suggesting above assumed that the ConceptSetPreview would be working off of a ConceptSet and not an array of ConceptSetItems...which is fine but in that case, I should say that the change is:

```
const existingConceptsCopy = this.conceptSetStore.current()
				? this.conceptSetStore.current().expression.items().map(item => new ConceptSetItem(ko.toJS(item)))
				: [];
```

And then, when the 'Apply' button is clicked in the preview editor, that will overwrite this.conceptSetStore.current() with the new state of the `ConceptSetItem`s.

But, overall, you have the right approach:   you need to clone the existing concept set for the user to modify, and only 'commit' the change when they click 'Apply' and discard those changes (ie: do nothing) if they cancel.

chrisknoll: Note, that last part (where when the click is made, the finalized change is saved back to the current concept set...if we're working off a clone, doesn't the changes made to the clone need to propagate tack to the object?

chrisknoll (changes requested):

js/components/conceptset/InputTypes/ConceptSetItem.js
anton-abushkevich: Ok, reverted

js/pages/vocabulary/components/search.js
anton-abushkevich: Fixed. 
Cloned concepts are concatenated with selected in the search results. They all stored in this.previewConcepts() and passed to current concept set in the addPreviewConcepts() method.

chrisknoll:
The `onPreview` param is only being set in `search.js` but if you use the preview button within a concept set editor, you get an error.

anton-abushkevich:
@chrisknoll 
Do you think we need the preview in places other than the Search page? If yes, than sure - I can move preview handling into the ConceptAddBox or make some wrapper around it and move preview functionality there. For now I added a check - the Preview button is hidden if no preview handler provided. 

Also I fixed multiple CS stores behavior - please review.

anthonysena:
@anton-abushkevich from discussion today on the Atlas WG, we feel that having such a concept set preview would be useful wherever we work with Concept Sets in the user interface. So if we could modify this PR to have this capability, that would be a great addition to this enhancement.

chrisknoll (approved):

PR #2794, opened 2022-12-06
Add venn diagram

### LINKED ISSUE #2793: Add Venn diagram for clear concept set comparison
The concept set comparison table can be hard for users to interpret.

Solution: add a Venn diagram view to the concept set comparison tool to show the scope of differences more clearly between 2 concept sets.

anton-abushkevich:
Resolves #2793


chrisknoll (changes requested):
After clicking compare button, I get this error in console:

`knockout-latest.js:63 Uncaught Error: Cannot find closing comment tag to match:  ko if: compareResultsSame() `

PR #2806, opened 2022-12-14
Add chart drilldown to concept pages

### LINKED ISSUE #2805: Add Achilles drilldown reports to Concept / Concept Set pages
The concept and concept set pages lack any practical time series statistics. These statistics can only be viewed in Data Sources; and in the case of concept sets, this information can’t be accurately pulled for the concept set.

1. Add Data Sources time series and charts to concept pages, so that users can drill down into these insights while examining concepts.
2. For concept sets, add an “Explore Time Series” tab that allows a user the ability to generate time series information for the concept set as a whole, against the CDMs available.
3. The execution may span multiple CDM clinical tables (e.g. condition occurrence, drug exposure, etc, based on the domains of the concepts).
4. The charts would be same as drilldowns in Data Sources – Prevalence by gender and year; prevalence by month; and age at first occurrence box plots, by gender.

anton-abushkevich:
Resolves #2805


chrisknoll:
From the UI perspective, let's remove the 'time series' functionality from the concept sets:   the way the aggregation is done across concept sets (talking the average of p10 or summing the monthly prevalence) is incorrect math, and an alternative approach to presenting the data could be identified. 

The drilldown reports from the concept view is fine and helpful: it saves navigating to data-sources and finding the concept in the right domain.  This is helpful and can stay.   I'd keep that function for 2.13, and think about how to deal with 'concept-set reporting' as a 2.14 feature.

chrisknoll:
Thanks. For future reerence, be careful about force-pushing a branch, if you want to go back to a prior commit, just revert to the desired commit which will cause git to make a 'revert commit' that removes the changes. I was lucky to notice you forced pushed the branch because if I tried to merge your latest commits' back to my local branch, it would have been chaos.

chrisknoll:
Ok, The Atlas update looks good, but I'd like to also remove the endpionts on the WebAPI side that has those improper calculations that aggregate statistics.   Before merging the Atlas side, let's remove the code from webAPI.  To do so, please revert commits (don't force push branches) and I'll take a look.

TitrS:
Hi @chrisknoll thanks for your comments. If we decide don't add the functionality for concept sets page, you just should reject PR on the WebApi side [OHDSI/WebAPI#2178](https://github.com/OHDSI/WebAPI/pull/2178), because  on the WebApi side all changes are related to concept sets page.

chrisknoll (approved):

PR #2835, opened 2023-02-22
Reduce polling to /user/refresh

### LINKED ISSUE #2820: UI Laggs when retrieving results from /refresh
### Expected behavior
UI should be responsive when clicking

### Actual behavior
The permission configuration we have is leading to a huge `permissions` payload when fetched from the server with /refresh.  See attached for reference. The size of the payload is 257KB as seen in this screenshot:

image

Our environment may be unique: we have 20-30 active CDM sources configured in webapi.  However, we don't adopt any special groups or role permissions.   We only leverage source-level permissions to generate and view results which should be fairly standard practice for security in atlas.

Note, opening the attached file (256KB) in notepad++ makes the app hang (probably trying to parse something in the file).

[refreshPayload.txt](https://github.com/OHDSI/Atlas/files/10526082/refreshPayload.txt)

chrisknoll:
Add refreshTokenThreshold to limit polling for permissions in Route.js.
Alter polling intervals to use app config settings.

Fixes #2820.


js/pages/Route.js
```diff
@@ -11,7 +11,7 @@ define([
 		checkPermission() {
 			if (authApi.authProvider() === authApi.AUTH_PROVIDERS.IAP) {
 				return authApi.loadUserInfo();
-			} else if (appConfig.userAuthenticationEnabled && authApi.token() != null && authApi.tokenExpirationDate() > new Date()) {
+			} else if (appConfig.userAuthenticationEnabled && authApi.token() != null && this.timeToExpire() < appConfig.refreshTokenThreshold) {
```
anton-abushkevich, line 14: Permissions are updated along with the token refresh, so this logic will stop the permissions update until the threshold (which will be 8 hours from the moment of login with default settings: 12 hours token lifetime minus 4 hours threshold).

So if another user gave (or revoked) this user a pemission to some cohort/cs/etc., then the user will have to refresh the page to recieve updated permissions. In case of revoked permission, user will lose their changes and recieve 403 while trying to save the asset (and refresh the token and permissions list in the process, see [js/services/AuthAPI.js:171](https://github.com/OHDSI/Atlas/blob/master/js/services/AuthAPI.js#L171)).

Maybe consider refreshing permissions at least when open the asset editor?

js/components/analysisExecution/analysis-execution-list.js
```diff
@@ -152,7 +152,7 @@ define([
     startPolling() {
       this.pollId = this.PollService.add({
         callback: silently => this.loadData({ silently }),
-        interval: 10000,
+        interval: config.pollInterval,
```
anton-abushkevich, line 155: Default polling interval is 1 minute. It was 10 seconds, so maybe add different polling intervals to the settings to keep the user experience?

js/pages/Route.js
chrisknoll: I believe that if you make a request that would violate permissions, a permission refresh is triggered (I believe it checks for status 403).  See [here](https://github.com/OHDSI/Atlas/blob/21c925c3b933929944ec0e0e8055db712214ebed/js/services/AuthAPI.js#L165).


My understanding/expectation is that when we get an exception that results in an accessDenied, it calls out to refrshToken(), which should call out to user/refresh, and then process the permissions.

The client state of permissions should have no impact on the permissions on the server.   if a permission is changed, it should be immediately in effect on the server, but potentially delayed from being known on the client.   If the code works as I think, then once we find out of a permission problem, we refresh perms from the server (and the UI will reflect the new permissions).

We could refresh permissions on every request, but that would bring us back to the original problem being solved here.

js/components/analysisExecution/analysis-execution-list.js
chrisknoll: I think we should use consistent poll intervals across the polling....for now, we should fix the bug where people used hard-coded poll intervals, but later we can introduce new configuration params for different poll contexts if we need it.

chrisknoll:
I've updated this PR to create a common executeWithRefresh() method in AuthAPI that will prevent a lot of boilerplate code that I was copying to call the HTTP request and then follow with refreshToken().

Some elements look awkward, but it's because a single function (in most cases: save()) had 2 different paths:   in the case it is a create-type of save, it would do a POST + refresh token.  In the case of a update-type of save, it would only do a PUT.   And then on top of that, there might be additional promise-chain type of activity, that was just too complciated to try to share between each path...so you might see duplicate code, but it's too complicated to share.

@anton-abushkevich : if you could run this branch on your environment and let me know if you see any issues. I would like to include this in 2.14.

js/pages/Route.js
chrisknoll: I've added permission refreshes on load, create (because we need to know the new one has new permissions), and copy.   That should cover the cases where we have permissions changes during login.

Longer term, we need another way to check permissions on entities.  It should be a simple flag on an entity if you have read/write access to it (sources may work slightly differently), but the current approach means we download nearly a megabyte of permission strings and parse them every time we want to ask a simple question 'is this element readable/writeable'.

js/services/Prediction.js
```diff
@@ -9,28 +9,30 @@ define(function (require, exports) {
...
 	}
 
-	function savePrediction(analysis) {
+	async function savePrediction(analysis) {
 		const url = config.webAPIRoot + predictionEndpoint + (analysis.id || "");
-		let promise;
+		let result;
 		if (analysis.id) {
-			promise = httpService.doPut(url, analysis);
+			result = await httpService.doPut(url, analysis).catch((error) => {
```
anthonysena, line 16: This looks like it needs the `authApi.executeWithRefresh` wrapper function?

anthonysena (approved):

PR #2849, opened 2023-04-04
Add read permissions

rkboyce:


rkboyce:
The style concerns focused the proper use of variable types in Permisssion.js and the use of 'perm_type' instead 'role' in parameter names where the former was the correct choice.

There is now a new configuration option for configLocal : 

```
        // This tells Atlas to show the 'lock' icon to the user when
        // viewing concept sets, cohort definitions, and other
        // artificats so that the can edit READ and WRITE
        // permissions. By default, this is set to true
        configLocal.enablePermissionManagement = true;
```
I added code throughout the pages to hide the button when the enablePermissionManagement = false;

I did not address using tabs or pills for the configure access modal now that read permissions are a part of that. Neither the ability to assert that an artifact is globally readable. I would like to punt these changes to 2.15 because of limited time on my part.

PR #2860, opened 2023-05-15
Correct vocabulary for person counts is used

### LINKED ISSUE #2859: Record Counts shows wrong data source
Atlas uses the first vocabulary in the list to show record and person counts

ssuvorov-fls:
fixes #2859

### DIFF
--- a/js/pages/vocabulary/components/search.html
+++ b/js/pages/vocabulary/components/search.html
@@ -79,7 +79,7 @@
           options: resultSources,
           optionsText: 'sourceName',
           optionsValue: 'sourceKey',
-          value: currentResultSource,
+          value: currentResultSource().sourceKey,
           event: { change: refreshRecordCounts }
         "></select>
 		</div>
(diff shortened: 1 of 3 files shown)


js/pages/vocabulary/components/search.html
```diff
@@ -79,7 +79,7 @@
           options: resultSources,
           optionsText: 'sourceName',
           optionsValue: 'sourceKey',
-          value: currentResultSource,
+          value: currentResultSource().sourceKey,
```
chrisknoll, line 82: This looks like 1currentResultSource()1 is an observable to a `Source` object, but should we be binding the dropdown to a full-source object instead of just a observable that stores what the current source selection?   I can't tell from this review, but what is in `currentResultSource()` and should we be updating the 1 field in it (sourceKey) when instead it would make more sense to just store the currentResultSource's key?  Maybe this var should be called `currentResultSourceKey`?

chrisknoll (changes requested):

chrisknoll (approved):

Issue #2857, opened 2023-05-10
accept authenticated remote_user from reverse proxy header


clarkevans:
For deployments in a cloud environment, user authentication is often upstream of OHDSI tools, including Atlas. By convention an HTTP header can be provided by a reverse proxy, containing the email address of the authenticated user.  I suggest that 'X-Remote-User' could be the header name used for this purpose, but I'm not being prescriptive; perhaps it should be a configuration option, that, when left unset doesn't enable this feature.

If this is an easy feature to implement, it is high value, especially for institutions that already have validated security procedures that use reverse proxy authentication.

anthonysena:
@konstjar @alex-odysseus any thoughts on this?

konstjar:
My thoughts on this. I would avoid relying or using any (custom) headers for user authentication. It could be a security breach to get access to application including admin permissions.

What I would recommend the following and what was raised in https://github.com/OHDSI/WebAPI/issues/2268#issuecomment-1538866065

1. Enable SSO (OpenID or SAML) authentication workflow
2. If WebAPI is configured with OpenID or SAML, we can force user redirect in browser to Identity Provider authentication URL.

But this could be done if only one and single security provider is configured in ATLAS config.

clarkevans:
> My thoughts on this. I would avoid relying or using any (custom) headers for user authentication. It could be a security breach to get access to application including admin permissions.

The standard reverse proxy setup keeps applications like Atlas on a private network that is otherwise inaccessible.
What I'm asking for is actually more secure since requests that get to Atlas are already authenticated with high-level authorization... or they don't get past the reverse proxy.  What we want is to avoid annoying our user, and continue to use the rest of Atlas user authorization to associate it with roles and other access specific to Atlas.

What would be needed is an exact configuration parameter, say "Remote_User_Header="X-TuftsRemoteUser", that if unset, has the current behavior. If this parameter is set, then the existence of that header will be asserted and its value becomes the username.  The reverse proxy would be setting this header, if someone tries to submit this header with another value, it'll be overwritten with what ever authentication the reverse proxy has. Hence, there's no way for this header to be set by the user to bypass security.

This mechanism is standard, a great many web applications have these features, and it permits the reverse-proxy to be the user authentication gateway, centralizing control, etc.

PR #2861, opened 2023-05-17
Add option to force authentication

RowanErasmus:
This is half related to these two issues:
#https://github.com/OHDSI/Atlas/issues/2857
#https://github.com/OHDSI/WebAPI/issues/2268

The idea is to automatically log in to Atlas when a user is already authenticated with SSO.

To this end I have added a configuration userAuthenticationForced (better suggestions are welcome)
If this is enabled and a user is not logged in it will open the sign in modal, if there is only one auth provider it will trigger that one. As a result if a session is already established with your provider it will just log you in without any clicks.

If you have multiple outlook accounts and are using Azure AD it plays nicely with #https://github.com/OHDSI/WebAPI/pull/2254 to suggest which account it should be using

A video of what it looks like when not yet authenticated:
https://github.com/OHDSI/Atlas/assets/74538318/b1c93559-39a5-4602-b8a8-6858a0ef2ec9

If you have already have a session with your provider but not with Atlas it goes like so:
https://github.com/OHDSI/Atlas/assets/74538318/4595193d-a7f1-4809-adb4-e9d93a6479fd

It is not very 'smooth' and takes a couple of seconds but it's good enough for me :-)
 
(I've also fixed some indentation in the welcome.js file, although it still looks a bit weird in the pr...)

### DIFF
--- a/js/config/app.js
+++ b/js/config/app.js
@@ -11,7 +11,7 @@ define(function () {
     appConfig.pollInterval = 60000;
     appConfig.cohortComparisonResultsEnabled = false;
     appConfig.userAuthenticationEnabled = false;
-    appConfig.enableSkipLogin = true; // automatically opens login window when user is not authenticated
+    appConfig.enableSkipLogin = false; // automatically opens login window when user is not authenticated
     appConfig.plpResultsEnabled = false;
     appConfig.useExecutionEngine = false;
     appConfig.viewProfileDates = false;


chrisknoll (changes requested):
Looks good, but make 2 changes to behavior:

1.  Rename flag to `enableSkipLogin` which will automatically attempt to open the login when your authentication fails in `loadUserInfo()`.
2.   Remove the check to the flag in `welcome.js` so that if you only have 1 provider, you invoke it automatically (we don't need a flag for this behavior).

RowanErasmus:
@chrisknoll as agreed upon during yesterdays meeting I have renamed the flag, made the logon with 1 provider always automatically go by default and as agreed I hit autoformat on the all over the place alligned app.js

I very much second this https://github.com/OHDSI/WebAPI/issues/2203 btw, automagically consistently formatted code is a blessing well worth that initial single mega pull request in my opinion ;-)

RowanErasmus:
@chrisknoll 
I took a closer look at the names and there are 12 flags of which 4 are 'enable ...' and 3 are '...Enabled'.
I guess more importantly there are 9 that start with the verb so probably that would be the way forward ;-)
So I've changed it to enableSkipLogin

I also set the default to false, that seemed to have slipped through...

cacheSources
cohortComparisonResultsEnabled
disableBrowserCheck
enableCosts
enablePersonCount
enableTaggingSection
enableTermsAndConditions 
plpResultsEnabled
showCompanyInfo
useExecutionEngine
userAuthenticationEnabled
viewProfileDates

chrisknoll:
Thank you for your patience.   This probably feels pretty arbitrary since the rules are all over the place, and I didn't notice the ones that were xxEnabled but I think starting with enable makes more sense in the long run and we can think about a PR to normalize that one later.

I also appreciate changing default to false because we want to maintain backwards behavior where it makes sense and have people 'opt in' to new behavior (if they are given the choice).  

This looks good, thanks very much.

chrisknoll (approved):

chrisknoll:
Can you do me a favor: can you go through the issues that are opened that can be closed with this change?  I had some trouble trying to determine if the referenced issues should be closed, or if this was a stand-alone feature (bypass clicking on the authentication provider) and the other issues are still open.

RowanErasmus:
Closed one of the issues, the other has not really been resolved as the person raising it was asking for something completely different ;-)

Issue #2586, opened 2021-07-28
Add SOLR options in Configuration screen


alondhe:
With SOLR as a viable option for vocab search, the only way to know it is enabled is in the WebAPI/info endpoint.

Can we add information about the SOLR source to Configuration, and add buttons for "Check connection" and "Re-import data"? Perhaps we can also allow enabling/disabling of the SOLR source.

alondhe:
Re-importing data won't be feasible. Solr 8.11.1 is one of the last versions that support dataimport via jdbc natively, we need to determine how best to handle this in more up-to-date Solr versions.

PR #2869, opened 2023-06-14
Adds Solr Vocab status to Configuration page

alondhe:
This addresses part of #2586. 

Only is visible if the app config's useSolrVocabSearch setting is true. If false (which is by default), then you see:

image

If set to true but no cores are accessible based on the WebAPI's info endpoint:

image

If set to true and at least 1 core is available based on the WebAPI's info endpoint, the first core is shown. This matches current functionality, but we should look at being able to pick different cores.

image

### DIFF
--- a/js/pages/configuration/configuration.js
+++ b/js/pages/configuration/configuration.js
@@ -45,6 +45,7 @@ define([
       this.sourceJobs = new Map();
       this.sources = sharedState.sources;
       this.isSolrVocabEnabled = ko.observable();
+      this.isSolrVocabCoreAvailable = ko.observable();
       this.solrVocabCoreName = ko.observable();
 
       this.priorityOptions = [
@@ -120,7 +121,8 @@ define([
       const info = await buildInfoService.getBuildInfo();
       await sourceApi.initSourcesConfig();
       super.onPageCreated();
-      this.isSolrVocabEnabled = info.configuration.vocabulary.solrEnabled;
+      this.isSolrVocabEnabled = info.configuration.vocabulary.solrEnabled && info.configuration.vocabulary.hasOwnProperty("cores");
+      this.isSolrVocabCoreAvailable = false;
       if (this.isSolrVocabEnabled) {
         this.solrVocabCoreName = this.getSolrVocabCoreName(info);
       }


chrisknoll:
Hi, @alondhe , we spoke off line, but for the record: let's avoid the need to set a new config param to make the core versions show, and instead depend on the informatino returned from `/info` to tell if solr is enabled, and which cores are enabled, and render the UI based on that.  This prevents someone 'forgetting' to set a param and simplifies the functionality.

js/pages/configuration/configuration.js
```diff
@@ -120,7 +121,8 @@ define([
       const info = await buildInfoService.getBuildInfo();
       await sourceApi.initSourcesConfig();
       super.onPageCreated();
-      this.isSolrVocabEnabled = info.configuration.vocabulary.solrEnabled;
+      this.isSolrVocabEnabled = info.configuration.vocabulary.solrEnabled && info.configuration.vocabulary.hasOwnProperty("cores");
+      this.isSolrVocabCoreAvailable = false;
```
chrisknoll, line 125: At the top of your object, you've defined isSolrVocabEnabled as an observable, but on this line (126) you are overwriting it with a  boolean.  So, if it's not important to have this value observable (ie: have the UI respond to changes in this value) then just declare it to default to 'false'.   Or, if you do want the UI to respond to changes in the value of that variable: then you should say `this.isSolrVocabCoreAvailable(false)`...btw, just looking at your coe changes, it seems that it is always false...unless I'm missing whre this.isSolrVocabCoreAvailable is set to true somewhere?

chrisknoll (changes requested):

js/pages/configuration/configuration.js
```diff
@@ -113,11 +118,25 @@ define([
...
+      if (this.isSolrVocabEnabled) {
+        this.solrVocabCoreName = this.getSolrVocabCoreName(info);
+      }
       this.loading(false);
     }
 
+    getSolrVocabCoreName(info) {
+      if (typeof(info.configuration.vocabulary.cores) != "string") {
+        this.isSolrVocabCoreAvailable = true;
```
chrisknoll, line 134: I think i found where isSolrVocabCoreAvailable is assigned to true here.  Note, it's not observable, you're just storing a boolean value.   This may work if this all happens when the page loads (ie: the configuration page loads) and there is no change in the isSolrVocabCoreAvailable state while the page is alive.  If that's the case, don't make anything observable.

PR #2873, opened 2023-06-25
Address style concerns and config option to hide permission manager button

rkboyce:
The style concerns focused the proper use of variable types in Permisssion.js and the use of 'perm_type' instead 'role' in parameter names where the former was the correct choice.

There is now a new configuration option for configLocal : 

```
        // This tells Atlas to show the 'lock' icon to the user when
        // viewing concept sets, cohort definitions, and other
        // artificats so that the can edit READ and WRITE
        // permissions. By default, this is set to true
        configLocal.enablePermissionManagement = true;
```
I added code throughout the pages to hide the button when the enablePermissionManagement = false;

I did not address using tabs or pills for the configure access modal now that read permissions are a part of that. Neither the ability to assert that an artifact is globally readable. I would like to punt these changes to 2.15 because of limited time on my part.


anthonysena:
@rkboyce from our discussion on the Atlas WG call, these changes are looking good from my review. Per our discussion, you'll be working out the corresponding WebAPI changes for this work after the 7/4 break.

chrisknoll:
I'm confused by these 2 PRS:
https://github.com/OHDSI/Atlas/pull/2873
https://github.com/OHDSI/Atlas/pull/2849

Are these the same things?  If so, should these be put into a single PR so we don't have confusion between them?

rkboyce:
Sorry - closing this pull request since the update should have happened on [draft pull request](https://github.com/OHDSI/Atlas/pull/2849)

PR #2888, opened 2023-10-02
Prevent multiple calls to concept set list when loading cohort editor

### LINKED ISSUE #2887: Cohort definition editor loads full list of concept sets multiple times
### Expected behavior

When the cohort definition editor opens, the only requests made to WebAPI are the ones necessary to set up the editor

### Actual behavior

When loading a cohort definition, there is a call to the `WebAPI/conceptset/` endpoint which loads the full list of concept sets in the WebAPI database for each concept set present in the cohort definition. 

Load a cohort definition with > 0 concept sets. Observe the network trace via Google Chrome developer tools:

image

The screenshot above shows 5 calls to `WebAPI/conceptset/` which is the number of concept sets in my cohort definition entry event section. If you open an inclusion rule with > 0 concept sets, `WebAPI/conceptset/` is once again called for each concept set present in the inclusion rule.

anthonysena:
Aims to fix #2887 by extending the concept set browser to only load the repository concept set list when the modal is activated.

### DIFF
--- a/js/extensions/bindings/datatableBinding.js
+++ b/js/extensions/bindings/datatableBinding.js
@@ -298,7 +298,7 @@ define([
 			table.clear();
 
 			// Rebuild table from data source specified in binding
-			if (data.length > 0)
+			if (data && data.length > 0)
 				table.rows.add(data);
 
 			// drawing may access observables, which updating we do not want to trigger a redraw to the table
--- a/js/services/AuthAPI.js
+++ b/js/services/AuthAPI.js
@@ -517,7 +517,9 @@ define(function(require, exports) {
 
     const executeWithRefresh = async function(httpPromise) {
         const result = await httpPromise;
-        await refreshToken();
+        if (config.userAuthenticationEnabled) {
+            await refreshToken();
+        }
         return result;
     }
 
--- a/js/components/atlas.cohort-editor.html
+++ b/js/components/atlas.cohort-editor.html
@@ -17,6 +17,7 @@
 					<div class="modal-header" data-bind="text: ko.i18n('components.atlasCohortEditor.importConceptSet', 'Import Concept Set From Repository...')"></div>
 					<div class="paddedWrapper">
 						<concept-set-browser params="
+							showModal: $component.showModal,
 							criteriaContext: $component.criteriaContext,
 							cohortConceptSets: $component.currentCohortDefinition().expression().ConceptSets,
 							onActionComplete: $component.onAtlasConceptSetSelectAction,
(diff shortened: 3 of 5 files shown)


js/extensions/bindings/datatableBinding.js
```diff
@@ -298,7 +298,7 @@ define([
 			table.clear();
 
 			// Rebuild table from data source specified in binding
-			if (data.length > 0)
+			if (data && data.length > 0)
```
anthonysena, line 301: This change is unrelated to the referenced issue. This fixes an error whereby the `data` object is not defined and so calling .length causes an error in the browser.

js/services/AuthAPI.js
```diff
@@ -517,7 +517,9 @@ define(function(require, exports) {
 
     const executeWithRefresh = async function(httpPromise) {
         const result = await httpPromise;
-        await refreshToken();
+        if (config.userAuthenticationEnabled) {
+            await refreshToken();
+        }
         return result;
```
anthonysena, line 523: This change is also unrelated to the referenced issue. This addition fixes a bug that prevents running Atlas when security is disabled.

chrisknoll:
Suggestion (sorry I didn't try this myself):  I noticed you have set up subscriptions (make sure you dispose them) to check if the showModal observable changes, and it is true, to do the concept set fetch....

I think the problem is that inside the modal component, we place the concept set component that does the fetch and it's just the lifecycle of the component that wants to do it: we instantiate 5 components so we will fetch the 5 concept sets.

To prevent the instantiation of the component could we put an `if:` block around the contents of the modal such that everything inside the module won't be instantiated until the showModal goes true, and then the component gets instantiated and the service call is made? (This is theoretically)

The end effect shoudl be the same as what you have (ie: the if: binding observes the showModal and will control the load) but without creating addiitional manual subscriptions that you need to clean up.

anthonysena:
> Suggestion (sorry I didn't try this myself): I noticed you have set up subscriptions (make sure you dispose them) to check if the showModal observable changes, and it is true, to do the concept set fetch....
> 

That's correct - I need to clean up the subscription for sure.

> I think the problem is that inside the modal component, we place the concept set component that does the fetch and it's just the lifecycle of the component that wants to do it: we instantiate 5 components so we will fetch the 5 concept sets.
> 
> To prevent the instantiation of the component could we put an `if:` block around the contents of the modal such that everything inside the module won't be instantiated until the showModal goes true, and then the component gets instantiated and the service call is made? (This is theoretically)
> 

In theory that makes sense and would work well. What I recall from the code base is that the component for displaying the concept set is used in a number of places. So, my thinking was that by extending the component to be "modal aware" through its exposed parameter list, it would make modifying other things in Atlas a bit easier so that we only need to worry about where it is used modally vs. having to think about decorating those code blocks where it is used modally. They are functionally equivalent in my mind so I don't have a strong opinion either way. 

> The end effect shoudl be the same as what you have (ie: the if: binding observes the showModal and will control the load) but without creating addiitional manual subscriptions that you need to clean up.

Agreed - in the meantime, I will update this PR to clean up the subscription but if we want to go with the `if:` block approach we can close this out in favor of that set of changes.

anthonysena:
@chrisknoll I've updated this PR to use the `if:` approach you described and removed the subscription. It should be ready for review now.

Issue #2928, opened 2024-04-23
[Enhancement] "Globally" shared artifacts


rkboyce:
Some implementers of Atlas/WebAPI need to restrict read access to artifacts e.g., to just those user accounts that have authored the artifacts. This functionality became available in v2.14 of WebAPI and Atlas. In such a case, a user can only see the artifacts they own and that they have explicitly been given READ access to. As a result, it becomes desirable to have a way to share at least some artifacts "globally", meaning to all users of the system. For example, in a data commons environment, the leadership might want to provide certain concept sets and cohorts as examples to the broad community of users.  This is an enhancement proposal to add that ability.

Pre-conditions: 
- Atlas and WebAPI versions >= 2.14 with security enabled and security.defaultGlobalReadPermissions set to 'false' in WebAPI
- In the WebAPI - a system role that all users will receive that indicates the ability to read globally-shared artifacts (e.g., 'shared artifacts reader')
- (optional) In the WebAPI - A non-system role used to restrict the ability to share artifacts globally to a specific set of users (e.g., 'shared artifacts owner')

Suggested changes to Atlas:

- js/config-local.js: 
   - a new config  `configLocal.permissionManagementRoleId`  
      - By default set to an empty string ('') indicating that all users can share globally. 
      - Optionally, it can be set to the id of the a non-system role used to restrict the ability to share artifacts globally to a specific set of users

- The configure access modal:
   - js/components/security/access/configure-access-modal.js: functions to grant and revoke 'global' read access. These simply call existing async functions to have WebAPI add or remove GET permissions for a given artifact to  the sec_role_permission table for all users with the 'shared artifacts reader' role
   - js/components/security/access/configure-access-modal.html: toggle buttons that a user can click to share or unshare a given artifact globally.

- The "manager" JS pages for each of the Atlas sub-apps (e.g., for cohort definitions `js/pages/cohort-definitions/cohort-definition-manager.js` and for concept sets `js/pages/concept-sets/conceptset-manager.js`) : 
   - a new observable to flag if a user is allowed to share (`userCanShare`). This is set to false by default and changed to true in two cases:
      -  the `configLocal.permissionManagementRoleId`  is set to an empty string (default) or,
      -  an async function (`checkIfRoleCanShare`)  that checks if the user has the role id assigned to `configLocal.permissionManagementRoleId`  returns true
    
- The "manager" HTML pages for each of the Atlas sub-apps:
[...]

PR #2929, opened 2024-04-23
Feature/global sharing

rkboyce:
This is a draft pull request that shows a test implementation of the enhancement described in [this Atlas ticket ](https://github.com/OHDSI/Atlas/issues/2928). 

Here is a [video of the functionality](https://pitt.hosted.panopto.com/Panopto/Pages/Viewer.aspx?id=ec122108-d7fb-4789-9335-b15b012066d6) 

This changes in the draft pull request implement the general functionality needed in the configure-access-modal and show the role dependent  visibility of the lock icon for the concept set and cohort definition managers. The final pull request will apply the latter to all relevant sub-apps in Atlas.


pieterlukasse (commented):
thanks @rkboyce ! Please see my questions below.

js/services/ShareRoleCheck.js
```diff
@@ -0,0 +1,32 @@
+define(function (require, exports) {
+    var $ = require('jquery');
+    var constants = require('const');
+    const httpService = require('services/http');
+    
+    async function getRoleUsers(subject, permissionManagementRoleId) {
+	return await httpService.doGet(constants.apiPaths.roleUsers(permissionManagementRoleId))
```
pieterlukasse, line 7: this call is failing on my side with 403, which I think makes sense. I would argue that the ability to query which users have any specific role, is an "admin level" type of query and should not be open to all. Can we change this code to just check if the current user has the role?
Or perhaps base the ability to share on a specific (new?) permission. Permissions can be accessed at `/user/me` endpoint.

pieterlukasse: Or use the `user/{userId}/roles` endpoint:
https://github.com/OHDSI/WebAPI/blob/6a43da50ab2bf10df18306df658933ec6a49235e/src/main/java/org/ohdsi/webapi/service/UserService.java#L135-L138

pieterlukasse: I tried this endpoint above here: https://github.com/uc-cdis/Atlas/pull/46/files 
But that also fails with 403. So, we'll need something else. I'm considering a new webapi endpoint as well as a possible solution if we don't want to introduce a specific permission for this share functionality (see my comments in my own PR)

pieterlukasse:
thanks for merging in my proposed changes @rkboyce 👍 
You can mark all my comments above as resolved.

js/pages/cohort-definitions/cohort-definition-manager.html
```diff
@@ -77,20 +77,20 @@
...
-			<!--
+			
 			<li role="presentation"
-				data-bind="visible: !previewVersion(), css: { active: $component.tabMode() == 'explore' }, click: function() { $component.selectTab('explore'); }">
-				<a data-bind="text: ko.i18n('cohortDefinitions.cohortDefinitionManager.tabs.explore', 'Explore')"></a>
+				  data-bind="visible: !previewVersion(), css: { active: $component.tabMode() == 'explore' }, click: function() { $component.selectTab('explore'); }">
+			    <a data-bind="text: ko.i18n('cohortDefinitions.cohortDefinitionManager.tabs.explore', 'Explore')"></a>
 			</li>
-			-->
+
```
pieterlukasse, line 93: @rkboyce is this change enabling a new tab that was previously hidden? If yes, can you add some info on why?

pieterlukasse:
@rkboyce FYI - I found one issue in this solution where the "lock icon" only appears to the owner of the cohort definition, regardless of our global sharing configuration. I am trying to work around this (see this commit https://github.com/uc-cdis/Atlas/pull/47/commits/b8fa9399de36fa5a6b697ee922d6693ba9943dbf added to https://github.com/uc-cdis/Atlas/pull/47). However, this seems to reveal some more issues. Please take a look and let me know if you're able to reproduce it as well.

js/pages/cohort-definitions/cohort-definition-manager.html
pieterlukasse: Also discussed here: https://github.com/vinci-ohdsi/Atlas-2.15-dev/commit/8396b4d2d38ad9eaaa05bc2d2cc587f563a00385#r142058479 
It seems this has been disabled for the last 7y ⚠️

chrisknoll:
> @rkboyce FYI - I found one issue in this solution where the "lock icon" only appears to the owner of the cohort definition, regardless of our global sharing configuration. I am trying to work around this (see this commit [uc-cdis@b8fa939](https://github.com/uc-cdis/Atlas/commit/b8fa9399de36fa5a6b697ee922d6693ba9943dbf) added to [uc-cdis#47](https://github.com/uc-cdis/Atlas/pull/47)). However, this seems to reveal some more issues. Please take a look and let me know if you're able to reproduce it as well.

I think that's correct:  I think the current decision is that only owner can control access to an asset.   I don't think that this is documented, but I can't recall anything about permissions that are granted that grant permissions to control access rights.   Apologies if I am mistaken.

pieterlukasse:
thanks @chrisknoll . I found the PR where this was introduced: https://github.com/OHDSI/WebAPI/pull/1273
This is the endpoint that is called when the permissions modal is opened: [AtlasSecurity.java - Line214](https://github.com/OHDSI/WebAPI/pull/1273/files#diff-11a1d193ee2b34d10b038b8e20a905d60efd8564bf12761afbdec21739c191e4R214)
In my case, the call `WebAPI/permission/access/COHORT_DEFINITION/445/READ` returns 403. 

The PR mentioned above, introduced a new [PermissionController.java](https://github.com/OHDSI/WebAPI/pull/1273/files#diff-6a8d7003cf74a2a8e1d14452db2b5ae61bed872aea634fb7768b46433828b25b) and a [PermissionService.java](https://github.com/OHDSI/WebAPI/pull/1273/files#diff-73a7b40ab690bfc552a8c9b9b714799846309d76fbe19c2820210817f3ce7d86). 
@rkboyce let's discuss with the business today and see if they are OK with this restriction.

pieterlukasse:
This might actually work as a solution for now, based on our current "teamproject" WebAPI code: https://github.com/uc-cdis/WebAPI/pull/137/files

pieterlukasse:
> This might actually work as a solution for now, based on our current "teamproject" WebAPI code: https://github.com/uc-cdis/WebAPI/pull/137/files

update: this solves the problem. Let me know what you think!

pieterlukasse:
@rkboyce please merge in https://github.com/vinci-ohdsi/Atlas-2.15-dev/pull/7 and then I think this PR is go (for testing phase).

rkboyce:
I think that we have completed the testing and discussion for this feature. I am closing the draft pull request and opening an actual one.

PR #2951, opened 2024-08-30
Tools module functionality

alex-odysseus:
Addressing #2330

There was a new widget-like view designed in contrast to the usual tabular views. The toggle button component defines if a Tool should be displayed to a standard User. Administrators should see all Tools display and change the Tool visibility


js/pages/tools/tool-manager.html
```diff
@@ -0,0 +1,103 @@
+<heading-title params="name: ko.i18n('common.tools'), icon: 'toolbox', theme: 'dark'"></heading-title>
```
alex-odysseus, line 1: We may need some formatting applied to a few rows

anthonysena:
Just some notes on the content of this pull request based on my testing. 

1. When adding a new tool, the button "Apply" should instead be "Add" or "Save":

image

2. The layout of this page feels a little awkward but it is functional. Some suggestions: 
- Move the edit/delete icons to appear along side the link title (assuming these are only shown to admins).
- The entire area of the "tool" is clickable to take you to the URL of the tool in a new tab. Perhaps add some hover-over to show when the user is in an active area of the screen to note that it is clickable? Right now there is no hover-over behavior as I would expect when clicking a link.
- I'm unclear if the created by and date are always displayed or is this only for admins?

image

anthonysena (dismissed):
Please see suggested changes.

pieterlukasse:
hi, this PR mentions "Addressing https://github.com/OHDSI/Atlas/pull/2330" , but that seems like a typo. Can you please check?

alex-odysseus:
Thank you, Pieter, I have incorrectly referenced the issue from the Atlas project which should be the WebAPI one, corrected it out @pieterlukasse

PR #2956, opened 2024-09-16
Feature/global sharing

rkboyce:
This pull request adds OPTIONAL functionality to Atlas/WebAPI where folks have chosen to implement optional enhancement that allows read permission to be restricted so that only users with read access to a given artifact can view them (see [description in the Atlas set up guide](https://github.com/OHDSI/Atlas/wiki/Atlas-Setup-Guide) and [more details on the WebAPI wiki](https://github.com/OHDSI/WebAPI/wiki/Read-restricted-Configuration). 

The feature makes it possible for users of an Atlas instance that is restricting read access to share an artifact they own publicly. Example:

1. User 1 creates a concept or cohort definition and wants to make it public so that it is visible in the list of artifacts returned when another user goes to the respective listing
2. User 1 clicks on the lock icon which pops up a modal to configure permissions. The user scrolls to the bottom and selects the button and selects "Granted" under where it says "Status of global READ access"
3. The effect of this is to add permission mappings in the sec_role_permission table so that the public role has read  access to the artifact. 

This operation can be reversed:
1. User 1 decides to remove the artifact from public shared status. The user clicks on the lock icon which pops up a modal to configure permissions. The user scrolls to the bottom and selects the button and selects "Not Granted" under where it says "Status of global READ access"
2. The effect of this is to remove permission mappings in the sec_role_permission table so that the public role no longer has read  access to the artifact. 

One additional feature is that artifact sharing can be restricted to only certain users. This pull request adds the enablePermissionManagement config option to config-local.js. If is set to true then, only users a specific permission ('artifact:global:share:put'), are able to share. This is useful for data commons or other collaboratives where there are many users and a small group of admins or moderators  would like to filter the items shared publicly. 

If this pull request is accepted, the documentation above should be copied into the  the Atlas set up guide](https://github.com/OHDSI/Atlas/wiki/Atlas-Setup-Guide) and referenced from the WebAPI wiki](https://github.com/OHDSI/WebAPI/wiki/Read-restricted-Configuration).


chrisknoll:
Hi, I planned on working on merging this in but I'm now confused as to if there is an outstanding issue that needs to be resolved from https://github.com/OHDSI/WebAPI/pull/2342?  The atlas side looks like it adds the 'global read access' option to the security (lock) screen, and that would just put this artifact permission into role 15 (where all users have the read access permissions for global assets stored).

So can anyone please clarify what this PR depends on in order to be merged?   My understanding is:

Role 15 (read access role) is already set up on WebAPI, and this PR just puts assets into it from the UI.  Is there anything more needed from that perspective?

chrisknoll:
Ok, not sure if this is working as intended:

I created a new one and want to grant perms as global read:

image

I click the granted button (note: this UI is not clear what the status is if it is granted or not granted, I would make the selected option green and the non selected red or something to distinguish it better.

image

The 'public' role was added but I was expecting role id 15 (read restricted atlas users) to be the permission granted to this asset.

chrisknoll:
Ok, cool thank you. So the main ask is that if the global access button should be a little more clear about if it is globally shared or not.

pieterlukasse:
@chrisknoll yes, we can cleanup the webapi PR and this PR to strictly just what is needed for this feature and push the other items to separate tickets for example. 
Having said that, adding a "global sharing" feature without solving the security/access issues in webapi (i.e. artifacts being accessible while they shouldn't in "read restricted" mode) might be a bit misleading/confusing.

Maybe we can discuss a path forward in one of the next meetings?

chrisknoll:
I was going to add some commits to clean up the button behavior, but it's on a fork which I don't have perms on, and I don't want to fork-a-fork.

So, would it be possible for someone to make the follwing update which should be quick and dirty:

For the buttons 'granted' and 'not granted', can we alter the rendering to just show one or the other. If it's granted, show the granted button. if it is not granted, show the not granted button.  Clicking on the button 'toggles' the state of the read only grant (ie: clicking 'granted' will set it to not-granted, clicking 'not granted' will set it to 'granted').

PR #2971, opened 2024-10-29
Extending a Concept Set with a new tab 'Annotations'

alex-odysseus:
Addressing https://github.com/OHDSI/WebAPI/issues/2318


chrisknoll:
Let's add a configuration option to hide the annotation tab and bypass any annotation state saving calls from the UI.

By default this can be disabled, and we'll have release notes instructing how to enable the feature.

Issue #2812, opened 2022-12-19
Characterization - Custom Feature - "Copy Analysis SQL Template to Clipboard"


ericaVoss:
### Expected behavior
That regardless if I own a feature, I can copy the SQL to clipboard.

### Actual behavior
I can only copy the SQL if I own the feature

Someone else's feature:
![image](https://user-images.githubusercontent.com/6305439/208495332-2c7d9c59-2597-43f8-ab85-27bd9ad9f758.png)

My own feature:
![image](https://user-images.githubusercontent.com/6305439/208495762-909b7c53-8b76-4022-b8ac-22af90ae1afd.png)

### Steps to reproduce behavior

1. Navigate to "Characterizations"
2. Search "feature Analyses" for a custom feature that you do not own, made through SQL not the GUI
3. The "Copy Analysis SQL Template to Clipboard" button is greyed out.

katy-sadowski:
+1!  Also, for other people's features, scroll is disabled in the SQL text box so we can't see the full SQL code.

Issue #2532, opened 2021-05-05
Can not view concept sets in Custom Features when only view access


chrisknoll:
### Expected behavior
When viewing a custom feature that you did not create (in secured mode, this becomes read-only), you should be able to look at the concept sets that are defined in the custom feature.

### Actual behavior
Since it is read only, you can't switch tabs into the custom features and therefore can not view anything about the concept sets.

### Steps to reproduce behavior
In secured mode, create a custom feature under one identity, then open the feature under a different identity.  The form will appear read-only and you will not be able to switch tabs.

PR #2975, opened 2024-11-06
Allow read-only custom feature browsing

anthonysena:
Modifies the custom feature editor to allow for browsing concept sets when in read-only mode per #2532. These changes also enable the copy SQL to clipboard per #2812.

### DIFF
--- a/js/components/multi-select.js
+++ b/js/components/multi-select.js
@@ -14,6 +14,7 @@ define(
 
       self.multiple = params.multiple;
       self.options = params.options;
+	  self.disable = (params.disable || false);
       self.selectedValues = ko.observableArray(params.selectedValues && params.selectedValues());
       self.selectedValue = params.selectedValue;
       self.selectedTextFormat = params.selectedTextFormat || 'count > 2';
--- a/js/pages/characterizations/components/feature-analyses/fa-view-edit/fa-design.html
+++ b/js/pages/characterizations/components/feature-analyses/fa-view-edit/fa-design.html
@@ -16,13 +16,13 @@
             <!-- /ko -->
             <textarea data-bind="
                     css: classes({ element: 'descr', extra: 'form-control' }),
-                    attr: { disabled: $component.canEdit() ? null : true },
+					attr: { disabled: $component.canEdit() ? null : true }
                     textInput: $component.data().descr
                 "></textarea>
         </div>
     </div>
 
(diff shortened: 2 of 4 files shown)


js/pages/characterizations/components/feature-analyses/fa-view-edit/fa-design.html
```diff
@@ -100,35 +103,37 @@
...
-                    <div data-bind="if: $data.criteriaType === 'CriteriaGroup'">
-                        <criteria-group params="{ expression: { ConceptSets: $component.data().conceptSets }, group: $data.expression }"></criteria-group>
-                    </div>
-                    <div data-bind="if: $data.criteriaType === 'DemographicCriteria'">
-                        <demographic-criteria params="{ criteria: $data.expression }"></demographic-criteria>
-                    </div>
-                    <div data-bind="if: $data.criteriaType === 'WindowedCriteria'">
-                        <windowed-criteria params="{ expression: { ConceptSets: $component.data().conceptSets }, criteria: $data.expression, disableObservationPeriod: true, defaultObservationPeriod: true }"></windowed-criteria>
-                    </div>
+				<div data-bind="css: $component.canEdit() ? '' : 'feature-analysis-view-edit__content--disabled'">
```
anthonysena, line 106: This CSS class `feature-analysis-view-edit__content--disabled` is defined by the parent component and I was not sure how to best use it so for now it is hardcoded.

anthonysena: @alex-odysseus do you or anyone from your team provide a suggestion to handle this CSS approach?

anthonysena: Just to follow up on this item: I've attempted to edit the child .less sheets to include the `disabled` BEM modifier without much luck. I'd propose just hard-coding this now.

chrisknoll (approved):

js/components/multi-select.js
```diff
@@ -14,6 +14,7 @@ define(
 
       self.multiple = params.multiple;
       self.options = params.options;
+	  self.disable = (params.disable || false);
```
pieterlukasse, line 17: should this be `self.disabled` to match `attr: disabled,` added to the .html above?

PR #2978, opened 2024-11-12
Adding UI to map non-Standard Concepts to the Standard ones to streamline Concept Set creation

alex-odysseus:
Addressing https://github.com/OHDSI/WebAPI/issues/2348


anthonysena (changes requested):
Looking good but looks like there are now conflicts. Can you resolve those? Thanks!

anthonysena:
This feature is looking good. I am going to walk through the functionality a bit here to capture some notes. 

I started by creating a concept set with 2 non-standard codes:

image

From there I went to the "resolve mappings" tab:

image

My interpretation of what is shown in the screen grab above: there are 3 standard concepts listed under the `Standard Concepts with their non-standard counterparts` heading that map to the 2 non-standard concepts that are listed under the `Result concept set` section. Within the first table the column "Mapped Non-Standard Count" indicates how many non-standard concept set in the concept set map to the standard concept.

I then click on a standard concept to understand the mapping:

image

I see the concept details and a table called `Non-standard concepts from the initial Concept Set` which includes the 2 non-standard concepts in my concept set. The "Map all selected to given Standard" button is activated but clicking it does nothing. I later find out that I have to pick one or more of the non-standard concepts that should be mapped to the selected standard. Can we disable this button when no items are selected? And can we provide some guidance to the user in this area so they know what to do? For example, can we add some text that says: "Select one or more non-standard concepts in the list below to map to the current standard concept: <current standard concept name>".

Below that there is another table that includes all of the source concepts presumably for the standard concept? One suggestion is to rename the title from `Included Source Codes` to "Source concepts mapped to <current standard concept name>". Additionally, this display should use the red encoded link colors to indicate they are non-standard concepts to be consistent with the rest of the ATLAS UI.

On the bottom of the "Manual Mapping Resolution" is an area to add the "Included Source Codes" to the current concept set. Is this the intended behavior? I only ask since I wasn't sure why you'd want to add non-standard concepts at this point:

image

So I decided to resolve my ICD10CM code to the standard code. I selected it and hit the "Map all selected to given Standard" button:

image

Very nice! Now my non-standard code is replaced. I can do the same for the other non-standard code. But then how do I make use of the "Result concept set"? This wasn't obvious but the answer was the "Preview" button:

image

Again, maybe some text in this area to help guide the user. As a suggestion below the "Result Concept Set" we could add "Use the "Preview" button to review the changes made and to apply them to the current concept set". I'd also suggest that the "Reset to initial state" be deactivated if the result concept set == the current concept set. If this is too complex to do, that is fine but it is a bit confusing to have the button active and have it do nothing.

Overall this seems to be in good working order. Nicely done!

chrisknoll (approved):

chrisknoll (approved):

chrisknoll (approved):

Issue #2847, opened 2023-03-30
Clear Server Cache goes to a dead endpoint


chrisknoll:
### Expected behavior
Clicking clear server cache in configuration screen should result in work being done.

### Actual behavior

Call returns a 404 in the console.

### Steps to reproduce behavior

1. Log into Atlas
2. Go to configuration screen
3. click Clear Server Cache
4. Observe 404 in console.

chrisknoll:
I think this button was a leftover from when encache was removed.

I don't think i see a UI element that goes to `/sources/refresh` so I was thinking that this button could be reclaimed to do this.

chrisknoll:
Other option is that we can clear `achilles_cache` and `cdm_cache` with this button.

konstjar:
I like the last proposal

PR #2980, opened 2024-11-15
Have the Clear Server Cache button clear the cdm_cache and achilles_cache tables

amarsan:
closes https://github.com/OHDSI/Atlas/issues/2847

Companion PR is https://github.com/OHDSI/WebAPI/pull/2406

I added an .editorconfig file so that this project can have its own style conventions. I noticed that there were mixed spaces/tabs and tab sizes within individual files. I recommend turning whitespace diffs off when reviewing this PR.

There was an existing Clear Server Cache button. I have it pointing now to the new webapi/cdmResults/clearCache endpoint. I fixed the code so that if the clearCache operation returns an error response we alert the user.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,13 @@
+# EditorConfig is awesome: https://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+end_of_line = crlf
+charset = utf-8
+trim_trailing_whitespace = false
+insert_final_newline = false
+quote_type = single
--- a/js/pages/configuration/configuration.js
+++ b/js/pages/configuration/configuration.js
@@ -14,7 +14,7 @@ define([
   'services/job/jobDetail',
   'services/CacheAPI',
   'less!./configuration.less',
-  'components/heading'
+  'components/heading',
 ], function (
   ko,
   view,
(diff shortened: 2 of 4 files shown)


chrisknoll:
The button doesn't seem to reset it's state after clicking:

image


Should it go back to the blue state after the call is complete, or is it intended to be red after?

amarsan:
@chrisknoll I did not work on that button. I only worked on the "Clear Server Cache" button that is below the table.

amarsan:
That "Refresh Cache" button calls that code that "warms" the cache. The warming is done in an async job.

chrisknoll:
Ok, sorry. I misunderstood. It hought the individual buttons called the clear cache by source, and you made the bottom one do the same thing for each source.

Ok, can I ask that the buttons next to the source do the following:

1. Call the delete cache results in the WebAPI db for the given source (sync operation)
2. Call warm cache (the existing functionality).

I think we want to be able to isolate the clear cache of the individual sources...usually you have mutliple cdms set up and you only want to refresh the cache on one of them at a time.

amarsan:
@chrisknoll Just to clarify...you'd like for the "Refresh Cache" buttons in the source table to both delete cache results and warm the cache. Did I understand that correctly?

chrisknoll:
Makes sense: in my DEV env, I disabled all warming caches because I'm launching and relaunching all the time. so I turn that off, but I do want to be able to clear the cache, so making the button do it and then try to warm cache (but reject because it's disabled)  is a good behavior.

chrisknoll (approved):

Issue #3003, opened 2025-03-19
Menu options visible even if user is not logged in


pieterlukasse:
Atlas menu options are visible, even if user is not logged in. It makes more sense to only display items that are really actionable. So, if the user should login, show only the login option. 

E.g. 

![Image](https://github.com/user-attachments/assets/bcdcfad5-338c-4a59-a2d1-b47fc224c898)

PR #3004, opened 2025-03-19
Hide menu options until user has logged in

pieterlukasse:
Addresses https://github.com/OHDSI/Atlas/issues/3003

Ensures menu options are hidden until user has logged in.

### DIFF
--- a/js/Application.js
+++ b/js/Application.js
@@ -96,6 +96,7 @@ define(
 				});
 				this.companyInfoTemplate = config.companyInfoCustomHtmlTemplate;
 				this.showCompanyInfo = config.showCompanyInfo;
+				this.isLoggedIn = ko.computed(() => authApi.isAuthenticated());
 			}
 
 			/**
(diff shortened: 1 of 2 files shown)


js/Application.js
```diff
@@ -96,6 +96,7 @@ define(
 				});
 				this.companyInfoTemplate = config.companyInfoCustomHtmlTemplate;
 				this.showCompanyInfo = config.showCompanyInfo;
+				this.isLoggedIn = ko.computed(() => authApi.isAuthenticated());
```
chrisknoll, line 99: Can you make this ko.pureComputed().  ko.computed is more heavy weight, and requires cleanup after component is disposed.  The pureComputed() is appropraite if there are no 'side effects' in the evaluation of the computed (ie: you set some other observable state in the process of evaluated the computed expression).  Since that's all you are doing here, I think pureComputed is appopriate.

chrisknoll (changes requested):

js/Application.js
pieterlukasse: Thanks! Done.

PR #3014, opened 2025-04-11
Fix: adjust attribute name "disable" to match what is defined in multi-select.js

pieterlukasse:
See also https://github.com/OHDSI/Atlas/pull/2975/files#r2040084688

Fixes error below:
```
knockout-latest.js:16 Uncaught ReferenceError: Unable to process binding "if: function(){return multiple == true }"
Message: Unable to process binding "attr: function(){return disabled }"
Message: disabled is not defined
```

### DIFF
--- a/js/components/multi-select.html
+++ b/js/components/multi-select.html
@@ -1,8 +1,8 @@
 <div class="multi-select" data-bind="eventListener: [{event: 'hide.bs.dropdown', selector: 'div.dropdown.bootstrap-select', callback: onSelectionComplete}]">
     <!-- ko if: multiple == true -->
-        <select data-bind="attr: disable, selectedOptions: selectedValues, options: optionVals, optionsText: optionsText, multiSelect: { liveSearch: true, selectedTextFormat: selectedTextFormat, noneSelectedText: ko.unwrap(noneSelectedText), noneResultsText: ko.unwrap(noneResultsText), countSelectedText: ko.unwrap(countSelectedText) }" multiple></select>
+        <select data-bind="disable: disable, selectedOptions: selectedValues, options: optionVals, optionsText: optionsText, multiSelect: { liveSearch: true, selectedTextFormat: selectedTextFormat, noneSelectedText: ko.unwrap(noneSelectedText), noneResultsText: ko.unwrap(noneResultsText), countSelectedText: ko.unwrap(countSelectedText) }" multiple></select>
     <!-- /ko -->
     <!-- ko if: multiple == false -->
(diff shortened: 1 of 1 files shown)


pieterlukasse:
CC @anthonysena

anthonysena (approved):

anthonysena (changes requested):
I think I introduced this bug based on the work you noted here: https://github.com/OHDSI/Atlas/pull/2975/files#r2040084688. Shouldn't we change the multi-select component to use `disabled` since that is the desired HTML attribute?

chrisknoll:
A few notes on this:
1. The [attr data-binding docs](https://knockoutjs.com/documentation/attr-binding.html): this is a cool binding allowing you to set multiple HTML dom attributes in 1 expression based on the values from the view model.  Works with things like title, href, etc where it's an attribute-value set up.
2.  the [html disabled attribute dom docs](https://www.w3schools.com/tags/att_disabled.asp): marks an element disabled. Note it doesn't have a name-value pair sort of assignment.  Just the presence of `disabled` disables the dom element.
3. the [enabled / disabled data-binding](https://knockoutjs.com/documentation/enable-binding.html):  This seems to bet he data-binding that is associated with disabling DOM elements.

Given the above, does the attr:disabled binding work the way we think?  I imagine it would based on that I think we've seen correct disabled behavior here. In the context of this PR, we need to call the attribute 'disabled' (the view model variable name doesn't matter).  But would it be more appropriate to associate the disabled data binding to this function instead of the proper naming of the HTML dom attribute?

anthonysena (approved):
Thanks for sharing those resources @chrisknoll and to @pieterlukasse for making the updates. The intent with the `disable` property of the component is to use the disabled HTML attribute so these changes look good from my side.

chrisknoll:
To answer your question about if it works dynamicaly:

The view-model field that the HTML template is binding to is called 'disable', defined here:

```
self.disable = (params.disable || false);
```

It will respond dynamically depending on what is in the `params.disable` payload (ie: what the component is being initialized with):
1. An observable:  if params.disable is an observable, then self.disable will reference it as the observable that was apassed in, and the `disable` data-bind will redraw whenever the observable state changes automatically.
2. A plain boolean object:  If you data-bind (via the `disable:` binding for example) to a non-observable value, the data-bind will not respond to anything and so whatever the initial state of the param is will be the final state of the DOM element.
3. nothing at all:  If it is nothing then the `|| false` clause takes effect, and the binding will bind to a non-observable value and follow item 2 above.

PR #3027, opened 2025-09-27
fix: use _ alias for lodash in config module

watilde:
Fix the AMD factory to inject Lodash as _, matching the symbol the module uses
  - Prevents _.mergeWith/_.isArray from falling back to a global lodash object and throwing when it’s
  unavailable
  - No functional behavior change beyond ensuring the config merge logic consistently works in environments
  without a global _

### DIFF
--- a/js/config.js
+++ b/js/config.js
@@ -4,7 +4,7 @@ define([
 	'optional!config-local',
 	'optional!config-gis',
 	'lodash',
-], function (app, termsAndConditions, localConfig, gisConfig, lodash) {
+], function (app, termsAndConditions, localConfig, gisConfig, _) {
 
 	if (JSON.stringify(localConfig) == JSON.stringify({})) {
 		console.warn('Local configuration not found.  Using default values. To use a local configuration and suppress 404 errors, create a file called config-local.js under the /js directory');


chrisknoll:
Interesting find!  I was concerned that renaming the imported var from `loadash` to `_` would have broken the code where it expected to find `lodash` as in here:

```
	function customizer(objValue, srcValue, key) {
		if (key === 'externalLibraries' && _.isArray(objValue)) {
			return objValue.concat(srcValue);
		}
		if (key === 'authProviders' && _.isArray(srcValue)) {
			return srcValue;
		}
	};
```

But lo-and-behold, it's using _ instead of 'lodash'!   I think the proper fix was to change _.isArray to lodash.isArray, but it raises the question about why lodash is being loaded in the global scope and then again as a module using AMD.

Is this worth investigating or should we just go with chainging the imported var from AMD to `_` as you've done here?  I just want to make sure we're not ignoring a symptom of a bigger problem (like incorrectly loading lodash).

watilde:
Apperently, it is not something Atlas does deliberately, Lodash’s published UMD build sets itself on the global object. When RequireJS pulls in node_modules/lodash/lodash.js, the factory runs and executes 
```
root._  = _; (node_modules/lodash/lodash.js)
```

Then `window._` is always present even if every AMD module asks for lodash explicitly. That global fallback is why the code such as `js/components/explore-cohort.js:6` can reach `_` without importing it first.

Keeping the AMD import aliased as `_ in js/config.js:5` will be the right immediate fix and the module uses the dependency it requested instead of silently falling back to `window._` Lodash’s UMD bundle assigns  itself globally (node_modules/lodash/lodash.js), so that global still exists, but it’s just a safety net we should treat. I will make a follow-up sweep to remove the few places that re‑expose or rely on the global and make sure every module imports lodash explicitly. That keeps behavior unchanged today and gives us a clear path to drop the global coupling later if we want to tighten things up.

chrisknoll (approved):

PR #3028, opened 2025-10-05
Remove unused modules

watilde:
I am wondering if we can remove those unused dependencies per my investigations.


chrisknoll:
This will take a little more depth, as some of these might have some side-effects that's not clear from just looking through code (ie: including ko.mapping may cause something to behave different implicitly in the framework that isn't obvious from a code scan.

At OHDSI this week, but will pull it and check it as soon as I can.

package.json
```diff
@@ -59,27 +58,22 @@
...
     "director": "1.2.8",
     "eonasdan-bootstrap-datetimepicker": "4.17.49",
-    "express": "4.18.2",
     "facets": "0.1.1",
     "file-saver": "1.3.8",
     "hash-it": "4.1.0",
     "html2canvas": "1.4.1",
     "jquery": "1.12.4",
     "jquery-ui": "1.13.2",
-    "jquery-ui.autocomplete.scroll": "0.1.9",
```
chrisknoll: The `jquery-ui.autocomplete.scroll` binding was in use but somehow got removed but it is in use. On my local branch I fixed this via the following:

add to settings.js under `paths`:
```
		"jquery-ui": "../node_modules/jquery-ui",
		"jquery-ui/ui/widgets/autocomplete" : "../node_modules/jquery-ui/ui/widgets/autocomplete",
		"jquery-ui.autocomplete.scroll": "../node_modules/jquery-ui.autocomplete.scroll/jquery.ui.autocomplete.scroll.min",
```

(the `jquery-ui` item is an existing line, I added the two elements for autocomplete and scroll so that I could add the shim element:

```
		"bootstrap": {
			"deps": [
				'jquery'
			]
		},
		'jquery-ui.autocomplete.scroll': {
			deps: ['jquery-ui/ui/widgets/autocomplete']
		},
```

Here, `bootstrap` is the existing element, and i added an element for jquery-ui.autocomplete.scroll because there's a load-dependency on the autocomplete element prior to loading the scroll plugin.

How I found this in use was the in `autocompleteBinding.js`,  a knockout binding `ko_autocomplete` is defied, but the options are passed into it from cohort editor like this:

```
ko_autocomplete: { value: PriorDays, source: $component.options.dayOptions, minLength: 0, maxShowItems: 5, scroll: true }
```

note the `maxShowItems` and `scroll` options.   This comes from the `.scroll` plugin.

The last part of the fix was to tell the autocompleteBinding to load the scroll plugin with it, so this is the updated `define()` call in autocompleteBinding:

```
define(['jquery', 'knockout', 'jquery-ui/ui/widgets/autocomplete', 'jquery-ui.autocomplete.scroll'], function ($, ko) {
```

Now we get scrollbars in the ko_autocomplete UI elements!  For example in the cohort builder the 'prior days' dropdown should have a scrollbar.

Would you be able to update your fork with this change to keep the jquery-ui.autocomplete.scroll dependency?

